PR #6, opened 2015-02-26
Pass aura id to custom init function

colinstein:
This change passes an aura's ID into the custom init function. This will enable initializers to pull details about the aura like it's triggers, region settings (e.g. x/y position) and modify them if necessary by going through WeakAuras.auras

It would also allow for simple/consolidating loading of "saved" data and registration of events for certain kinds of auras.

Thoughts: might be better to pass a table named "aura" with ID among other properties?

### DIFF
--- a/WeakAuras.lua
+++ b/WeakAuras.lua
@@ -287,9 +287,9 @@ function WeakAuras.ActivateAuraEnvironment(id)
       -- Run the init function if supplied
       local actions = data.actions.init;
       if(actions and actions.do_custom and actions.custom) then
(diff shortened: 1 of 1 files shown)


Stanzilla:
Linked it on the previous ticket so it's open for discussion http://www.wowace.com/addons/weakauras-2/tickets/212-scratch-area-for-custom-trigger-untrigger-name-icon/

Stanzilla:
Closed by https://github.com/WeakAuras/WeakAuras2/commit/9d21bc00b6402c405a7fda733a51136e02f7759d

PR #4, opened 2015-01-01
Fix taint while deleting auras

stewunknown:
How to reproduce: Create a new icon aura, rename it, move it to an existing dynamic group, duplicate the aura and delete the duplicate.

Error:
1x WeakAuras\WeakAuras-2.0.9.7-19-ge8def1e.lua:2819: attempt to index local 'v' (a string value)
WeakAuras\WeakAuras-2.0.9.7-19-ge8def1e.lua:2819: in function `Delete'
WeakAurasOptions\WeakAurasOptions-2.0.9.7-19-ge8def1e.lua:1173: in function `DeleteOption'
...eGUI-Widgets\AceGUIWidget-WeakAurasDisplayButton.lua:146: in function `func'
FrameXML\UIDropDownMenu.lua:748: in function `UIDropDownMenuButton_OnClick'
[string "*:OnClick"]:1: in function <[string "*:OnClick"]:1>

PR #8, opened 2015-04-09
Ticket #4 - Load Condition on ENCOUNTER_START

Tercioo:
Add an extra option on Load panel to type the Encounter ID.
Tracks ENCOUNTER_START/END to load and unload the aura, also has
protection agaisnt disconnect and combat leave by zone change.


Tercioo:
I forgot to mention, this adds "desc = arg.desc" to support tooltips declared directly on the prototype,
Option panel, also, was replacing the loadFrame OnEvent function without sending the args.

Stanzilla:
merged, thanks. for the next time, please use the same indenting that is already used in the WA files, it is (sadly) using spaces :/ But I'd like to keep it consistent.

PR #9, opened 2015-08-08
Load Init script if the aura has EncounterID trigger

Tercioo:
Auras with EncounterId load condition now has their Init script (if any)
run immediately upon entering on a raid instance.

### DIFF
--- a/WeakAuras.lua
+++ b/WeakAuras.lua
@@ -126,6 +126,8 @@ WeakAuras.me = GetUnitName("player",true)
 WeakAuras.myGUID = nil
 local playerLevel = UnitLevel("player");
 
+WeakAuras.currentInstanceType = "none"
+
 local function_strings = WeakAuras.function_strings;
 local anim_function_strings = WeakAuras.anim_function_strings;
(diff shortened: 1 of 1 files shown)


elvador1:
I can't figure out what you are trying to solve with that commit. Can you post a WeakAura which doesn't work right now but would work with your changes?

Tercioo:
This is just a hotfix for a implementation that I've made months ago about load from encounter ids. 
It fixes an issue where callbacks registered inside init script won't get initial callbacks sent by other addons because they receive encounter_start event first than us.

elvador1:
Can you link such an Aura? I don't see why you would register a callback in the init script.

I had this idea to solve that problem: http://i.imgur.com/e0hebJ7.png

Tercioo:
Sorry for the delay mate, github emails went to my spam box.

http://pastebin.com/pFevCPF8
Import, type /dbm -> options tab -> general options on left menu -> click "Start test bars",
Expect to see the aura icon when the first bar called "Test Bar" hits 6 seconds remaining.

I didn't understood your schema on the image, so I can't give my opinion about it.

PR #10, opened 2015-08-26
Smallpatches

InfusOnWoW:
Various small patches

### DIFF
--- a/GenericTrigger.lua
+++ b/GenericTrigger.lua
@@ -1516,6 +1516,7 @@ function GenericTrigger.GetNameAndIcon(data)
       name = WeakAuras.event_prototypes[trigger.event].nameFunc(trigger);
     end
   end
+  return name, icon
 end
 
 function GenericTrigger.CanHaveTooltip(data)
--- a/GenericTrigger.lua
+++ b/GenericTrigger.lua
@@ -718,6 +718,9 @@ end
 
 function GenericTrigger.AllAdded()
   -- Remove GTFO options if GTFO isn't enabled and there are no saved GTFO auras
+  if (GTFO) then
+    return
+  end
   for id, event in pairs(events) do
     for triggernum, data in pairs(event) do
       if (data.trigger.event == "GTFO") then
--- a/GenericTrigger.lua
+++ b/GenericTrigger.lua
@@ -143,6 +143,9 @@ function TestForLongString(trigger, arg)
   return test;
 end
 
+-- Uh oh...
+WeakAuras.TestForLongString = TestForLongString;
+
 function TestForMultiSelect(trigger, arg)
   local name = arg.name;
   local test;
@@ -478,6 +481,7 @@ function GenericTrigger.Add(data, region)
             elseif(trigger.unevent == "auto") then
               untriggerFuncStr = ConstructFunction(event_prototypes[trigger.event], trigger, true);
             end
+
             if(untriggerFuncStr) then
               WeakAuras.debug(id.." - "..triggernum.." - Untrigger", 1)
               WeakAuras.debug(untriggerFuncStr);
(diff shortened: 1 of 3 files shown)


Stanzilla:
In the future, it would help me quite a bit, if you could just list the newly added translations in the pull request. I have to add them to the translation system manually and sometimes miss them in the code.

InfusOnWoW:
Okay, I'll try to do that.

Stanzilla:
http://dev.deadlybossmods.com/rss/DBM_Alpha_Changes-r14433.html

Stanzilla:
Left some more comments on stuff. Is there anything you really want to get done before the next release?

InfusOnWoW:
Yeah let me fix the DBM trigger then and investigate the PLAYER_MOVING thing.

Stanzilla:
@InfusOnWoW Okay, sounds good. I also added some input to http://www.wowace.com/addons/weakauras-2/tickets/403-load-conditions-bug-lua-error/ in case you did not notice.

InfusOnWoW:
Right, that bug report too, and then your comments in that commit, I'll do that today or tomorrow.

And I'm planning on doing a short video showing the new bigswigs / dbm triggers.

Stanzilla:
Alright, I'm currently going through WA.lua, cleaning stuff up here and there and fixing the horrible indenting. I will only commit that after your stuff so I don't give you any trouble with the whitespace changes.

InfusOnWoW:
Here you go: https://github.com/WeakAuras/WeakAuras2/pull/11

Some notes:
I added a version check for DBM. I could have also supported both the old callbacks and the new ones, but I guess people update DBM frequently enough, that it doesn't matter much if only the newest version is supported.

The PLAYER_MOVING thing is pretty funny, apparently to get the one leg forward pose, WoW is pretending that the user is pressing/release the foward button all the time.

InfusOnWoW:
Ah wait with merging, I've found one small issue in the bigwigs options, I'll do one update later today.

InfusOnWoW:
Okay, I've updated the merge request.

InfusOnWoW:
Seems I forgot to include this patch http://www.wowace.com/addons/weakauras-2/tickets/363-support-for-arena1-5-uid-as-specific-units/ in the merge request.

Stanzilla:
https://github.com/InfusOnWoW/WeakAuras2/commit/c668412df1dac6f1c7a7d7c458436fb08c468ea0 Doesn't look right, if you look at my commit https://github.com/InfusOnWoW/WeakAuras2/commit/6dfb997413dcafd5113f1ca87a0fe4dc91362a26 you can see that I did not even touch that line, did you maybe forget to merge this one? https://github.com/WeakAuras/WeakAuras2/commit/08f83cf49543061e879381788ad6ead7fe78038c

Stanzilla:
I cherry picked the others already though.

InfusOnWoW:
Hmm? I don't see any changes when I fetch. Anyway, the updated pull request contains the same patches except the last one is new.

Stanzilla:
Okay, I might be blind, pushed the last one, too. When do you want me to push my indenting changes?  I guess it will mess up with your WIP patches quite a bit and you would have to handle a few merge conflicts, I can delay pushing that after you are done with those?

InfusOnWoW:
Nah just do it. I'll be fine...

InfusOnWoW:
And thanks for merging.

Stanzilla:
Pushed. Thank *you* for your fixes :) I hope the file is more readable for you too now.

InfusOnWoW:
So, as to the global that you mentioned in WeakAurasOptions.lua. I think that's my fault, and a simple copy and paste error. I think you can safely remove the msg from that call to ShowOptions(), but I don't understand what the msg is

This ShowOptions() is to reshow the options dialog after combat ends and we have hidden the options on combat start. So if you figure out what the msg actually is, then we would probably need to store it in the code 4 lines above and then pass it to ShowOptions.

Stanzilla:
Yeah I realized what it is and what it is used for but not where it came from. Just going to remove it then..

PR #13, opened 2015-11-22
Dialog as a Sound Channel

Faithmelter:
Not sure if this has been left out for a good reason or not?, but I updated Types.lua to include the "Dialog" sound channel as a selection during WeakAura creation.


Stanzilla:
Thanks!

PR #16, opened 2016-02-08
Smallpatches

InfusOnWoW:
The first commit is also attached to the ticket: http://www.wowace.com/addons/weakauras-2/tickets/465-quirk-on-progress-bar/

The second commit is a feature I've been working on for some months, where I feel it's now in a state where someone else should review it. It's not 100% done, but I feel I've reached a good quality now.


Stanzilla:
So I'm pretty sure I can't and won't accept that into the mainline addon, that's just too much. So what I think would be really is a 3rd party plugin, as in WeakAuras_StopMotion, if you can get that to work we can have a repo in the WeakAuras project here and a separate project on WowAce and advertise it on the main one.

Also I have a few threads that might be of interest for you if you don't know them already:

* http://www.wowinterface.com/forums/showthread.php?t=45918
* http://www.wowinterface.com/forums/showthread.php?t=52843

Which are experiments by zork for radial stuff, maybe it helps you. One was using cooldown textures and the other setup was to just take a scrollframe and rotate a half-circle inside of it, then for the other half of the circle you'd either display nothing or the entire half circle texture based on how much is supposed to be missing.

InfusOnWoW:
Yeah I know those posts, that's a different technique, that is somewhat better for the simple cases, but has serious limitations in that e.g. the either the start or end angle needs to be 90°. And that it can't be used with more complex animations. This approach here has its own problems, e.g. using lots of textures feels not good. And the first time each texture is loaded it might flicker and so on.

I'll take a look at trying to make it a separate addon, can you point me to any guides on what to do to distrubute a addon, and what I need to do in terms of infrastructure work? 

Ah IRC, well, someday...

Stanzilla:
It's essentially just filling out http://www.wowace.com/addons/create-project/ if you need any help, just ask

Stanzilla:
Oh and I forgot to ask but, why not one texture and repeated TexCoord calls?

InfusOnWoW:
Hmm, that would probably work a bit better, but be more complex to create for the custom textures. I'll give that a try anyway.

InfusOnWoW:
So I experimented and it seems the maximum texture size is 8192*8192 nowadys. I read somewhere that it is 1024*1024. If I use textures with 256*256 pixels. Which should be fine for most use cases, I can put 1024 frames into one big texture. Or if I use 512x512, that would be 256 frames, which is probably good for most cases. 

I guess I have some work to do, thanks for that suggestion. That should definetly improve how it behaves.

Okay, I'll rework the patch into a separate addon then and will fill out that form. I guess, after that I'll have plenty of questions, but let's see.

semlar:
I thought I'd chime in, since Stanzilla mentioned this on IRC and I had an idea for an alternative solution.

First, I draw 4 textures in a square and offset their coordinates so they appear to form a single image.

Then I create another texture and anchor one of its corners to the center of the other 4 and use an animation to rotate it to a specific angle, while setting its texture coordinates to rotate it in the opposite direction to keep it lined up with the image.

I put that rotated texture inside of a scroll frame to clip its edges, and move the scroll frame to one of the quadrants based on the angle it's supposed to be displaying.

The result is something similar to the cooldown animation, and it can be applied to any image without modification.

Here are a couple animations to demonstrate the concept..

This one shows the rotating texture outside of its scrollframe, which is represented as a red square behind it: http://gfycat.com/BriefShowyElephantseal

And here it is showing clockwise / counterclockwise and reverse fills: http://gfycat.com/RealisticBaggyDarklingbeetle

Here's a link to the script I wrote if you'd like to adapt it for weak auras: http://pastebin.com/2ekWh40c

InfusOnWoW:
Oh the gifs look interesting, I haven't yet groked your explanation how it works though, I'll take a look at the script to figure it out...

InfusOnWoW:
I still don't understand that code, though it took me also a while to understand the technique zork posted about. But from the looks of it, this looks vastly superior (though slightly different in use case) to what I have. 

@Stanzilla Also this looks pretty small, and my fist instinct would be to add it to the already existing progress texture as a separate "orientation", e.g. add clockwise rotation and anticlockwise rotation to that mode.

@semlar  Ah I think I understood how it works. That is very clever.

Suppose we want to show a mushrrom

There are 4 textures in the background arranged in the grid, those are all the same texture and each of them shows 1/4 of the mushroom. Now, the reason for that seems to be that we want to hide the "corners" that should be completly invisible. 

Thus our job is reduced to "clipping" one corner texture. For that we position a scroll frame over that corner. Then we put add a texture that is rotated in the right way into it, so that the scrollframe + texture rotation clips it in the right place. And then we set the texture coords on that texture to the right values to get the missing wedge texture.

Have you showd that to zork? It seems to build on his technique and I think he would appreciate knowing this.

Stanzilla:
Yeah agree, that would be an addition as orientation

InfusOnWoW:
Take a look at https://github.com/WeakAuras/WeakAuras2/pull/18

zorker:
Just to post that here aswell. Yes I appreciate that very much! :) semlars technique is awesome.

PR #17, opened 2016-02-13
Smallfixes

InfusOnWoW:
Two small tickets.


InfusOnWoW:
So do you prefer merge requests? Probably slightly less work for both of us...

Stanzilla:
Yeah I do actually, I also notice them faster as you probably noticed :D

PR #18, opened 2016-02-13
Progress Texture: Add circular progress

InfusOnWoW:


InfusOnWoW:
This adds the "orientation" modes clockwise and anticlockwise. I tried to implement most of the existing features for progress textues, but couldn't figure out how to implement mirror. (I don't have a intuition for SetTexCoords at all.) And I also didn't figure out how to independently crop x and y. Even though those are missing, I think this is good enough for the first version and likely to be very well received by our users.

Stanzilla:
Is something maybe? Getting an error right when I create a new progress texture:

```
3x ...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0-60.lua:1803: AceConfigRegistry-3.0-15:ValidateOptionsTable(): WeakAuras.args.region.args.orientation.values: expected a methodname, funcref or table, got 'nil'
[C]: ?
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:50: in function <...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0.lua:45>
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:187: in function <...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0.lua:182>
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:224: in function <...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0.lua:196>
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:231: in function <...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0.lua:196>
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:231: in function <...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0.lua:196>
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:259: in function `ValidateOptionsTable'
...-3.0\AceConfigRegistry-3.0\AceConfigRegistry-3.0-15.lua:301: in function `app'
...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0-60.lua:1803: in function `Open'
WeakAurasOptions\WeakAurasOptions-2.1.0.20.lua:7930: in function `FillOptions'
WeakAurasOptions\WeakAurasOptions-2.1.0.20.lua:8051: in function `finishPicking'
WeakAurasOptions\WeakAurasOptions-2.1.0.20.lua:8083: in function `PickDisplay'
WeakAurasOptions\WeakAurasOptions-2.1.0.20.lua:8391: in function `PickAndEditDisplay'
WeakAurasOptions\WeakAurasOptions-2.1.0.20.lua:7992: in function <WeakAurasOptions\WeakAurasOptions.lua:7972>

Locals:
msg = ": expected a methodname, funcref or table, got 'nil'"
errlvl = 6
t = <table> {
 1 = "WeakAuras"
 2 = "args"
 3 = "region"
 4 = "args"
 5 = "orientation"
 6 = "values"
}
select = <function> defined =[C]:-1
tinsert = <function> defined =[C]:-1
error = <function> defined =[C]:-1
MAJOR = "AceConfigRegistry-3.0"
tconcat = <function> defined =[C]:-1
```

Stanzilla:
When I set `Start Angle` to `90` it looks like this 
wow-64_2016-02-18_21-36-53
 is this correct? It does not start where the highlight texture begins, is what I'm trying so say.

Stanzilla:
Setting both `Start Angle`and `End Angle` to `180` creates an interesting looking glitch :)

Stanzilla:
`Start Angle` `360` and `End Angle` to `0` breaks it, as expected. Makes me wonder if we should sanitize the input?

Stanzilla:
Somehow the aura shows something different when adjusting the settings vs actually triggering it: 

See https://dl.dropboxusercontent.com/s/nv6m1hur5nc0c2o/2016-02-18_22-06-36.webm?dl=0

Stanzilla:
Rotate bugs out a bit with that straight line showing up that does not show up in other animations (not sure if related) https://dl.dropboxusercontent.com/s/jei593jqh79u77j/2016-02-18_22-44-57.webm?dl=0

InfusOnWoW:
I'll push a update of it shortly. I fixed every combination of clockwise and inverted to actually do what it was supposed to do, and also fixed the two cases of start/end combination that didn't work.

As, to your last two comments: The preview shows only a limited amount of progress, because that's what the options dialog does. If you create a aura bar, you see it also only animate from roughly 30% to 60%. (We might want to change that.)

The rotation, is due to that particular texture using up the edge pixels, which WoW unhelpfully extends. That happens with all kind of rotations, e.g. you can see the same effect with the standard progress textures.

InfusOnWoW:
And many thanks for the good testing.

Stanzilla:
`As, to your last two comments: The preview shows only a limited amount of progress, because that's what the options dialog does. If you create a aura bar, you see it also only animate from roughly 30% to 60%. (We might want to change that.)`

Yeah I remembered that after I was done writing my comments, might wanna change that some day indeed. Will test your new patch soon.

InfusOnWoW:
Did you have time to test this again?

semlar:
I think the background texture should also be cropped from the starting angle to the ending angle, rather than display the entire texture and only partially fill it up.

InfusOnWoW:
Hmm, right, that would mean duplicating the whole 4 + wedge texture thing.

Stanzilla:
@InfusOnWoW yes it worked well so far, not sure if @semlar found some more stuff but I just have to write down all the new localization before I merge this or do you want to get more into this PR?

InfusOnWoW:
You can merge that if you want, and I'll fix the background issue in a second patch. @semlar Maybe you want to make more comments?

semlar:
If you limit the end angles to 90 degree increments you could do it with 3 or 4 textures.

I realize that this reduces the functionality, but I think it would make it simpler to use, as well as design.

Every time I went to adjust the end angle I had to manually type the number in because I only ever wanted a 90 degree angle and it doesn't move in 90 degree steps.

semlar:
If you want to keep the ability to end at an arbitrary angle you would basically have to put 2 spinners on top of each other, but there isn't really much alternative.

Right now, if you have a circular texture and you want to create an arc, you can't, because it displays the entire circle in the background and only limits how much it gets filled up.

http://gfycat.com/VioletJauntyGelding This is the kind of thing I'm talking about, except with 90 degree angles.

InfusOnWoW:
Yeah, I think I want arbitary end angles, so yeah I think I know what I need to do there. Can you take a look at the transform code? There are now two Transform functions, but I always get confussed with the SetTexCoords calls.

semlar:
One thing I didn't take into account when I wrote the original script was that the texture in the scroll frame is on a different frame level than the other textures, so if you overlap 2 spinners the wedge on the lower spinner will show up above the textures in the higher spinner. 

I had to create another scroll frame and parent the 4 corner textures to its scroll child to keep them on the same level as the wedge texture.

There might be a better way to do this, and I'm not sure if it's an issue here, but I thought I'd mention it.

InfusOnWoW:
I already had the same problem with two progress textures above each other. I think I added the necessary setFrameLevel call, so I think I can figure that one out.

Stanzilla:
@InfusOnWoW are we done here then or do you want to add something before I merge?

InfusOnWoW:
I'll try do the one thing semlar pointed out with the background, but as far as I'm concerned that could also come in a separate patch afterwards.

Stanzilla:
Alright then!

PR #19, opened 2016-02-13
fix position of centered dynamic groups

InfusOnWoW:
So that it is based on the center of the group rather than the
top/left for VERTICAL/HORIZONTAL dynamic groups

Ticket-number: 467


InfusOnWoW:
http://www.wowace.com/addons/weakauras-2/tickets/467-centered-dynamic-group-animations/#comments

PR #24, opened 2016-02-26
WeakAura Ticket 480

nekizalb:
Add capability to specify a custom trigger combination function that will be run to decide whether the overall aura is triggered based on the status of the individual triggers. This enhances the current funcitonality of ANY or ALL triggers to allow complex logic, such as "Trigger 1 AND (Trigger 2 OR Trigger 3)"

http://pastebin.com/Dw1nVVJs is an example aura using this change.

Additionally, this change REMOVES the checks in Tranmission.lua:scamCheck that only showed code int the code review if the code was enabled. With this change, if their is custom code present in the WA when transmitted, the code is shown.

remove checks in scamCheck that only showed enabled code. All code is shown in the transmitted aura

http://pastebin.com/ki2heZzx is an example of an aura that (in today's client) would show as having no code, but after this change, exposes it has a large number of functions defined, but disabled.

### DIFF
--- a/Options/WeakAurasOptions.lua
+++ b/Options/WeakAurasOptions.lua
@@ -4728,8 +4728,54 @@ function WeakAuras.ReloadTriggerOptions(data)
       order = 0,
       hidden = function() return not (data.additional_triggers and #data.additional_triggers > 0) end,
       values = WeakAuras.trigger_require_types,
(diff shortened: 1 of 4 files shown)


nekizalb:
Sorry for recreating this... was trying to clean up my repo after you merged 482, and deleted the wrong branch by mistake. No changes from the previous version

PR #30, opened 2016-03-20
WeakAuras: Prepare for StopMotion Addon

InfusOnWoW:
There are 3 parts to that preparation:
- In Transmission.lua change how the auras are created on login, basically some code moves from ADDON_LOADED to PLAYER_LOGIN. That allows other addons to register their region types in their ADDON_LOADED handler and be ready before the regions are created.

- Improve the code that deals with unknown region types. Regions that are not supported are displayed as text auras with a informative text message.

- Make the texture selector useable for the StopMotion addon, by allowing that addon to pass in the textures table and allow it to display it's special textures.

If I uploaded the current state of my StopMotion addon to: https://onedrive.live.com/redir?resid=6F73AA556E38CB8B%2154583

I do have plans for some additional features and more textures, but I think I'm done with a initial version. And I'd like to have this patch in WA, so that I can release the first version.

### DIFF
--- a/Options/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/Options/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -1,4 +1,4 @@
-local Type, Version = "WeakAurasDisplayButton", 22
+local Type, Version = "WeakAurasDisplayButton", 23
 local AceGUI = LibStub and LibStub("AceGUI-3.0", true)
 if not AceGUI or (AceGUI:GetWidgetVersion(Type) or 0) >= Version then return end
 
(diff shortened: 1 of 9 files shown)


Stanzilla:
Looks good so far, 
1) "Region type %s not supported" is the only new translation string, correct?
2) If you change AceGUI widgets, you also have to increase the minor version number at the top of the file
3) I will probably merge this on thursday when Asia got 6.2.4 and I tagged after that.

InfusOnWoW:
There's also  L["This region of type \"%s\" is not supported."], which is the text that is shown in the options dialog. 

Ah, didn't realize that with the version, do you want me to upload a new version?

Stanzilla:
Would be cool yeah, just force push to update the PR. I have the changes for 6.2.4 ready, just saying so we don't duplicate work.

InfusOnWoW:
I've updated it. Thanks.

Stanzilla:
You forgot the minor in AceGUIWidget-WeakAurasTextureButton.lua :smiley:

InfusOnWoW:
Ah, yes, I've updated that too now. (Does github send you notifications on updating the pull request? Or only on comments?)

Stanzilla:
Comments but I usually have it open anyway...

InfusOnWoW:
Btw, I've now created http://www.wowace.com/addons/weakauras-stop-motion/ . I'll upload my addon in the next days and see if everything works. Also I've already created a "announcement" video: https://www.youtube.com/watch?v=ybhrGvPWEcc .

Stanzilla:
Nice, looking good!

PR #31, opened 2016-03-26
Smallpatches

InfusOnWoW:
Fix various bugs

### DIFF
--- a/Options/RegionOptions/aurabar.lua
+++ b/Options/RegionOptions/aurabar.lua
@@ -456,6 +456,14 @@ local function createOptions(id, data)
             disabled = function() return not data.spark end,
             hidden = function() return not data.spark end,
         },
+        sparkHidden = {
+            type = "select",
+            values = WeakAuras.spark_hide_types,
+            name = L["Hide on"],
+            order = 45.3,
+            disabled = function() return not data.spark end,
+            hidden = function() return not data.spark end,
+        },
 		border_header = {
 			type = "header",
 			name = L["Border Settings"],
--- a/Types.lua
+++ b/Types.lua
@@ -257,9 +257,16 @@ WeakAuras.inverse_point_types = {
   LEFT = "RIGHT",
   CENTER = "CENTER"
 };
+
 WeakAuras.spark_rotation_types = {
     AUTO = L["Automatic Rotation"],
     MANUAL = L["Manual Rotation"]
+}
+
+WeakAuras.spark_hide_types = {
+  NEVER = L["Never"],
+  FULL  = L["Full"],
+  EMPTY = L["Empty"]
 }
 
 WeakAuras.containment_types = {
--- a/WeakAuras.lua
+++ b/WeakAuras.lua
@@ -1810,6 +1810,28 @@ function WeakAuras.Add(data)
   -- end
 end
 
+
+function removeSpellNames(data)
+    local trigger
+    for triggernum=0,(data.numTriggers or 9) do
+        if(triggernum == 0) then
+            trigger = data.trigger;
+        elseif(data.additional_triggers and data.additional_triggers[triggernum]) then
+            trigger = data.additional_triggers[triggernum].trigger;
+        end
+        if (trigger.spellId) then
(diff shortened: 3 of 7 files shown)


Stanzilla:
nice round of fixes, will test in game and then push to wowace if all goes well.

Stanzilla:
http://i.imgur.com/uaGvcok.png Do we have a problem here? It's the AS tracker from here http://howtopriest.com/viewtopic.php?t=6059

Stanzilla:
Also maybe we should change the text of the placeholder message to something like `This comment is used as separator of this group's triggers, do not remove! ` or something like that? Oh and I guess we have to localize it, too. And add a space before the trigger name.

InfusOnWoW:
It should work in dissecting the text, even if there are other comments. And yeah, if you feel that the comment isn't strong enough, we can do that. (Feel free to just update that patch, or to push a patch on top.)

The other patches looked good to you?

Stanzilla:
Yep, I just broke the PR by force pushing though :D

InfusOnWoW:
Right that error isn't new, that's due to the edit frame also trying to load the text constantly to check for syntax errors. I'm not sure what to do for that, so I left it as is...

Stanzilla:
Okay

InfusOnWoW:
That is, the correct way would be to split it up and check each block by itself, but I think that's for another day. I think this is a good improvement though, in that it makes it actually possible to do bulk edits.

Stanzilla:
I agree

Options/RegionOptions/aurabar.lua
```diff
@@ -456,6 +456,14 @@ local function createOptions(id, data)
             disabled = function() return not data.spark end,
             hidden = function() return not data.spark end,
         },
+        sparkHidden = {
+            type = "select",
+            values = WeakAuras.spark_hide_types,
+            name = L["Hide on"],
```
Stanzilla: New or re-used translation(s)?

Types.lua
```diff
@@ -257,11 +257,18 @@ WeakAuras.inverse_point_types = {
   LEFT = "RIGHT",
   CENTER = "CENTER"
 };
+
 WeakAuras.spark_rotation_types = {
     AUTO = L["Automatic Rotation"],
     MANUAL = L["Manual Rotation"]
 }
 
+WeakAuras.spark_hide_types = {
```
Stanzilla: New or re-used translation(s)?

WeakAuras.lua
```diff
@@ -1810,6 +1810,28 @@ function WeakAuras.Add(data)
   -- end
 end
 
+
+function removeSpellNames(data)
```
Stanzilla: WA.lua is a 2-spaces indented file, might have to change that here, too when you paste it in

Options/RegionOptions/aurabar.lua
InfusOnWoW: New

Types.lua
InfusOnWoW: New too

WeakAuras.lua
InfusOnWoW: Ah, it would be nice if all the files would follow one indentation..

I'll wait until you are through with the review and then update all of the patches, okay?

Stanzilla: Yeah I know, we should decide on one standard one day, I think only transmission.lua is 4 spaces, what do you prefer?

InfusOnWoW:
Me, personally prefer 4 spaces, but the majority is I think 2 spaces, so that should be it. I just hate tabs.

Ah you already merged the first version, so shall I just upload a patch which fixes your comments then?

Stanzilla:
I was lazy and just merged it to test it locally but since force pushed the old state so you'd need to do a new PR with the changes.

InfusOnWoW:
Ah okay, then just tell me once you are done with reviewing. :) Or if you want to have the pull request split up between patches you think are ready and patches which might need some more work. :)

Stanzilla:
I'm done!

InfusOnWoW:
Yeah, I'll use that one.

PR #32, opened 2016-03-28
Patches

InfusOnWoW:
A round of fixes and small improvements for various issues.

### DIFF
--- a/Options/WeakAurasOptions.lua
+++ b/Options/WeakAurasOptions.lua
@@ -7093,7 +7093,7 @@ function WeakAuras.CreateFrame()
     self.oldOnTextChanged = texteditorbox.editBox:GetScript("OnTextChanged");
     texteditorbox.editBox:SetScript("OnTextChanged", function(...)
       local str = texteditorbox.editBox:GetText();
-      if not(str) then
+      if not(str) or texteditorbox.combinedText == true then
         texteditorError:SetText("");
       else
         local _, errorString
@@ -7126,13 +7126,16 @@ function WeakAuras.CreateFrame()
         if not(combinedText == "") then
           combinedText = combinedText.."\n\n";
         end
-        combinedText = combinedText.."--"..childId.."\n";
+
+        combinedText = combinedText.. L["-- Do not remove this comment, it is part of this trigger: "] .. childId .. "\n";
         combinedText = combinedText..(text or "");
       end
       if(sameTexts) then
         texteditorbox:SetText(singleText or "");
+        texteditorbox.combinedText = false;
       else
         texteditorbox:SetText(combinedText);
+        texteditorbox.combinedText = true;
       end
     else
       if(addReturn) then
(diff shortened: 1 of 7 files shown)


Options/WeakAurasOptions.lua
```diff
@@ -7126,13 +7126,21 @@ function WeakAuras.CreateFrame()
         if not(combinedText == "") then
           combinedText = combinedText.."\n\n";
         end
-        combinedText = combinedText.."--"..childId.."\n";
+
+        local formatString
```
Stanzilla: That looks super weird, should probably be something like 
```lua
L["-- The following part is the\"%s\" trigger, do not remove this comment"]:format(childId)
```

InfusOnWoW: It's intentionally... I need the first and second part of the string later to split it up, and I need thos string fragments to be the same between here and there. This makes it so that the translators can't break that.

Options/WeakAurasOptions.lua
```diff
@@ -7093,7 +7093,7 @@ function WeakAuras.CreateFrame()
     self.oldOnTextChanged = texteditorbox.editBox:GetScript("OnTextChanged");
     texteditorbox.editBox:SetScript("OnTextChanged", function(...)
       local str = texteditorbox.editBox:GetText();
-      if not(str) then
+      if not(str) or texteditorbox.combinedText == true then
```
InfusOnWoW: This is also new, and prevents any errors from showing if a combined text is edited.

Options/WeakAurasOptions.lua
```diff
@@ -7126,13 +7126,21 @@ function WeakAuras.CreateFrame()
...
           combinedText = combinedText.."\n\n";
         end
-        combinedText = combinedText.."--"..childId.."\n";
+
+        local formatString
+           = L["-- The following part is the\""]
+             .. childId
+             .. L["\" trigger, do not remove this comment"] .. "\n";
+
+        combinedText = combinedText.. string.format(formatString, childId);
```
nebularg: This is where it was weird, since there's no replacement to be made?

InfusOnWoW: Ah inedd, that is right, I forgot to remove that, once I realized that I properly don't want to use string.format

Options/WeakAurasOptions.lua
InfusOnWoW: I know that this makes the translators job harder, so maybe we should choose a different message... You decide.

Stanzilla: @InfusOnWoW like this then? 

```lua
L["-- The following part is th e\"%s\" trigger, do not remove this comment!"].. "$%s$"
```

InfusOnWoW: Huh, what does that do? Also I'm really not sure whether translators won't mess that up and make the translation below that is used to splice up the text again different from this one.

Stanzilla: Well you need a string to look for when you are matching so we can just add something at the end that you can look for and keep the translation string sane

InfusOnWoW: Nah, I need to properly split up the string too, to figure out the ids. And I don't want to match random comment lines that are actually part of the aura.

Stanzilla: Well we can make sure that it's something that won't come up usually, something like λ or so

Options/WeakAurasOptions.lua
```diff
@@ -7148,14 +7162,48 @@ function WeakAuras.CreateFrame()
     frame.window = "default";
   end
 
+  local function extractTexts(input, ids)
+    local texts = {};
+
+    local currentPos, id, startIdLine, startId, endId, endIdLine;
+    while (true) do
+      startIdLine, startId = string.find(input, L["-- The following part is the\""], currentPos);
```
InfusOnWoW: So what I'm trying to say is, that I need the splitted comment here. And by reusing the same strings above, I actually ensure that they are translated the same. If I use %s and format, then translaters can mess up, and make the strings different, and thus break this feature. I think the best solution a string, where the id comes at the end.

Stanzilla: @InfusOnWoW 

`-- Do not remove this comment, it is part of this trigger: XXX`
Ok maybe this then?

PR #34, opened 2016-04-09
Fixes for Tickets

InfusOnWoW:
Fix full scan auras resetting their tracked buffs
Fix a error on loading
Add a tooltip to the glyph load condition


InfusOnWoW:
The tooltip has two new localizations:  L["Enter a glpyh ID\n"]; and  L["%s: %s\n"]:format(name, glyphSpell);

PR #37, opened 2016-05-15
Add Trigger States to store information

InfusOnWoW:
Those store all information of the triggers, e.g. duration, icons.
This allows for e.g. showing the information of the second trigger if the
first trigger is not active.

This is the basis of more to come.

Also first evaluate all triggers before applying change, fixing:
Ticket-Nr: 501

### DIFF
--- a/BuffTrigger.lua
+++ b/BuffTrigger.lua
@@ -70,7 +70,6 @@ local specificBosses = WeakAuras.specificBosses;
 local specificUnits = WeakAuras.specificUnits;
 local loaded_auras = WeakAuras.loaded_auras;
 local duration_cache = WeakAuras.duration_cache;
-local clones = WeakAuras.clones;
 
 -- GLOBALS: GameTooltip UNKNOWNOBJECT
 
@@ -296,70 +295,84 @@ do
 end
 WeakAuras.aura_cache = aura_cache;
 
-function WeakAuras.SetAuraVisibility(id, triggernum, data, active, unit, duration, expirationTime, name, icon, count, cloneId, index, spellId)
-  local region;
-  local showClones;
+function WeakAuras.SetAuraVisibility(id, triggernum, cloneId, inverse, active, unit, duration, expirationTime, name, icon, count, index, spellId)
+  local triggerState = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
 
-  if(cloneId) then
-    if(data.numAdditionalTriggers > 0) then
-      showClones = data.region.toShow and true or false;
-    else
-      showClones = true;
-    end
-    region = WeakAuras.EnsureClone(id, cloneId);
-  else
-    region = data.region;
-  end
-
-  region.index = index;
-  region.spellId = spellId;
-
-  local show;
+  local show = false
   if(active ~= nil) then
-    if not(data.inverse and UnitExists(unit)) then
-      show = true;
-    end
-  elseif(data.inverse and UnitExists(unit)) then
+    if not(inverse and UnitExists(unit)) then
+     show = true;
+   end
+  elseif(inverse and UnitExists(unit)) then
     show = true;
   end
 
-  if(show) then
-    if(triggernum == 0) then
-      if(region.SetDurationInfo) then
-        region:SetDurationInfo(duration, expirationTime > 0 and expirationTime or math.huge);
-      end
+  cloneId = cloneId or "";
 
-      WeakAuras.ControlChildren(id);
-
-      duration_cache:SetDurationInfo(id, duration, expirationTime, nil, nil, cloneId);
-      if(region.SetName) then
-        region:SetName(name);
-      end
-      if(region.SetIcon) then
-        region:SetIcon(icon or "Interface\\Icons\\INV_Misc_QuestionMark");
-      end
-      if(region.SetStacks) then
-        region:SetStacks(count);
-      end
-      if(region.UpdateCustomText and not WeakAuras.IsRegisteredForCustomTextUpdates(region)) then
-        region.UpdateCustomText();
-      end
-      WeakAuras.UpdateMouseoverTooltip(region);
-    end
-
-    if(data.numAdditionalTriggers > 0 and showClones == nil) then
-      region:EnableTrigger(triggernum);
-    elseif(showClones ~= false) then
-      region:Expand();
-    end
-  else
-    if(data.numAdditionalTriggers > 0 and showClones == nil) then
-      region:DisableTrigger(triggernum)
-    elseif(showClones ~= false) then
-      region:Collapse();
-    end
+  if (not show and not triggerState[cloneId]) then
+    return false;
   end
-end
+
+  triggerState[cloneId] = triggerState[cloneId] or {};
+  local state = triggerState[cloneId];
+  if (state.index ~= index) then
+    state.index = index;
+    state.changed = true;
+  end
+  if (state.spellId ~= spellId) then
+    state.spellId = spellId;
+    state.changed = true;
+  end
+
+  if (state.show ~= show) then
+    state.show = show;
+    state.changed = true;
+  end
+  if (state.progressType ~= "timed") then
+    state.progressType = "timed";
+    state.changed = true;
+  end
+  if (state.expirationTime ~= expirationTime) then
+    state.resort = true;
+    state.expirationTime = expirationTime;
+    state.changed = true;
+  end
+  if (state.duration ~= duration) then
+    state.duration = duration;
+    state.changed = true;
+  end
+
+  local autoHide = not inverse and duration ~= 0;
+  if (state.autoHide ~= autoHide) then
+    state.autoHide = autoHide;
+    state.changed = true;
+  end
+
+  if (state.name ~= name) then
+    state.name = name;
+    state.changed = true;
+  end
+
+  if (state.icon ~= icon) then
+    state.icon = icon;
+    state.changed = true;
+  end
+
+  if (state.stacks ~= count) then
+    state.stacks = count;
+    state.changed = true;
+  end
+
+  if (state.GUID ~= UnitGUID(unit)) then
+    state.GUID = UnitGUID(unit);
+    state.changed = true;
+  end
+
+  if (state.changed) then
+    return true;
+  end
+  return false;
+ end
 
 local aura_scan_cache = {};
 local aura_lists = {};
@@ -423,6 +436,7 @@ function WeakAuras.ScanAuras(unit)
     -- Iterate over all displays (display lists)
     for id,triggers in pairs(aura_list) do
       -- Iterate over all triggers
+      local updateTriggerState = false;
       for triggernum, data in pairs(triggers) do
         if(not data.specificUnit or UnitIsUnit(data.unit, unit)) then
           -- Filters
@@ -503,15 +517,16 @@ function WeakAuras.ScanAuras(unit)
                 WeakAuras.SetTempIconCache(name, icon);
                 if(data.autoclone) then
                   local cloneId = name.."-"..(casGUID or "unknown");
-                  if(not clones[id][cloneId] or clones[id][cloneId].expirationTime ~= expirationTime) then
-                    WeakAuras.SetAuraVisibility(id, triggernum, data, true, unit, duration, expirationTime, name, icon, count, cloneId, index, spellId);
-                    clones[id][cloneId].expirationTime = expirationTime;
+                  if (WeakAuras.SetAuraVisibility(id, triggernum, cloneId, data.inverse, true, unit, duration, expirationTime, name, icon, count, index, spellId)) then
+                    updateTriggerState = true;
                   end
                   active = true;
                   cloneIdList[cloneId] = true;
                   -- Simply show display (show)
                 else
-                  WeakAuras.SetAuraVisibility(id, triggernum, data, true, unit, duration, expirationTime, name, icon, count, nil, index, spellId);
+                  if (WeakAuras.SetAuraVisibility(id, triggernum, nil, data.inverse, true, unit, duration, expirationTime, name, icon, count, index, spellId)) then
+                    updateTriggerState = true;
+                  end
                   active = true;
                   break;
                 end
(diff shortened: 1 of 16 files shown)


InfusOnWoW:
So I once proposed a similar patch months ago, if you remember. I've reduced the patch to the bare minimum now, but it's still around 2500 lines long.

The basic idea behind this patch is, to no longer directly apply the information that is retrurned from the triggers to the aura, but to store that information and then apply it.

This allows for several interesting new features, though this patch only includes:
- Showing information from the second or third trigger, and switching between triggers depending on which is active. This is part of the patch.

Stuff, where I have experimental patches, that need polishing:
- Storing additional trigger information and allowing the user to use it via new text replacements.
- Custom trigger functions that show clones
- Cloning for generic triggers, e.g. a combat log trigger that shows a clone per combat log event that matches the criteria.

BuffTrigger.lua
```diff
@@ -296,70 +295,84 @@ do
...
 
-  if(cloneId) then
-    if(data.numAdditionalTriggers > 0) then
-      showClones = data.region.toShow and true or false;
-    else
-      showClones = true;
-    end
-    region = WeakAuras.EnsureClone(id, cloneId);
-  else
-    region = data.region;
```
InfusOnWoW: The old code for showClones was rather strange. If you had a aura with: 1.trigger being a multi trigger, 2. trigger being a always inactive trigger and set the combine mode to "OR", then the multi trigger would never show anything. This now works just like I would expect it.

BuffTrigger.lua
```diff
@@ -296,70 +295,84 @@ do
...
+  if (not show and not triggerState[cloneId]) then
+    return false;
   end
 
-  if(show) then
-    if(triggernum == 0) then
-      if(region.SetDurationInfo) then
-        region:SetDurationInfo(duration, expirationTime > 0 and expirationTime or math.huge);
-      end
+  triggerState[cloneId] = triggerState[cloneId] or {};
```
InfusOnWoW: This triggerState here is the basic structure in which all the dynamic information of each trigger is stored, this code here instead of directly modifying the region, stores all the information in that state and finally calls UpdatedTriggerState, which applies it to the region.

BuffTrigger.lua
```diff
@@ -1320,8 +1381,13 @@ function BuffTrigger.Modernize(data)
   end
 end
 
-function BuffTrigger.CanGroupShowWithZero(data)
-  local trigger = data.trigger;
+function BuffTrigger.CanGroupShowWithZero(data, triggernum)
```
InfusOnWoW: Since, now the dynamic information from all triggers can be shown, all the CanHave functions now get a triggernum and return the information of that trigger.

GenericTrigger.lua
```diff
@@ -267,22 +265,71 @@ function ConstructFunction(prototype, trigger, inverse)
...
+  allStates[""] = allStates[""] or {};
+  local state = allStates[""];
+
+  state.show = false;
+  state.changed = true;
 end
 
 function WeakAuras.ActivateEvent(id, triggernum, data)
-  WeakAuras.SetEventDynamics(id, triggernum, data);
-  data.region:EnableTrigger(triggernum);
```
InfusOnWoW: Same for the generic trigger, instead of directly changing the region, the triggerState is changed and then UpdatedTriggerState is called.

RegionTypes/aurabar.lua
```diff
@@ -1152,21 +1159,15 @@ local function modify(parent, region, data)
         if customValue then
       -- Update via custom OnUpdate handler
             if type(customValue) == "function" then
-                local value, total = customValue(data.trigger);
+                local value, total = customValue(region.state.trigger);
```
InfusOnWoW: It's no longer always the first trigger that is active, data.trigger is no longer correct.

Types.lua
```diff
@@ -57,6 +57,9 @@ WeakAuras.trigger_require_types = {
   all = L["All Triggers"],
   custom = L["Custom Function"]
 };
+WeakAuras.trigger_modes = {
+  ["first_active"] = -10,
+};
 WeakAuras.trigger_types = {
```
InfusOnWoW: More are certainly possible, I was considering: Show the x auras with the least remaining time, Show x auras with highest stacks.

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
   tinsert(triggerSystems, triggerSystem);
 end
 
+function WeakAuras.GetTriggerStateForTrigger(id, triggernum)
+  triggerState[id][triggernum] = triggerState[id][triggernum] or {}
+  return triggerState[id][triggernum];
+end
+
+local function startStopTimers(id, cloneId, triggernum, state)
```
InfusOnWoW: Here are the functions that take a triggerState and evaluate them. They are called from UpdatedTriggerState below.

This specific function starts or stops auto hide timers.

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
...
+        timer:CancelTimer(record.handle);
+      end
+      record.handle = nil;
+      record.expirationTime = nil;
+    end
+  end
+end
+
+local function ApplyStateToRegion(id, region, state)
+  region.state = state;
```
InfusOnWoW: This function actually applies the changes to the region.

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
...
+  region:Expand();
+end
+
+-- Fallbacks if the states are empty
+local emptyTable = {}
+local emptyState = {};
+emptyState[""] = emptyTable;
+
+local function applyToTriggerStateTriggers(stateShown, id, triggernum)
+  if (stateShown and not triggerState[id].triggers[triggernum + 1]) then
```
InfusOnWoW: This function maintains a boolean array triggerState[id].triggers that contains per trigger a bool whether that trigger is active.

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
...
+    if (not anyCloneShown) then
+      triggerState[id].triggers[triggernum + 1] = false;
+      triggerState[id].triggerCount = triggerState[id].triggerCount - 1;
+      return true;
+    end
+  end
+  return false;
+end
+
+local function evaluateTriggerStateTriggers(id)
```
InfusOnWoW: And a function interpreting the bool array triggerState[id].triggers. Should the aura be shown?

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
...
+local emptyState = {};
+emptyState[""] = emptyTable;
+
+local function applyToTriggerStateTriggers(stateShown, id, triggernum)
+  if (stateShown and not triggerState[id].triggers[triggernum + 1]) then
+    triggerState[id].triggers[triggernum + 1] = true;
+    triggerState[id].triggerCount = triggerState[id].triggerCount + 1;
+    return true;
+  elseif (not stateShown and triggerState[id].triggers[triggernum + 1]) then
+    -- Check if any other clone is shown
```
InfusOnWoW: We are hiding a state for the trigger x, but there could be clones that are visible. In that case the trigger x is still active.

WeakAuras.lua
```diff
@@ -3138,9 +2980,268 @@ function WeakAuras.RegisterTriggerSystem(types, triggerSystem)
...
+          and triggerState[id].triggerLogicFunc
+          and triggerState[id].triggerLogicFunc(triggerState[id].triggers))
+    ) then
+    return true;
+  end
+  return false;
+end
+
+local toRemove = {};
+function WeakAuras.UpdatedTriggerState(id)
```
InfusOnWoW: This function is called by the BuffTrigger and GenericTrigger after they changed the triggerStates. It takes care of calling the above functions and showing/hiding regions as needed.

BuffTrigger.lua
```diff
@@ -52,6 +52,9 @@ CanHaveTooltip(data)
 GetNameAndIcon(data)
   Returns the name and icon to show in the options
 
+GetAdditionalProperties(data)
+  Returns the a tooltip for the additional properties
+
 ]]--
```
InfusOnWoW: So with this patch, the bufftrigger stores a bit of extra information to the trigger state, namely spellId and the caster name. The caster name was asked for in http://www.wowace.com/addons/weakauras-2/tickets/291-caster-dynamic-text/ . The spellId could be useful for writing custom texts.

This patch thus does two things: a) Add infrastructure to allow for additional states, and b) modifies the buff trigger to actually store the caster and spellId.

BuffTrigger.lua
```diff
@@ -94,8 +97,8 @@ do
 
     -- Test OwnOnly
     if (
-      data.ownOnly == true  and WeakAuras.myGUID ~= acEntry.unitCaster or
-      data.ownOnly == false and WeakAuras.myGUID == acEntry.unitCaster
+      data.ownOnly == true  and WeakAuras.myGUID ~= acEntry.casterGUID or
+      data.ownOnly == false and WeakAuras.myGUID == acEntry.casterGUID
     ) then
```
InfusOnWoW: I renamed acEntry.unitCaster to casterGUID, and added a new unitCaster that is the actual name.

BuffTrigger.lua
```diff
@@ -1558,4 +1580,11 @@ function BuffTrigger.GetNameAndIcon(data, triggernum)
   return name, icon;
 end
 
+function BuffTrigger.GetAdditionalProperties(data, triggernum)
```
InfusOnWoW: This function is called from the options and extends the tooltip shown at the text entries.

GenericTrigger.lua
```diff
@@ -349,10 +368,39 @@ function WeakAuras.ScanEvents(event, arg1, arg2, ...)
       local updateTriggerState = false;
       for triggernum, data in pairs(triggers) do
         if(data.triggerFunc) then
-          if(data.triggerFunc(event, arg1, arg2, ...)) then
-            WeakAuras.ActivateEvent(id, triggernum, data);
-            updateTriggerState = true;
+          local untriggerCheck = false;
```
InfusOnWoW: So this slightly extends how trigger functions work. Trigger functions can declare in their prototype that their first parameter should be the trigger states. Either "all", or the "one" if they only need one state. 

If they have a statesParameter "one", they can store arguments into the state by just adding a "store = true" to that argument, that is what the unit characteristics trigger does.

If they have a statesParamter "all", they also need to have a variable "cloneId", which is the state they want to modify. 

Some of the later patches in the series make use of those new modes too.

Prototypes.lua
```diff
@@ -1786,39 +1786,63 @@ WeakAuras.event_prototypes = {
...
         name = "message",
         init = "arg",
         display = L["Message"],
-        type = "longstring"
-      }
+        type = "longstring",
+        store = true
+      },
+      {
+        name = "name",
```
InfusOnWoW: In the old code, event triggers could not return any dynamic information. With this series, all they need to do to store a bit of information, is to set "store = true" on the parameter and that information is retained in the trigger state.

For the DBM announce trigger, this means the message and icon is stored.

Prototypes.lua
```diff
@@ -1786,39 +1786,63 @@ WeakAuras.event_prototypes = {
...
     type = "status",
     events = {
-      "DBM_TimerUpdate"
+      "DBM_TimerStart", "DBM_TimerStop", "DBM_TimerStopAll", "DBM_TimerUpdate", "DBM_TimerForce"
     },
-    force_events = "DBM_TimerUpdate",
+    force_events = "DBM_TimerForce",
     name = L["DBM Timer"],
-    init = function(trigger)
+    triggerFunction = function(trigger)
```
InfusOnWoW: Instead of Construct Function creating the function, for the DBM triggers I want to just create the trigger function myself. This patch adds that possibility by setting triggerFunction in the prototype.

Prototypes.lua
```diff
@@ -2560,15 +2560,18 @@ WeakAuras.event_prototypes = {
...
           event = event:sub(0, -8);
         end
         if (event == 'CHAT_MSG_TEXT_EMOTE') then
           event = 'CHAT_MSG_EMOTE';
         end
+         local use_cloneId = %s;
       ]];
+      return ret:format(trigger.use_cloneId and "true" or "false");
     end,
+    statesParameter = "all",
```
InfusOnWoW: This patch shows how simple it is for triggers to use the new trigger states.

By setting statesParameter to all, and having a cloneId the chat trigger can have clones.
And by setting "store = true" on the message and sourceName, there are additional text replacements for #message# and #sourceName#.

GenericTrigger.lua
```diff
@@ -660,6 +660,9 @@ function GenericTrigger.Add(data, region)
               end
             end
           end
+          if (trigger.custom_type == "stateupdate") then
+            statesParameter = "full";
+          end
```
InfusOnWoW: Most of this patch just modifies the UI to not show certain elements for the new functionality. This line is one that matters.

Types.lua
```diff
@@ -943,7 +943,8 @@ WeakAuras.rune_specific_types = {
 };
 WeakAuras.custom_trigger_types = {
   ["event"] = L["Event"],
-  ["status"] = L["Status"]
+  ["status"] = L["Status"],
+  ["stateupdate"] = L["Trigger State Updater"]
```
InfusOnWoW: And here is the other.

I'll write the documentation for that.

GenericTrigger.lua
```diff
@@ -617,27 +617,27 @@ function GenericTrigger.Add(data, region)
           end
```
InfusOnWoW: A draft documentation is at: http://www.wowace.com/addons/weakauras-2/pages/custom-triggers-temporary/

GenericTrigger.lua
```diff
@@ -1562,36 +1690,71 @@ do
...
+          return false;
         end
       end
-      if (found and (expirationTime == nil or v.expirationTime < expirationTime)) then
-        expirationTime, duration, icon = v.expirationTime, v.duration, v.icon;
+    end
+    return true;
+  end
+
+  function WeakAuras.GetAllBigWigsTimer()
```
Stanzilla: *Timer**s** I guess?

Prototypes.lua
Stanzilla: Hrm why not make ´store=true´ the default since that is what almost every trigger wants and set ´store=false´ for those who don't? Do we have ones that don't? If yes, why don't they want to store?

Options/WeakAurasOptions.lua
```diff
@@ -4799,6 +4786,31 @@ function WeakAuras.ReloadTriggerOptions(data)
         end
       end
     },
+    activeTriggerMode = {
+      type = "select",
+      name = L["Use information from trigger"],
```
Stanzilla: This might need to be worded differently, what it does is use this triggers icon and name, right? So maybe `Use this trigger's data for display`?

InfusOnWoW: Right, it uses the "dynamic information" from that trigger.

Prototypes.lua
InfusOnWoW: After looking at the other triggers, I would say that in most cases it doesn't make sense to store it and that it should be opt-in.

Stanzilla: ok!

PR #38, opened 2016-05-20
3 small fixes

InfusOnWoW:
Fix the options dialog filter box to not use lua patterns
Add UNIT_DISPLAYPOWER to the trigger events for Power triggers
Clean up OnShow handlers

### DIFF
--- a/Prototypes.lua
+++ b/Prototypes.lua
@@ -803,7 +803,8 @@ WeakAuras.event_prototypes = {
       "PLAYER_TARGET_CHANGED",
       "PLAYER_FOCUS_CHANGED",
       "INSTANCE_ENCOUNTER_ENGAGE_UNIT",
-      "WA_DELAYED_PLAYER_ENTERING_WORLD"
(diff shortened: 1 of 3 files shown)


Prototypes.lua
```diff
@@ -803,7 +803,8 @@ WeakAuras.event_prototypes = {
       "PLAYER_TARGET_CHANGED",
       "PLAYER_FOCUS_CHANGED",
       "INSTANCE_ENCOUNTER_ENGAGE_UNIT",
-      "WA_DELAYED_PLAYER_ENTERING_WORLD"
+      "WA_DELAYED_PLAYER_ENTERING_WORLD",
+      "UNIT_DISPLAYPOWER"
```
InfusOnWoW: So the way to test this is:
Make a aura with a power trigger, set the power type condition to e.g. "mana".
Switch from caster form to druid. (The aura vanishes due to getting rage.)
Wait until all rage is gone, switch back to caster form.
Without this patch, the aura won't reappear, with this patch the aura will be shown.

PR #45, opened 2016-05-27
Make pasting huge import strings lightning fast

InfusOnWoW:
Use code provided by Semlar to make pasting a lot faster.
The edit box no longer shows the contents of the pasted text,
but that's not a huge loss.

### DIFF
--- a/Options/WeakAurasOptions.lua
+++ b/Options/WeakAurasOptions.lua
@@ -6960,10 +6960,10 @@ function WeakAuras.CreateFrame()
         elseif(mode == "table") then
           displayStr = WeakAuras.DisplayToTableString(id);
         end
+        importexportbox.editBox:SetMaxBytes(nil);
         importexportbox.editBox:SetScript("OnEscapePressed", function() importexport:Close(); end);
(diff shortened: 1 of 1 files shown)


Options/WeakAurasOptions.lua
```diff
@@ -6960,30 +6960,45 @@ function WeakAuras.CreateFrame()
...
+      local function clearBuffer(self)
+        self:SetScript('OnUpdate', nil)
+          local pasted = strtrim(table.concat(textBuffer))
+          importexportbox.editBox:ClearFocus();
+          pasted = pasted:match( "^%s*(.-)%s*$" );
+          if (#pasted > 20) then
+            WeakAuras.ImportString(pasted);
+            importexportbox:SetLabel(L["Processed %i chars"]:format(i));
+            importexportbox.editBox:SetMaxBytes(10);
+            importexportbox.editBox:SetText("Done");
```
InfusOnWoW: So Semlar did something different in his code, the truncates the string at 2500 chars and shows those 2500 via a fontstring overlayed. I found that more confussing then helpful, so I did something different.

PR #50, opened 2016-07-01
Trigger Templates

InfusOnWoW:


Options/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -679,7 +679,7 @@ local methods = {
         self.frame.description = {...};
     end,
     ["SetIcon"] = function(self, icon)
-        if(type(icon) == "string") then
+        if(type(icon) == "string" or type(icon) == "number") then
```
Stanzilla: file needs a minor bump

Options/AceGUI-Widgets/AceGUIWidget-WeakAurasNewButton.lua
```diff
@@ -39,7 +39,7 @@ local methods = {
         self.frame:SetScript("OnClick", func);
     end,
     ["SetIcon"] = function(self, icon)
-        if(type(icon) == "string") then
+        if(type(icon) == "string" or type(icon) == "number") then
```
Stanzilla: file needs a minor bump

Options/WeakAurasOptions.lua
```diff
@@ -7375,15 +7428,6 @@ function WeakAuras.CreateFrame()
   codereviewCancel:SetWidth(100);
   codereviewCancel:SetText(L["Okay"]);
 
-  function codereview.Close(self)
```
Stanzilla: we don't need that anymore because of newView, right?

Templates/WeakAurasTemplates.toc
```diff
@@ -0,0 +1,25 @@
...
+## Title: WeakAuras Templates
+## Author: Infus
+## Version: @project-version@
+## Notes: Templates for WeakAuras
+## X-Category: Interface Enhancements
+## DefaultState: Enabled
+## LoadOnDemand: 1
+## Dependencies: WeakAuras, WeakAurasOptions
+
+Localization-enUS.lua
```
Stanzilla: locale files are missing atm, just copy paste them over from Tutorials and replace `Tutorials` with `Templates` in them, I will hook it up with the translation system then

Stanzilla:
Needs a rebase to get rid of the two extra commits

Stanzilla:
Definitely needs a "back" button in the navigation

Options/WeakAurasOptions.lua
InfusOnWoW: For some reason there were two codereview.Close() functions. There should be one left just a few lines later.

Stanzilla: Ah ok, makes sense. It should behave the same now when the second one already came after this one

InfusOnWoW:
So I agree on all of those comments.

I'm not sure where to put the druid shapeshift trigger though. That's why the section was called Other. But yeah Ressources is better for all the other classes. I don't think the Shapeshift trigger fit into either Buffs/Debuffs or Cooldowns too well. So for now for Druids the secion is called Ressources and Shape Shift Form.

PR #95, opened 2016-08-20
activating aura environment when aura triggers custom logic

pwfff:
Something like this custom trigger currently only works when none of the triggers used are of the 'aura' type:
```
function(values)
    if values[2] and values[3] and values[4] then
        ActionButton_ShowOverlayGlow(WeakAuras.regions[aura_env.id].region)
    else
        ActionButton_HideOverlayGlow(WeakAuras.regions[aura_env.id].region)
    end
    
    return values[1]
end
```

My patch fixes this by activating the environment in evaluateTriggerStateTriggers. I tested it for my broken case and everything seems to be working: https://wago.io/EkwSOOpKW

### DIFF
--- a/WeakAuras.lua
+++ b/WeakAuras.lua
@@ -3230,15 +3230,18 @@ local function applyToTriggerStateTriggers(stateShown, id, triggernum)
 end
 
 local function evaluateTriggerStateTriggers(id)
+  local result = false;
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
Looks fine to me.

PR #100, opened 2016-09-14
Wrapping the WoW API for testability.

antoinne85:
**EDIT:** A lot has changed since this was written. You can safely ignore most of this hunk of text. The gist of is:
Hey, why don't we mock out a bunch of the WoW API so we can write some unit tests and make the addon easier to develop?

---

Let me start by saying _**This PR is NOT ready to merge.**_ 

It's just intended to start a dialog about what I've been working on and to get some feedback. Additionally, in order for me to verify that the work hasn't broken anything, to get some information about how you normally test/release the addon.

I've added several comments to the code to highlight the more interesting bits. You'll probably want to click "View full changes" on them to get a good look at what's going on.

There are likely a lot more API calls that are being used that need to be added, but this is enough to get the conversation going.
# What's in here?

The gist of the change is to access a wrapper for the WoW API when you need to call it instead of calling it directly. This would allow us to mock/fake the Wow API to start getting some tests around some of the features (to pave a safer way to tackle more difficult enhancements, as well as keep the quality high).
## The "Live" Wrapper

So, for example, instead of calling:

``` lua
local UnitIsDead = UnitIsDead
local isDead = UnitIsDead("player")
```

You would do something like:

``` lua
local apiWrapper = WowApiWrapper:create()
local isDead = apiWrapper.UnitIsDead("player")
```

The API Wrapper would then call the WoW API for you and give the result.

You don't have to make a new wrapper every time you want to call the API. In fact, a new wrapper is only created the first time you ask for one. Every other time you ask for one you get the one that already exists. This should make calling WowApiWrapper:create() a cheap operation.

Furthermore, if you _really_ wanted to have the methods be local, you could continue with the existing practice:

``` lua
local UnitIsDead = apiWrapper.UnitIsDead
local isDead = UnitIsDead("player")
```

The advantage to this is that once we're not calling the WoW API directly we've started to reduce the need for the game to be running to test functionality and can begin writing automated tests to reproduce bugs and maintain a fully functional AddOn.
## The Mock Wrapper

In addition to WowApiWrapper.lua, there's also WowApiMockWrapper.lua.

This variant exposes the same functions as WowApiWrapper, but it also exposes more functions.

I can use the WowApiMockWrapper to write code like:

``` lua
local apiWrapper = WowApiMock.create()
apiWrapper.MockGetScreenWidth(1680)
local width = apiWrapper.GetScreenWidth()
print(width) -- 1680
```

``` lua
local apiWrapper = WowApiMock.create()
apiWrapper.MockLoadAddOn({1,"Successful Load"})
local success, reason = apiWrapper.LoadAddOn("WeakAuras2")
print(success) -- 1
print(reason) -- Successful Load
```

``` lua
local apiWrapper = WowApiMock.create()

local loadAddOnFunc = funtcion(addOnName)
    if addOnName == "WeakAuras2" then
       return {1, "Successful Load"}
    else
        return {nil, "BANNED"}
    end
end

apiWrapper.MockLoadAddOn(loadAddOnFunc)

local success, reason = apiWrapper.LoadAddOn("WeakAuras2")
print(success) -- 1
print(reason) -- Successful Load

success, reason = apiWrapper.LoadAddOn("SomethingElse")
print(success) -- nil
print(reason) -- BANNED
```
# Questions

How do you normally test your changes? Do you take the files from your local copy of the code and simply drop them into your addons folder? Which files do you move and what goes where?

I'm pretty new to Lua, so there might be some finer points that I don't yet know about, so feel free to point any of that out.

Where does the team normally communicate with one another about code changes, enhancements, bug fixes, etc? I didn't see a Discord channel for dev-related stuff.


.gitignore
```diff
@@ -2,3 +2,4 @@ Locales/WeakAuras.lua
 Locales/WeakAuras_Options.lua
 Locales/WeakAuras_Tutorials.lua
 Locales/WeakAuras_Templates.lua
+.vscode/launch.json
```
antoinne85: I've been working in VSCode, this just keeps the VSCode debugging settings file out of the repository.

GenericTrigger.lua
```diff
@@ -63,7 +63,7 @@ local tostring, select, pairs, type = tostring, select, pairs, type
 local error, setmetatable = error, setmetatable
 
 -- WoW APIs
-local IsPlayerMoving = IsPlayerMoving
+local apiWrapper = WowApiWrapper:create()
```
antoinne85: This kind of change is in a lot of the files.

I could have done the assignment like:
```
local apiWrapper = WowApiWrapper:create()
local IsPlayerMoving = apiWrapper.IsPlayerMoving
```

But that makes it more difficult to find usages of the WoW API.

If everything went through the wrapper then you could fairly easily do a "Find All" for "apiWrapper" to find every access to the WoW API throughout the whole project.

WowApiWrapper.lua
```diff
@@ -0,0 +1,82 @@
...
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "IsSpellKnown")
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "GetTalentInfo")
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "GetPvpTalentInfo")
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "GetAddOnEnableState")
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "IsPlayerMoving")
+    WowApiWrapper:ConnectGlobalFunction(wrapper, "UnitIsDead")
+end
+
+function WowApiWrapper:ConnectGlobalFunction(wrapper, name)
+    wrapper[name] = _G[name]
```
antoinne85: Connects the global function with some name to the API wrapper. This is what makes the API Wrapper just pass-through to the actual in-game function.

WowApiWrapper.lua
```diff
@@ -0,0 +1,82 @@
...
+
+local instance;
+
+function WowApiWrapper:create()
+    -- We don't want to make more than one instance of the API wrapper
+    -- as there's no need.
+
+    -- Instead, store the one we created locally and if someone asks for a new
+    -- one, just give them the one we already have.
+    if not instance then
```
antoinne85: Uses a really simplified singleton style setup to make sure we only ever go through the work of connecting to the global WoW API one time.

WowApiMockWrapper.lua
```diff
@@ -0,0 +1,122 @@
...
+    MockMethodPartial(wrapper, "GetFrameCPUUsage", {100, 20})
+    MockMethodPartial(wrapper, "debugprofilestop", 100)
+    MockMethodPartial(wrapper, "debugstack", "FAKE_DEBUG_STACK")
+    MockMethodPartial(wrapper, "IsSpellKnown", true)
+    MockMethodPartial(wrapper, "GetAddOnEnableState", 2)
+    --MockMethodPartial(wrapper, "IsPlayerMoving", "TODO: Mock IsPlayerMoving")
+    MockMethodPartial(wrapper, "UnitIsDead", nil)
+end
+
+-- TESTING When we change the mock on one instance it doesn't affect others.
```
antoinne85: There are several tests at the end of this file right now to make sure that as I'm modifying the way the mock works I don't break any of the important functionality.

Things tested here are:

- Each mock can be modified without affecting other mocks.
- When the mocked data is a function we call the function with all the right parameters and return the correct values.
- When the mock has multiple return values we make sure they come back the way they're supposed to.
- When the mock returns "nil" we don't lose it along the way.
- When we assign one of the mocked functions into a local variable and later call it using the local it still works.

Utility/Mocks.lua
```diff
@@ -0,0 +1,69 @@
+--- Stubs out a mocked method.
```
antoinne85: This file contains methods used to configure the mocks.

Basically it does some magic to create functions on the mock that will allow you to call into the fake API as well as change the values that the fake API returns.

Znuff:
but why?

InfusOnWoW:
That's the wrong approach, there's no way we want to replace all calls to the wow api with calling a wrapper first. If you want to mock the wow api, then add the mock functions into the enviroment before evaluatin the lua code.

But, the wow environment is massive, there's no end to the amount of functions in it, and the interfaces are ill-defined with various suprising behaviour. Mocking all of that is for all intends and purposes impossible.

antoinne85:
@Znuff WeakAuras is a big project and sufficiently complex that it could benefit from some automated tests. Additionally, having automated tests allows you to make changes with more confidence that you haven't broken anything.

@InfusOnWoW  The reason I went with the wrapper instead of defining all the mock functions globally is because when you go to write the tests, having all of the tests share a single global wrapper will cause the tests to share state (and therefore affect one another) instead of starting from a clean slate.

I suppose you could work around that by having each test call some kind of "reset" method at the beginning of the test to get the global mock state back into a default position. I suppose that's not such a terrible way to go about things. In your opinion, would that address the wrapper/global issue?

As for mocking the WoW environment, you don't have to mock the entire thing--only the functions that are in use in WeakAuras. And it's not attempting to actually replicate the _real_ behavior of the WoW API, just to provide a way to force the API call to return a known value. As far as I can tell, WeakAuras is probably using less than 100 of the total API functions that are available, and you don't have to provide mock data for every function on every test--only the functions you intend to be used.

antoinne85:
I've pushed this a little further. 

@InfusOnWoW, I've taken your advice and set the mocks up in the global scope. 

The running of the tests is handled by `TestHelper.lua` > `RunAllTests()` which makes sure to reset the mock environment before each test. 

I also took the code that creates the `talent_types_specific` and `pvp_talent_types_specifc` and created a reusable "class" out of it and put some tests around it. 

As a side-effect of this work I noticed there are several places in the code that have 6/3 hard-coded for talent tiers/columns instead of using the constants. (One such place is in the existing talent cache code that got pulled into this class.)

antoinne85:
I've tried these change out using the client (I just dumped the relevant changes to `WeakAuras.lua`, `WeakAurasTalents.lua`, and 'WeakAuras.toc` into my addon folder and ran the client.

No errors and the class/spec/talent dropdowns for the "Load" tab still seem to work as expected.

Stanzilla:
I'm super torn on this, on one hand, tests are nice, on the other, I really don't want to change the addon in that way. Not sure what to do yet.

antoinne85:
@Stanzilla When you say "in that way" are you referring to breaking apart functionality into smaller components or another aspect of what's happening in this PR? I'm guessing its the former since that's really the only change that has an impact on the project at runtime--everything else is just dev-time boilerplate and helpers that don't actually affect the code that gets shipped to the user.

As far as breaking it apart, if we truly want to pursue testing as a goal (which is, of course, a decision to be made by the project maintainers, not newcomers like myself) it's likely that lots of pieces of the addon would wind up broken apart in some way or another. At the moment, several of the files are 2000 or 3000 lines long. I know its just one man's opinion, but getting some functionality separated out and independently functional (in a more traditional OOP style) would make the code much easier to read and work on--as well as remove a lot of duplication (which leads to bugs as one section gets updated but not another).

If breaking the code apart into smaller chunks is undesirable that's okay, but I think it makes testing (almost) prohibitively difficult.

PR #115, opened 2016-10-15
Smallpatches

InfusOnWoW:


Types.lua
```diff
@@ -1313,3 +1313,21 @@ WeakAuras.cooldown_progress_behavior_types = {
   showOnReady    = L["Not on cooldown"],
   showAlways     = L["Always"]
 };
+
+WeakAuras.item_slot_types = {
```
Stanzilla: Should be able to use those https://www.townlong-yak.com/framexml/live/Constants.lua#845

PR #119, opened 2016-10-30
Order the texture pickers's okay/cancel button and the textures

InfusOnWoW:
This feels like a Blizzard bug. Apparently the scroll frame no
longer clips the mouse interaction. So increase the okay and cancel
button's frame level to ensure that they are in front.


InfusOnWoW:
Doesn't work...

Stanzilla:
Does this help? http://wow.gamepedia.com/XML_properties#clipChildren

InfusOnWoW:
I think that's independant, though I'll try tomorrow. As is the texture selector is completly unuseable, you can't click on the okay button if the textures extend below it...

PR #120, opened 2016-10-31
Order the texture pickers's okay/cancel button and the textures

InfusOnWoW:
This feels like a Blizzard bug. Apparently the scroll frame no
longer clips the mouse interaction. So increase the okay and cancel
button's frame level to ensure that they are in front.


Stanzilla:
ready to merge or a you still playing around with it?

InfusOnWoW:
Ready to be merged as far as I'm concerned, but needs follow up with guil or thedanw

PR #137, opened 2017-01-13
Update Templates for 7.1.5

InfusOnWoW:
Probably a whole lot missing....


WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -69,7 +69,6 @@ templates.class.WARRIOR = {
         { spell = 188923, type = "buff", unit = "player" }, -- Cleave
         { spell = 197690, type = "buff", unit = "player", talent = 12 }, -- Defensive Stance
         { spell = 202164, type = "buff", unit = "player" }, -- Bounding Stride
-        { spell = 207982, type = "buff", unit = "player", talent = 15 }, -- Focused Rage
```
Stanzilla: Focused Rage is only gone for Prot, Arms still has it

Stanzilla:
Want to do https://www.wowace.com/projects/weakauras-2/issues/826 in the same PR? I think one msg per session would be enough or just none/behind a debug flag?

InfusOnWoW:
Those should only happen once per session. And this update does fix all of those messages so I think it is fixed.

PR #145, opened 2017-01-28
[WIP] Refactor options

jgersti:
### Done
+    move `CreateFrame(...)`  into `Frames\main.lua`  (_~ 2.5k lines_)
+    move `CreateImportButton`, `SortIMportButton`, `EnableAddonDispaly`, `CollisionResolved` and `DisableAddonDisplay` into `addon.lua` (_~ 350 lines_)
+    move  `StartFrameChooser` and `StopFrameChooser` into `Framea\FrameChooser.lua` (_~ 100 lines_)

### Note:
Ignore the first commit.  
The merge commits keep the history intact. __DO NOT REMOVE__

_edit_: I removed the merged branches for now so i can edit them easier.

PR #147, opened 2017-02-05
Refactor external addon support

jgersti:
see #145


InfusOnWoW:
Looks fine to me now. I'm fine with merging that.

There's some tiny improvement I see, that is WeakAuras.Delete uses WeakAuras.importDisplayButtons. If that's abstracted behind a function WeakAuras.DeleteImportButton then importDisplayButtons could be made local to ExternalAddons.lua

WeakAuras.DisableAddonDisplay() could be made local too, as it is used only inside the file.

And WeakAuras.SortImportButtons() can also be made local with a trivial change.

Anyway, I'm happy to have this merged as is.

PR #148, opened 2017-02-05
Smallpatches

InfusOnWoW:


WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -937,7 +937,7 @@ local methods = {
                     self.frame:SetScript("OnClick", nil)
                 end
             end
-        else
+        elseif (self.dragging) then
```
InfusOnWoW: If I read that code correctly, data is nil only iff SetDragging is called on HideOptions(). In that case we only need to reset the button that was actually being dragged.

jgersti: while it is only called in `HideOptions()`, this branch needs to enable and restore `OnClick` on _all_ displays not only on the dragged elements the `Drop()` can be restricted to only be called when dragged.

InfusOnWoW: Right, that's already enough. The problem is that this was previously calling UpdateButtonsScroll() for each aura.

PR #163, opened 2017-03-03
Improve GetUnifiedSpellCooldown

InfusOnWoW:


InfusOnWoW:
I haven't tested that yet. But that's imho how it should work. One function that wraps Blizzards apis and handles all cases.

This should improve charges for spells that don't actually have charges.

Stanzilla:
That shows a "1" as cooldown on every aura that is Cooldown Progress(spell) and Display "Always" that is not on cooldown, here.

PR #164, opened 2017-03-04
fix wrong argument order in GetBigWigsTimer

jcai1:
This is a non-breaking change that fixes the argument order in WeakAuras.GetBigWigsTimer.


InfusOnWoW:
Thank you

PR #173, opened 2017-03-18
Fixed two very small typos

juaoose:


Stanzilla:
Hey there! Thanks for your PR but those notes are currently not used, translation is done via the localization tool at Wowace, see here https://www.wowace.com/projects/weakauras-2/localization

PR #176, opened 2017-03-21
Conditions

InfusOnWoW:


InfusOnWoW:
The main thing missing is that the options don't work if multiple auras are selected. Otherwise it should work.

InfusOnWoW:
Numbering the variables for different triggers with "1." and "2." is confussing.
Duration as a variable could refer to the remaning duration or the total duration

Stanzilla:
You could use the line that says "remove this line" atm to say Trigger 1, 2 etc so it acts as a trigger separator

Stanzilla:
Mockup: https://i.imgur.com/1YZJmCT.png

(1) is with a conditions tab like current version (2) is with the conditions on the trigger tab

InfusOnWoW:
The test size property needs to fetch a font.

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -163,14 +163,14 @@ local function filterUsedProperties(indexToProperty, allDisplays, usedProperties
...
 local function addControlsForChange(args, order, data, conditions, i, j, allProperties, usedProperties)
-  local thenText = (j == 1) and L["Then "] or "and ";
+  local thenText = (j == 1) and L["Then "] or L["And "];
 
   local display = isSubset(data, conditions[i].changes[j]) and allProperties.displayWithCopy or allProperties.display;
   local valuesForProperty = filterUsedProperties(allProperties.indexToProperty, display, usedProperties, conditions[i].changes[j].property);
 
   args["condition" .. i .. "property" .. j] = {
     type = "select",
-    name = thenText .. blueIfSubset(data, conditions[i].changes[j]) .. L["Property"],
```
InfusOnWoW: The reason there's the "Property" string, is so that it can be colored blue if the property is only in some auras.

PR #183, opened 2017-04-17
Smallpatches

InfusOnWoW:


InfusOnWoW:
I think we should wait a bit before merging this. So that whatever problems show up can be attributed to the conditions merge.

The Pet Spell fix is maybe worth merging immediately, as that is pretty bad. The rest are new features

PR #191, opened 2017-05-01
Add a sound RegionType

Mpstark:
Currently, the support of sounds in WeakAuras is fairly barebones; sounds can be played either when an aura is shown or when an aura is hidden, and the sound will always play all the way through, ignoring the current state of the containing aura. For many people, including myself, this means that they have many empty text displays (that is that the display text is a space: ' ') that use the sound actions and that solutions to common desires, like canceling sounds, looping, and countdowns are out-of-reach without custom-code solutions or many, many duplicate auras. I believe that an introduction of a "Sound" display type/`RegionType` can alleviate most of these issues.

In this pull-request, I have implemented a basic "Sound" `RegionType` that allows for sounds to be played and optionally looped while the region is 'active', and interrupted/canceled when the region is 'hidden'. All of the changes are self-contained and were originally implemented as a plugin.

This type of implementation has a drawback, however. As this is effectively a display-less display type, it presents the same issue that empty text displays have, in that the position of the display doesn't matter and inclusion into a `DynamicGroup` doesn't make much sense. While moving using the `MoverSizer` is possible, I've excluded the standard positioning menu because it doesn't affect the display at all. Some modifications could be made to underlying systems to better support this type of 'display'.

EXAMPLE AURAS:
---

Using a custom-code trigger for `IsSpellInRange` along with some extra triggers for ensuring the target is hostile, alive, and attackable. The 'display' is configured to play a sonar ping sound on repeat. In effect, while the player is targeting a valid target and out of range, the sound keeps going off. When I come into range, the sound immediately shuts off.

Using a spell cooldown trigger, a sound can be made to play on repeat while an important rotational button on a medium cooldown is available.

A long sound, such as a song, can be played while a longish buff is active. The song will be shut off as soon as the cooldown is over. Suggestion: Miley Cyrus' The Climb while Surrender to Madness is triggered.

FURTHER IMPROVEMENTS
---

As this is kind of a proof-of-concept, there is sizable room for additional improvements and features. Countdown sounds, such as those in DBM or Bigwigs could be implemented tying into Duration Info. Additional options, such as a ramp up/down repeating time (i.e. the longer it goes off, the faster/slower it goes off). Properties (such as repeat time) could be exposed to the Conditions feature.

### DIFF
--- a/WeakAuras/RegionTypes/Sound.lua
+++ b/WeakAuras/RegionTypes/Sound.lua
@@ -1,0 +1,101 @@
+local SharedMedia = LibStub("LibSharedMedia-3.0");
+
+local default = {
+  selfPoint = "BOTTOM",
+  anchorPoint = "CENTER",
+  anchorFrameType = "SCREEN",
+  xOffset = 0,
+  yOffset = 0,
+  frameStrata = 1,
+  
+  sound = "",
+  sound_channel = "Master",
+  sound_path = "",
+  sound_kit_id = "",
+  
+  shouldRepeat = false,
+  repeatTime = 1,
+}
+
+local function create(parent)
+  local region = CreateFrame("FRAME", nil, parent);
+  region:SetMovable(true);
+  
+  return region;
+end
+
+local function modify(parent, region, data)
+  data.width = 16;
+  data.height = 16;
+  region:SetWidth(data.width);
+  region:SetHeight(data.height);
+  
+  WeakAuras.AnchorFrame(data, region, parent);
+  
+  local function killTicker()
+    if (region.ticker) then
+      region.ticker:Cancel();
+      region.ticker = nil;
+      region.tickerCallback = nil;
+    end
+  end
+  
+  killTicker();
+  
+  local function playSound()
+    if (not WeakAuras.IsOptionsOpen()) then
+      if(data.sound == " custom") then
+        if(data.sound_path) then
+          local _, soundHandle = PlaySoundFile(data.sound_path, data.sound_channel or "Master");
+          region.soundHandle = soundHandle;
+        end
+      elseif(data.sound == " KitID") then
+        if(data.sound_kit_id) then
+          local _, soundHandle = PlaySoundKitID(data.sound_kit_id, data.sound_channel or "Master");
+          region.soundHandle = soundHandle;
+        end
+      else
+        local _, soundHandle = PlaySoundFile(data.sound, data.sound_channel or "Master");
+        region.soundHandle = soundHandle;
+      end
+    end
+  end
+
+  local function stopSound()
+    if (region.soundHandle) then
+      StopSound(region.soundHandle);
+      region.soundHandle = nil;
+    end
+  end
+  
+  local function OnShow()
+    if (region.toShow) then
+      -- immediately play the sound
+      playSound();
+      
+      -- setup a callback every 'repeatTime' seconds if repeating
+      if (data.shouldRepeat) then
+        region.tickerCallback = function()
+          stopSound();
+          playSound();
+        end
+        
+        region.ticker = C_Timer.NewTicker(data.repeatTime, region.tickerCallback);
+      end
+    end
+  end
+  
+  region:SetScript("OnShow", OnShow);
+  
+  local function OnHide()
+    -- immediately stop playing sound
+    stopSound();
+    
+    -- kill our ticker if it exists
+    killTicker();
+  end
+  
+  region:SetScript("OnHide", OnHide);
+end
+
+WeakAuras.RegisterRegionType("sound", create, modify, default);
--- a/WeakAuras/RegionTypes/Sound.lua
+++ b/WeakAuras/RegionTypes/Sound.lua
@@ -15,6 +15,9 @@ local default = {
   
   shouldRepeat = false,
   repeatTime = 1,
+  
+  shouldDelay = false,
+  delayTime = 1,
 }
 
 local function create(parent)
@@ -32,15 +35,14 @@ local function modify(parent, region, data)
   
   WeakAuras.AnchorFrame(data, region, parent);
   
-  local function killTicker()
-    if (region.ticker) then
-      region.ticker:Cancel();
-      region.ticker = nil;
-      region.tickerCallback = nil;
+  local function killTimer()
+    if (region.timer) then
+      region.timer:Cancel();
+      region.timer = nil;
     end
   end
   
-  killTicker();
+  killTimer();
   
   local function playSound()
     if (not WeakAuras.IsOptionsOpen()) then
@@ -68,19 +70,28 @@ local function modify(parent, region, data)
     end
   end
   
+  local function startPlaying()
+    -- immediately play the sound
+    playSound();
+    
+    -- setup a callback every 'repeatTime' seconds if repeating
+    if (data.shouldRepeat) then
+      local function callback()
+        stopSound();
+        playSound();
+      end
+      
+      region.timer = C_Timer.NewTicker(data.repeatTime, callback);
+    end
+  end
+  
   local function OnShow()
     if (region.toShow) then
-      -- immediately play the sound
-      playSound();
-      
-      -- setup a callback every 'repeatTime' seconds if repeating
-      if (data.shouldRepeat) then
-        region.tickerCallback = function()
-          stopSound();
-          playSound();
-        end
-        
-        region.ticker = C_Timer.NewTicker(data.repeatTime, region.tickerCallback);
+      if (not data.shouldDelay) then
+        startPlaying();
+      else
+        -- wait for the timer to fire, then start
+        region.timer = C_Timer.NewTimer(data.delayTime, startPlaying);
       end
     end
   end
@@ -88,11 +99,11 @@ local function modify(parent, region, data)
   region:SetScript("OnShow", OnShow);
   
   local function OnHide()
-    -- immediately stop playing sound
+    -- immediately stop playing sound if we're playing one
     stopSound();
     
-    -- kill our ticker if it exists
-    killTicker();
+    -- kill our ticker or timer if they exist
+    killTimer();
   end
   
   region:SetScript("OnHide", OnHide);
--- a/WeakAuras/WeakAuras.toc
+++ b/WeakAuras/WeakAuras.toc
@@ -35,3 +35,4 @@ RegionTypes\Text.lua
 RegionTypes\Group.lua
 RegionTypes\DynamicGroup.lua
 RegionTypes\Model.lua
+RegionTypes\Sound.lua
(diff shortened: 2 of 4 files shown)


Mpstark:
As with other drive-by contributions, I see this as the start of the conversation about this. Feel free to point out problems and barriers to inclusion.

If yall aren't comfortable including this in WeakAuras proper, I would be fine making this a standalone plugin, though I believe that it would be better to not split the WA string sharing community too much.

Stanzilla:
I'm not sure this is the correct way to tackle that problem, implementing this as a new RegionType would make the sound options on the other RegionTypes a second class citizen because they offer less options than the new RegionType. Which would mean that everything you add/change about sound handling would have to be backported to the other region's sound options as well and then we would have a hard time justifying where it even is its own RegionType again. But I'm open to suggestions and discussion, thank you for your contribution so far :)

Mpstark:
The reason that I felt that a new `RegionType` was the way to go was that the Actions features have no lifecycle and are primarily meant for side-effect type behavior. Take the simplest improvement of the pull-request over the current system: using `StopSound` to cancel an already playing sound. How would you implement such a feature within the current Actions system?

In effect, it's a *On-Hide* action for an already existing *On-Show* action or vice-versa. Just making it so that **all** *On-Show* sounds are stopped *On-Hide* would be a breaking change since people expect their auras to have the [same behavior as when they made them](https://xkcd.com/1172/). Heck, I have an aura that would be broken by this change (a metronome tick sound when casting particular spells, not all on GCD). Implementing it as an option is a little troubling as well, where do you put the checkbox? I suppose that you could add a checkbox for 'Cancel Sound if Hidden/Shown', but again, this adds a lifecycle to the Actions feature and dilutes its simplicity.

The other large improvement in this pull-request, looping/repetition is equally hard to add to the existing system, because it's even more of a paradigm shift, now potentially hidden auras could have lasting effects. Even worse, actions trigger when auras are loaded/unloaded (which is an often sore spot for new users that load auras for available cooldowns in combat and all of their attached sounds play -- obviously this RegionType has a similar issue, except that it should be obvious that this will occur), which means that side effects could persist through an aura not being loaded anymore.

Ultimately I think that introducing this, admittedly odd, "`RegionType`" is the easiest way, both conceptionally and practically, to achieve these features. I suppose that I might be in the minority with my many text auras with no actual text, with custom code triggers pulsing the triggers on-and-off to achieve looping sound effects. The improvement over my previously hackish solutions is really very noticeable.

Stanzilla:
Good points!

InfusOnWoW:
So I tryed it out yesterday and had to think a bit about it:

I do know that I want to add some actions to the conditions. But I don't have it eniterly mapped out on how to do.

And for further context, I think most users that use sound (which are properly already a minority), don't use it as extensive as you. And most users want to attach sounds to their auras instead of having invisible auras. (Though invisible auras are sometimes useful for other rare use cases.)

So, what the PR contains:
- A delay timer on sound
- Looping/Repeat after x seconds
- Cancel Sound on "Hiding"

I think I have a pretty good idea on how to solve 2½ of that without adding a new region type. 

That's points to discuss: Why I think that not adding a region type is better, and how I think it should work.

Most users use sound in addition to their existing auras, and thus it would be somewhat odd to have sounds on the existing region as a second class citizen. And what I'm currently trying is to enable users to have less auras for their setups.

So the way I think that could work is to add a "sound loop" property to the conditions. This is very similar to how I transfered the "glow" action to a property for icons.

This property would consist of a "sound path" and a "repeat after x setting", and would loop the while the property is set. Thus e.g. you could have multiple conditions that loop different sounds, or with different repeat timers. (Since it is a property, only one would be active at any time.) The sound should obviously stop once the aura is hidden. The delay timer then maps easily to a new "active for x" seconds condition check that I want to add anyway.

Now, what that doesn't really solve is the question of how to cancel sounds for non looping sounds.

Also, adding a loop sound property now, could limit how users expect a "Play Sound" once "property" to work and thus predispose how that should be implemented.

And lastly I'm pretty busy atm and properly won't have time this weekend to prototype that.

InfusOnWoW:
sound

I asked our "alpha testers" for their opinion, for their benefit I'll attach a screenshot on how the pr looks like.

asakawa-k:
Maybe Looping sounds could be handled by adding a "While running" section to Actions?
So you can trigger a single sound effect on the _events_ OnShow or OnHide. Or you can loop/repeat sounds "While Running". I imagine people would also find uses for repeating chat messages while a trigger is true too (I know I've made chat spam auras for people before).

emptyrivers:
Personally, I'm torn on this one. On the one hand, I really like this idea, and think it should actually go farther than what you've already implemented, and just make an "action" regionType, instead of just sound. I know that personally, I've used the text-with-a-blank-space aura many times myself, and only on occasion did it have a sound attached. :-)

On the other hand, that's even more feature duplication.

krazyito65:
Looks like this is somewhat already in the works, but maybe this can help give some ideas?  Maybe not.  

Refer to this [post ](http://www.voximmortalis.com/threads/4328-WeakAuras-Tutoring-Thread?p=39388&viewfull=1#post39388) for the description of the workaround that was made.  Maybe it gives you some idea on what you can do.

https://wago.io/SoundLooper

Mpstark:
> Maybe Looping sounds could be handled by adding a "While running" section to Actions?

That's an interesting idea, one that solves a lot of issues with the current Action system. It introduces a lifecycle and allows for, uh, expectations of how the feature should work to be relatively minimal since it's a new set of options.

I could see this, in combination to tying the Actions feature into the Conditions feature (as Infus is alluding to), as fulfilling most of the current features of the pull request as a whole.

> Refer to this post for the description of the workaround that was made. Maybe it gives you some idea on what you can do.

I've read through the attached post, but it seems that my approach in code is far more simple (and honestly could be easily duplicated with `OnShow` and `OnHide` scripts). Basically, my approach is to use the `C_Timer` timer and ticker functionality. When the region is shown, a timer is used for delays, the sound is played (and the `SoundHandle` is stored), and a ticker is created which will repeat the sound if desired. When the region is hidden, `StopSound` is used with the `SoundHandle` to cut it off. The attached approach seems extremely heavy in comparison.

Additionally, I'm not sure why I've heard that "...the game API doesn't offer a means to stop the playback of a sound once it has begun..." a bunch. Since `StopSound` was introduced in 5.0.4 in 2012, it seems odd that a post more than 3 years later wouldn't know of it. I'm not sure if it was broken or perhaps only used by Blizzard or something else, but in any case, it works as of 7.0.

> and just make an "action" regionType

You know, I thought about this in particular before I started implementing but it just didn't seem worth it. Perhaps something like a "Custom" RegionType would fit better into this role. It would work much like a custom trigger and allow for the creation of custom displays. Honestly, I think that this fits in quite nicely with the rest of WA. Thoughts?

emptyrivers:
> Additionally, I'm not sure why I've heard that "...the game API doesn't offer a means to stop the playback of a sound once it has begun..." a bunch. Since StopSound was introduced in 5.0.4 in 2012, it seems odd that a post more than 3 years later wouldn't know of it. I'm not sure if it was broken or perhaps only used by Blizzard or something else, but in any case, it works as of 7.0.

That's probably got a lot to do with the rather poor documentation of the WoW API. ;) The thing is absolutely massive, and only in the last few months has any kind of support for actual documentation appeared from blizzard (which leaves over a decade of API calls left to be documented); additionally, changes made to the API are generally left unannounced in patch notes, requiring users to go hunting for exactly which changes were made (this exact thing happened to me in 7.2, when GetSpellTextureFileName was removed, presumably since it had been made redundant by texture escape strings [e.g. "|Tinv_misc_questionmark:0|t"] being able to support textureIDs returned by GetSpellTexture. But of course they didn't _tell_ anyone they made the change, breaking a couple of my things with no warning!).

> ...Honestly, I think that this fits in quite nicely with the rest of WA. Thoughts?

I do like this idea myself, although I'm not sure how feasible this would be in practice. If I understand your suggestion properly this would require some assumptions in how how the display is generated to be changed to accommodate a more customized version.

InfusOnWoW:
So I've made good progress on adding actions to the conditions feature, and should have experimental patch for that up sometime next week.

Mpstark:
@InfusOnWoW what do you think about:

>Maybe Looping sounds could be handled by adding a "While running" section to Actions?

InfusOnWoW:
So, to expand a bit on my thinking, because on reading it again it probably wasn't as clear.

I think we should explore first how to extend the current sound options, including the actions in conditions feature, before deciding on how to best proceed with this pull request.

The reason for that, is that most people probably use sounds as a way to enhance their auras, and not in the way you are apparently using them. From what I understand, you have plenty of auras whose only purpose is a sound. There's merit in having explicit support for that, also there are a few related ideas, which others have suggested. 

Now, for the conditions feature I added both "Loop" and "Sound" as options, so for consistency purposes I should also add Looping to the current actions.  Either by adding a "While shown" section, or by adding a "Start Looping" action to "On Show" and a "Stop Sound" action to "On Hide", which would only stop sounds started by that aura.  (That does solves a additional use case, but the interface is pretty bad.)

Also for Conditions, from what I can tell from how the interface felt like, the straight forward way to add actions doesn't feel great. I'll probably finish a first version of that patch on Thursday/Friday.

InfusOnWoW:
I've pushed a first version to here: https://github.com/InfusOnWoW/WeakAuras2/tree/smallpatches

It's not finished, and I would have liked to polish it a bit more before pushing it, but I don't have that much time tomorrow.

That branch also contains a few unrelated changes, and even a change that we won't do in that way. (The Mask Texture change.)

That patch adds a "Add Condition" button to the bottom of the actions tab. While I only added that for icons, it's trivial to add to other regions. That allows for "Sounds", "Sound Loops" and "Chats".

All Sounds of a region share a handle, and they stop a previous playing sound if needed.
I didn't add a "Stop Sound" option yet, nor a "While running" section to the Actions yet.

And I only tested it for a few minutes.

InfusOnWoW:
So, with some fuirther polish:
https://github.com/WeakAuras/WeakAuras2/pull/199

Noteably: The "Play Sound" action now has a option "Loop", and there's a "Stop Sound" action in "On Hide".

On Hiding, by default sounds will continue to play, but no longer loop.
If the "Stop Sound" toggle is checked, this will immediately interrupt the sound playing on hiding.

For Conditions there is a "Sound" option, which then allows "Play", "Loop" or "Stop Sound".
This behaves the same as the actions sounds. 

All sounds use the same sound handler and stop each other.

For example:
A sound loop can be started by "On Show".
A condition overwrite this sound loop.
A different condition (or On Hide) can stop this sound loop.

@Mpstark, if you could test that with your setups and just play around with it. I think it should solve a lot of problems for you. Note, that a "Active for x seconds" condition will come.

InfusOnWoW:
Oh well, no more response. I guess this is dead now?

Mpstark:
@InfusOnWoW 

The new actions added work for almost all use-cases, though looping with "Stop Sound" in actions doesn't actually function -- I'll open up an issue for that.

Again, sorry for the silence on this, I think I need to adjust the GitHub notifications or something.

PR #199, opened 2017-05-27
Add Actions to Conditions

InfusOnWoW:


InfusOnWoW:
While most cases with most conditions do work, there's some behaviour where the mismatch between properties and actiond do surface. I need to think about those cases for a bit. As far as the interface is, I think that works quite well.

InfusOnWoW:
That's now fixed. The bug in question was:

Imagine a "wrong" setup e.g.
If remaining time < 3 => bg-color: red
If remaining time < 6 => bg-color: green
If remaining time < 9 => bg-color: blue

With that setup, the last condition overwrites the previous conditions, and thus the bg-color is always blue. 

Now, for chat actions:
If remaining time < 3 => Send "3"
If remaining time < 6 => Send "6"
If remaining time < 9 => Send "9"

The same code would figure out that the "dominant" value is "9". So each time a condition would trigger, it would set the "chat" property to "9", meaning sending a "9" on 3s, 6s and 9s remaining.

The new code special cases "actions" and those won't be subjected to the same analysis, and thus this will send the correct chat messages on 9s, 6s and 3s remaining.

PR #205, opened 2017-06-06
 Implement %c support for "Actions/Chat" and "Conditions/Chat" 

InfusOnWoW:


InfusOnWoW:
I've just finished this and it's late. This probably needs a bit of testing first.

PR #212, opened 2017-06-18
Static Duration Option for AuraBars

harag-on-steam:
Replaces the duration from triggers with a static value. Progress above this value results in a full bar (or empty if inverse).

This is usefull when several bars have the same width and should move with the same speed but their triggers have different durations.


emptyrivers:
Isn't this just a variant of sticky duration? I would think that expanding sticky duration slightly would be more appropriate than adding another option.

harag-on-steam:
Sticky duration prevents triggers from updating the duration info with decreasing values. This option ignores duration info from triggers entirely. I think it would be unintuitive to merge these two things into a single UI option. 

Btw.: I'm not particularly fond of "Static Duration". "Override Duration"? "Fixed Duration"?

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1188,7 +1191,8 @@ local function modify(parent, region, data)
 
   function region:SetTime(duration, expirationTime, inverse)
     local remaining = expirationTime - GetTime();
-    local progress = duration ~= 0 and remaining / duration or 0;
+    local adjustedDuration = region.useStaticDuration and region.staticDuration or duration
```
harag-on-steam: I put this here because I think staticDuration is only really useful for bars. I also considered [WeakAuras.regionPrototype.AddSetDurationInfo()](https://github.com/WeakAuras/WeakAuras2/blob/6bc18c124767e6fb29e022ee0bb54e32f41b1284/WeakAuras/RegionTypes/RegionPrototype.lua#L106) but I'm too unfamilar with the code to be sure that the duration isn't updated somewhere between `region:SetDurationInfo()` and `region:SetTime()`

InfusOnWoW:
So, this is something that is regularly request, but I'd want a bit more general of a solution. (And in fact AddSetDurationInfo exists to make this easier.)

So, I would want it to also work with Progress Textures.
Also, I think it should also work with "static" durations. (That is not timer based.)

The way AddSetDurationInfo works, is that each region can provide 3 functions: SetTime, SetValue and TimerTick. If a region provides those functions, AddSetDurationInfo adds a SetDurationInfo to the region, which basically handles the translation from what SetDurationInfo gets to calling SetTime or SetValue.

harag-on-steam:
I had a second look at `SetDurationInfo()` and how it is used but I think it's not a good fit for this functionality because it's responsible for setting `region.duration`. This is used in `WeakAuras.TimerTick()` and `.UpateRegionValues()` to update the display texts so I think this should to remain the actual duration because I only want the graphical display to use a different scale.

Per your request I extended this to current/total values and to progress textures (and icon cooldown spinners)

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -525,9 +527,10 @@ local function modify(parent, region, data)
     end
 
     function region:SetTime(duration, expirationTime)
-      if (duration > 0) then
+      local adjustedDuration = region.progressScale or duration;
+      if (adjustedDuration > 0) then
         cooldown:Show();
-        cooldown:SetCooldown(expirationTime - duration, duration);
+        cooldown:SetCooldown(expirationTime - adjustedDuration, adjustedDuration);
```
InfusOnWoW: Did you test that? I'm not sure the cooldown can handle start times that are in the future. E.g. if adjustedDuration is less than duration.

harag-on-steam: I did. It does. Both normal and inverse. But I can remove it if it's risky to rely on that.

InfusOnWoW:
I'll have to play around with your code to figure out whether we you are right that this shouldn't be integrated into SetDuration.

The long term aim is to have as much code as possible between the regions shared, so in a way this moves us in the wrong direction.

WeakAuras/RegionTypes/Icon.lua
InfusOnWoW: Nah, if you tested that, it's fine. I'm a bit suprised.

InfusOnWoW:
So I finally have some time to look at this and had some areas, and as I said I prefer if there's as little duplicated code between regions as possible.

After massaging your commit for a bit I ended up with this:
https://github.com/WeakAuras/WeakAuras2/pull/222

Apart from centralizing the code:
- Scale is imho the wrong word, that's something where I would expect a multiplicative effect
- I added a "minimum" override. While probably less useful than maximum, I think it has some use cases
- I removed the icon code. In my experiments the cooldown spiral that was shown started moving immediately, even if I overrode the max duration with a smaller value.
- I used sliders instead of a text input

harag-on-steam:
> started moving immediately, even if I overrode the max duration with a smaller value

That's odd, it definitely didn't do that for me. Either way, it's unreliable and I only included it for the sake of completeness. 

I'll checkout your PR and give it a try.

harag-on-steam:
Abandoned in favor of #222

PR #217, opened 2017-06-23
Fix potential Stack overflow

emptyrivers:
As it happens, there actually is a use for calling WeakAuras.ScanEvents inside of the init script of an aura: doing so triggers ActivateAuraEnvironment, which then runs the appropriate init script if it hasn't completed yet; handy if you want to ensure that one aura (a library of sorts) which defines several functions you use elsewhere is loaded first. However, the code has no notion of whether or not an init script has begun execution or not, which leads to the possibility of a ScanEvents chain leading back to your original init script. This leads to infinite recursion, and very quickly a stack overflow.

This PR addresses this by making data.init_completed trivalued; it can be 'not started', 'started', and 'complete', and ActivateAuraEnvironment will only run the init script when that script hasn't been started or completed yet.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -372,7 +372,7 @@ function WeakAuras.ActivateAuraEnvironment(id, cloneId, state)
     current_aura_env = aura_env_stack[#aura_env_stack] or nil;
   else
     local data = db.displays[id];
-    if data.init_not_started then
+    if data.init_started then
       -- Point the current environment to the correct table
       aura_environments[id] = aura_environments[id] or {};
       current_aura_env = aura_environments[id];
@@ -394,7 +394,7 @@ function WeakAuras.ActivateAuraEnvironment(id, cloneId, state)
         local func = WeakAuras.customActionsFunctions[id]["init"];
         if func then
           current_aura_env.id = id;
-          data.init_completed = 1;
+          data.init_started = 1;
           func();
         end
       end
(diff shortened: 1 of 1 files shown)


emptyrivers:
Also I changed the name of init_completed to init_progress, to make its purpose clear

InfusOnWoW:
From a very quick glance it doesn't look like we need a tristate. Setting the init_progress/init.complete to true before running the init function instead of after should be enough. Or am I missing something?

Also, not towards you, but whoever does  call ScanEvents from a init function.... We don't guarantee when that function is run, except that it is run before the trigger is run.

emptyrivers:
Setting the flag before running the init script would suffice, I suppose. I didn't think of that :)

As to your second point; I don't think that's especially a bad thing, so long as the author of the aura is aware of it. I would liken that behavior to a dependency, instead of a load order, I suppose (particularly if you use a dummy trigger function which immediately returns, and so its only purpose is to provide something that ScanEvents can latch onto). 

Perhaps a more formalized method of establishing a dependency/load order would be more appropriate in the long run?

emptyrivers:
Don't mind the 2 commits that were reverted, that would be me pushing before I intended (its late). The last commit gives a binary version of the original tristate.

InfusOnWoW:
You misunderstood me. Not two binaries. Two states. I don't think we need to distinguish between "started" and "completed". All we care is between "not run" and "started/completed".

(I think)

emptyrivers:
As best as I can tell this is ready for merging now. Last commit I managed to not find&replace correctly,  apologies.

WeakAuras/WeakAuras.lua
```diff
@@ -394,10 +394,10 @@ function WeakAuras.ActivateAuraEnvironment(id, cloneId, state)
         local func = WeakAuras.customActionsFunctions[id]["init"];
         if func then
           current_aura_env.id = id;
+          data.init_started = 1;
```
InfusOnWoW: you need to set it regardless of whether there's a action or not, thus do it before the if

emptyrivers: Yep. Fixed in [7c9f7c9](https://github.com/WeakAuras/WeakAuras2/pull/217/commits/7c9f7c970235a8d782d46fbea94fcb3ee33f2dab).

InfusOnWoW:
Looks fine to me, can you squash it?

PR #256, opened 2017-07-23
Smallpatches

InfusOnWoW:
Maybe, we should call SetDurationInfo with nil, nil instead.

But that would require further changes down the line

Ticket-Nr: 1010

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -852,7 +852,7 @@ WeakAuras.load_prototype = {
     },
     {
       name = "difficulty",
-      display = L["Dungeon Difficulty"],
+      display = L["Instance Difficulty"],
       type = "multiselect",
       values = "difficulty_types",
       init = "arg"
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -536,6 +536,21 @@ function WeakAuras.UseUnitPowerThirdArg(powerType)
   return nil;
 end
 
+function WeakAuras.GetNumSetItemsEquipped(setID)
+  if not setID or not type(setID) == "number" then return end
+  local itemList = C_LootJournal.GetItemSetItems(setID)
+  if not itemList then return end
+  local setName = GetItemSetInfo(setID)
+  local max = #itemList
+  local equipped = 0
+  for _,v in ipairs(itemList) do
+    if IsEquippedItem(v.itemID) then
+      equipped = equipped + 1
+    end
+  end
+  return equipped, max, setName
+end
+
 local function valuesForTalentFunction(trigger)
   return function()
     local single_class;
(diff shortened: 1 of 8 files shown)


InfusOnWoW:
The Item Set change could use a bit more work.

WeakAuras/Prototypes.lua
```diff
@@ -3404,7 +3419,45 @@ WeakAuras.event_prototypes = {
...
+      }
+    },
+    durationFunc = function(trigger)
+      return WeakAuras.GetNumSetItemsEquipped(trigger.itemSetId and tonumber(trigger.itemSetId) or 0)
+    end,
+    nameFunc = function(trigger)
+      return select(3, WeakAuras.GetNumSetItemsEquipped(trigger.itemSetId and tonumber(trigger.itemSetId) or 0));
+    end
+  },
+  ["Item Set Equipped"] = { -- Equipment Set
```
Stanzilla: Should probably rename that to Equipment Set Equipped to not be confusing.

Stanzilla: AceGUIWidget-WeakAurasDisplayButton.lua needs the minor bumped at the top

WeakAuras/Prototypes.lua
```diff
@@ -3404,7 +3419,63 @@ WeakAuras.event_prototypes = {
...
+
+          local sets = C_LootJournal.GetFilteredItemSets();
+          C_LootJournal.SetClassAndSpecFilters(classFilter, specFilter);
+
+          local description = "";
+          for index, set in ipairs(sets) do
+            description = description .. set.name .. ": " .. set.setID .. "\n";
+          end
+
+          description = description .. "\n" .. L["Older sets ids can be find on websites such as wowhead.com"];
```
Stanzilla: "Older set IDs can be found on websites such as wowhead.com/item-sets"

WeakAuras/Prototypes.lua
```diff
@@ -3404,7 +3419,63 @@ WeakAuras.event_prototypes = {
...
+    name = L["Item Set Equipped"],
+    automaticrequired = true,
+    init = function(trigger)
+      return string.format("local setid = %s;\n", trigger.itemSetId and tonumber(trigger.itemSetId) or "0");
+    end,
+    statesParameter = "one",
+    args = {
+      {
+        name = "itemSetId",
+        display = L["Item Set"],
```
Stanzilla: should probably add "id" here as well so it's clear what the input field wants

WeakAuras/Transmission.lua
```diff
@@ -826,6 +827,11 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
       tinsert(tooltip, {1, data.url, 1, 0.82, 0, 1});
     end
 
+    if (hasVersion) then
+      tinsert(tooltip, {1, L["Version: "] .. data.version, 1, 0.82, 0, 1});
+    end
+
+
```
Stanzilla: YOU AND YOUR BLANK LINES

PR #258, opened 2017-07-25
fix PRD hook when KNP option 'ignore personal nameplate' is used

kesava-wow:
If the user enables "ignore personal nameplate" in KNP, a WeakAura which is supposed to be anchored to the PRD simply isn't shown (since the .kui frame isn't shown, but still exists on the nameplate).

Also, you don't use KuiNameplatesPlayerAnchor. Am I doing something wrong with that? (not that using that would fix it, just wondering.)

<3


InfusOnWoW:
I think when I wrote it KuiNameplatesPlayerAnchor wasn't the same size as the visible nameplate.

kesava-wow:
Ah, well, it still isn't. So.. fair enough.

InfusOnWoW:
I probably should have told you that. :)

PR #279, opened 2017-09-02
Change one PlaySound to use kit ids

InfusOnWoW:
Report and patch by Exality on discord.


InfusOnWoW:
That's the enable/disable import checkbox. Not exactly critical enough to do a immediate release with the fix.

PR #283, opened 2017-09-10
protect _G lookups

lqnrd:
Dunno how else to post this. There are currently two different _G lookups that can be fed arbitrary keys from within custom display code to retrieve globals otherwise blocked through blockedFunctions. I can provide proof of concept code if needed (via Twitter pm @lqnrd), but will obviously not post that here.

Some comments about this pull:

- `_G` and `getglobal` being added to `blockedFunctions` is just a convenient way of having those two included in the check, and will not alter the rest of the code, as these two keys are checked and handled in `exec_env` before `blockedFunctions` is checked.

- The two functions `WeakAuras.PerformActions` and `GetAnchorFrame` will throw errors when called with malicious arguments, but should still function just as before with legitimate input. For now the user is warned about a lookup being blocked just like when calling a blocked function directly. Maybe there is a better way to cleanly exit them, but then again, calling a blocked function directly can also result in errors outside of the control of this addon.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -2576,7 +2576,7 @@ function WeakAuras.PerformActions(data, type, region)
         glow_frame = regions[frame_name].region;
       end
     else
-      glow_frame = _G[actions.glow_frame];
+      glow_frame = exec_env[actions.glow_frame];
       original_glow_frame = glow_frame
     end
 
@@ -4350,8 +4350,8 @@ local function GetAnchorFrame(id, anchorFrameType, parent, anchorFrameFrame)
       end
       postponeAnchor(id);
     else
-      if (_G[anchorFrameFrame]) then
-        return _G[anchorFrameFrame];
+      if (exec_env[anchorFrameFrame]) then
+        return exec_env[anchorFrameFrame];
       end
       postponeAnchor(id);
       return  parent;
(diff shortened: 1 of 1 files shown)


Stanzilla:
I don't think we can get away with blocking _G, too many auras currently use that for various reasons.

lqnrd:
That's why the order of checks inside the [metatable](https://github.com/lqnrd/WeakAuras2/blob/ce56767432bab5e4e726cb6e1e388b3517e290c4/WeakAuras/WeakAuras.lua#L420) is important.

Right now `_G["RunScript"]` is blocked but e.g. `_G["Minimap"]` is not. This patch does not do anything to change this because in the metatable index function `_G` and `getglobal` are checked _before_ checking for forbidden functions. That's why these two indexes will never be checked there.

The two other places changed in this patch expect a frame table reference. `actions.glow_frame` and `anchorFrameFrame` are expected to be strings and the global names of frames. Blocking the non-frame `_G` and the function `getglobal` (along with everything else in `blockedFunctions`) _just in these two places_ should not change the behaviour of any displays.

lqnrd:
-.- reopen because I hit the wrong button

emptyrivers:
A similar effect to this PR could be achieved by simply keying `exec_env` instead of `_G`.

InfusOnWoW:
I think using exec_env is a better fix. Can you rewrite your pull request to use exec_env instead of _G in those two functions?

lqnrd:
I was just waiting for confirmation on this

PR #284, opened 2017-09-10
Security Fix: Scrub inputs of ScanEvents

emptyrivers:
WeakAuras.ScanEvents can be manipulated to send any value, including `_G` into the sandbox. To prevent this, we can [scrub](http://wowprogramming.com/docs/api/scrub) the inputs to ensure that the only things passed in are innocuous.

This is a breaking change for any custom aura that relies on ScanEvents to pass complicated information between it and other auras.

Note: I discovered and reported this issue (via discord) on August 4; I simply couldn't think of a good fix until today.

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -479,11 +479,11 @@ function WeakAuras.ScanEvents(event, arg1, arg2, ...)
           local untriggerCheck = false;
           local allStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
           if (data.statesParameter == "full") then
-            if (data.triggerFunc(allStates, event, scrub(argi, arg2, ...))) then
+            if (data.triggerFunc(allStates, event, scrub(arg1, arg2, ...))) then
               updateTriggerState = true;
             end
           elseif (data.statesParameter == "all") then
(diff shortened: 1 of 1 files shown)


nebularg:
Stanzilla told me to look at this and the only thing that I noticed was that you should probably include arg1, arg2 in the scrub call. I didn't trace through to check, but if those are also user values, they should be scrubbed.

WeakAuras/GenericTrigger.lua
```diff
@@ -479,11 +479,11 @@ function WeakAuras.ScanEvents(event, arg1, arg2, ...)
           local untriggerCheck = false;
           local allStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
           if (data.statesParameter == "full") then
-            if (data.triggerFunc(allStates, event, arg1, arg2, ...)) then
+            if (data.triggerFunc(allStates, event, scrub(argi, arg2, ...))) then
```
nebularg: argi? >.>

emptyrivers: oops.

emptyrivers:
That's fair. The first arg to ScanEvents (the `event` itself) will only propogate into the sandbox, should it have already been loaded (which, from within WeakAuras can only be a safe value, as best as I can tell). `arg1` and `arg2` are certainly user values and shouldn't be trusted.

PR #301, opened 2017-11-11
Format percenthealth to limit to two decimal places

oscheller1:
Without this fix, %percenthealth would display a high number of decimal places


InfusOnWoW:
Yes, but I want the state.percenthealth to contain the full digits. A proper fix will eventually come.

InfusOnWoW:
Here's a bit of context: https://www.wowace.com/projects/weakauras-2/issues/958

I do have a idea how to improve that, and allow formatting.

PR #308, opened 2017-11-25
Smallpatches

InfusOnWoW:


WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -70,6 +70,65 @@ function WeakAuras.MultipleDisplayTooltipDesc()
   return desc;
 end
 
+function WeakAuras.DuplicateOneAura(data)
```
Stanzilla: nit: I think DuplicateAura() is a better name

PR #310, opened 2017-11-29
field 'spellCache' can be nil

lqnrd:
`WeakAuras.spellCache` is nil until WeakAurasOptons has been loaded.

Steps to reproduce: Reload interface. Have another person link a display with an Aura trigger (and Aura Name, like spellId). Click on link.

I'm not sure if the same can happen in line 468 in function `WeakAuras.DisplayToString`.


Stanzilla:
Thanks! It should not happen in 468 since that checks for the savedvar, I think

PR #374, opened 2018-04-03
Fix Copy/Paste code for Groups

InfusOnWoW:
- Don't copy controlledChildren.
- In the context menu for groups offer only "Copy Group"
- Pasting a individual auras settings onto a group paste it to
  each child.


Stanzilla:
bump the minor of the widget, please, otherwise looks good, will test later

PR #376, opened 2018-04-04
Smallpatches

InfusOnWoW:


InfusOnWoW:
The Profiling Support patch isn't "finished" yet. I know of a few issues.

PR #385, opened 2018-04-19
Add bindings

asakawa-k:
Small additions just to add an easy binding the the main game's KeyBindings settings. I would find this useful, personally, and it's unobtrusive enough that it shouldn't bother anyone that wouldn't find it useful.


Stanzilla:
Thanks!

PR #387, opened 2018-04-22
fix form_types

mrbuds:
GetShapeshiftFormInfo() doesn't return name


Stanzilla:
he copied the existing style which is ok imo. string.format vs concat is something we could talk about

PR #388, opened 2018-04-22
Fix UnitAura (unit, spellName, ...)

stsundermann:
UnitAura does not take a spell name anymore so we have to loop over
all auras.

### DIFF
--- a/WeakAuras/BuffTrigger.lua
+++ b/WeakAuras/BuffTrigger.lua
@@ -588,8 +588,18 @@ function WeakAuras.ScanAuras(unit)
 
             for index, checkname in pairs(data.names) do
               -- Fetch aura data
-              name, icon, count, _, duration, expirationTime, unitCaster, isStealable, _, spellId = UnitAura(unit, checkname, nil, filter);
-              if (name) then
+              -- TODO 8.0: Check if there is a better way than iterating all auras
+              local detected
+              for i = 1, _G.BUFF_MAX_DISPLAY do
+                name, icon, count, _, duration, expirationTime, unitCaster, isStealable, _, spellId = UnitAura(unit, i, filter);
+                if not name then break end
+                if name == checkname then
+                  detected = true
+                  break
+                end
+              end
+
+              if (detected) then
                 WeakAuras.SetDynamicIconCache(name, spellId, icon);
               end
               checkPassed = false;
@@ -960,8 +970,19 @@ do
       local updateTriggerState = false;
       for triggernum, data in pairs(triggers) do
         local filter = data.debuffType..(data.ownOnly and "|PLAYER" or "");
-        local name, icon, count, debuffType, duration, expirationTime, unitCaster, isStealable, nameplateShowPersonal, spellId = UnitAura(unit, spellName, nil, filter);
-        if(name and (data.spellId == nil or data.spellId == spellId)) then
+
+        -- TODO 8.0: Check if there is a better way than iterating all auras
+        local detected
+        for i = 1, _G.BUFF_MAX_DISPLAY do
+          local name, icon, count, debuffType, duration, expirationTime, unitCaster, isStealable, nameplateShowPersonal, spellId = UnitAura(unit, i, filter);
+          if not name then break end
+          if name == spellName then
+            detected = true
+            break
+          end
+        end
+
+        if(detected and (data.spellId == nil or data.spellId == spellId)) then
           data.GUIDs = data.GUIDs or {};
           data.GUIDs[destGUID] = data.GUIDs[destGUID] or {};
           data.GUIDs[destGUID].name = spellName;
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
Patch looks fine.

Two comments: Please join the wa-alpha-testing channel on discord, if you plan on fixing stuff. That's our secret hide out.

Secondly, we put TODO 8.0 comments onto lines that we might need to touch again. I think this qualifies.

stsundermann:
Sure, I have a working WeakAuras for 8.0 but I did not split up my fixes into seperate commits. That's what I'm doing now :smiley:

stsundermann:
I've added the 8.0 comments

InfusOnWoW:
Okay, looking forward to your stuff then.

WeakAuras/BuffTrigger.lua
```diff
@@ -588,8 +588,18 @@ function WeakAuras.ScanAuras(unit)
 
             for index, checkname in pairs(data.names) do
               -- Fetch aura data
-              name, icon, count, _, duration, expirationTime, unitCaster, isStealable, _, spellId = UnitAura(unit, checkname, nil, filter);
-              if (name) then
+              -- TODO 8.0: Check if there is a better way than iterating all auras
+              local detected
+              for i=1, BUFF_MAX_DISPLAY do
```
Stanzilla: `i = 1` and `_G.BUFF_MAX_DISPLAY`

WeakAuras/BuffTrigger.lua
```diff
@@ -960,8 +970,19 @@ do
       local updateTriggerState = false;
       for triggernum, data in pairs(triggers) do
         local filter = data.debuffType..(data.ownOnly and "|PLAYER" or "");
-        local name, icon, count, debuffType, duration, expirationTime, unitCaster, isStealable, nameplateShowPersonal, spellId = UnitAura(unit, spellName, nil, filter);
-        if(name and (data.spellId == nil or data.spellId == spellId)) then
+
+        -- TODO 8.0: Check if there is a better way than iterating all auras
+        local detected
+        for i=1, BUFF_MAX_DISPLAY do
```
Stanzilla: same as above

WeakAuras/BuffTrigger.lua
```diff
@@ -1119,8 +1140,19 @@ do
             local updateTriggerState = false;
             for triggernum, data in pairs(triggers) do
               local filter = data.debuffType..(data.ownOnly and "|PLAYER" or "");
-              local name, icon, count, debuffType, duration, expirationTime, unitCaster, isStealable, nameplateShowPersonal, spellId = UnitAura(uid, spellName, nil, filter);
-              if(name) then
+
+              -- TODO 8.0: Check if there is a better way than iterating all auras
+              local detected
+              for i=1, BUFF_MAX_DISPLAY do
```
Stanzilla: same as above

Stanzilla:
Hey there, thanks for your contribution. We are currently unsure if we should only loop for `BUFF_MAX_DISPLAY` which is actually `32` or `40` which we thought is the max of `UnitAura` (https://wow.gamepedia.com/API_UnitAura) but want to go with it for now since we can stil revisit later if this should be wrong. Blizzard is using it like you did in their own code (https://github.com/tomrus88/BlizzardInterfaceCode/search?utf8=%E2%9C%93&q=BUFF_MAX_DISPLAY&type=) so it's probably fine.

stsundermann:
Updated the PR

stsundermann:
I'm on the weakauras discord but I cannot find the #wa-alpha-testing channel.

InfusOnWoW:
Use this: !role alpha while on the discord

krazyito65:
What's your username? I can set you up so you don't have to wait the 15 inital minutes

InfusOnWoW:
Ah there's a waiting period? Seems our secret hide-out has top-notch security :)

krazyito65:
Haha, nah it's for speaking in the discord at all when you first join to prevent easy spam bots

WeakAuras/BuffTrigger.lua
```diff
@@ -588,8 +588,18 @@ function WeakAuras.ScanAuras(unit)
 
             for index, checkname in pairs(data.names) do
               -- Fetch aura data
-              name, icon, count, _, duration, expirationTime, unitCaster, isStealable, _, spellId = UnitAura(unit, checkname, nil, filter);
-              if (name) then
+              -- TODO 8.0: Check if there is a better way than iterating all auras
+              local detected
+              for i = 1, _G.BUFF_MAX_DISPLAY do
```
emptyrivers: There's no need to index _G;  a simple global access is sufficient. Additionally, since this particular value may get used many many times (given how often `UNIT_AURA` fires), it is probably worth localizing `BUFF_MAX_DISPLAY` into an upvalue for performance concerns

emptyrivers: Ah, i see @Stanzilla requested the _G index in an earlier review (serves me right for just scrolling to the bottom). My point still stands about performance concerns, however; no reason not to localize the value and only pay the cost of global access just the once.

InfusOnWoW:
Did you join? I'm interested in hearing what else you have.

stsundermann:
I did but I am not home right now

PR #393, opened 2018-04-27
Change progressPrecision to use floor

asakawa-k:
Having rounded numbers for a timer instead of floored feels odd, especially when showing no decimals, zero shows for half a second. Leaving the totalPrecision using Round seems fine though so I haven't altered that stuff.

### DIFF
--- a/WeakAuras/RegionTypes/RegionPrototype.lua
+++ b/WeakAuras/RegionTypes/RegionPrototype.lua
@@ -265,7 +265,7 @@ local function UpateRegionValues(region)
     elseif (region.progressPrecision == 4 or region.progressPrecision == 5) and remaining > 3 then
       remainingStr = remainingStr..floor(remaining);
     else
-      remainingStr = remainingStr..floor(remaining*(10^region.progressPrecision))/(10^region.progressPrecision);
+      remainingStr = remainingStr..floor(remaining*(10^(region.progressPrecision or 0)))/(10^(region.progressPrecision or 0));
     end
   else
     remainingStr     = " ";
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
So. I just looked at this again. And, there's a additional wrinkle, we aren't the only one formatting those numbers. There's the very popular omnicc addon and recently Blizzard added thier own formatting.

OmniCC's formatting matches our current formatting.
Whereas Blizzard uses ceil not floor.

Thus this PR would make us different from both.

asakawa-k:
Hmm, Blizz use `ceil` on `SecondsToTimeAbbrev` but `SecondsToTime`, which is used more often, defaults to `floor` unless told to roundUp. That is done in `StatusBar` timers and `BaseTextTimerMixin`, tooltip CDs. But left as floor in a bunch of other areas. 

I hadn't realised this was such a mess. 
Or that OmniCC used Round too. 
But maybe we can at least agree that Round is the least accurate/most misleading option?
In terms of trying to judge when a text timer will end, if you're following the rhythm of the seconds ticking away, the 0 shows for 0.5 seconds then the timer ends.

InfusOnWoW:
So Blizzard isn't even consistent in their own handling. Interesting.

Stanzilla:
What do we want to do here?

InfusOnWoW:
I think we need a solution, but a different one. It's something I want to tackle later this year.

emptyrivers:
Arguably, this is in the scope of #1618, so i'm closing it. If somebody disagrees we can keep it open so we don't forget about this pain point.

PR #394, opened 2018-04-27
Add state.percent

asakawa-k:
Adds a rounded state.percent to static duration triggers

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -385,6 +385,7 @@ function WeakAuras.ActivateEvent(id, triggernum, data, state)
     end
     state.value = nil;
     state.total = nil;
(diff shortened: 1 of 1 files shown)


Stanzilla:
@InfusOnWoW thoughts?

asakawa-k:
It was a bit of a daft idea, just to try and give people something simpler than :
```LUA
function()
    if aura_env.state and aura_env.state.percenthealth then
        return Round(aura_env.state.percenthealth).."%"
    end
end
```
For showing what they think of as something very simple and obvious. If it doesn't bloat things then maybe you think it's fine but it's "inelegant" to say the least.

InfusOnWoW:
I want to solve that a more generally. I just renamed the PR, because @asakawa-k didn't give it a good name.

PR #408, opened 2018-05-09
[Feature] Add HSV color animations

emptyrivers:
A straight RGB color animation from, say, green to red, has the unfortunate property of being really quite ugly often. This PR uses the ColorSelect widget to get HSV values for colors, which often produces a nicer looking transition.

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -148,36 +148,36 @@ local hsvFrame = CreateFrame("Colorselect")
 -- see http://www.wowinterface.com/forums/showthread.php?t=48236
 function WeakAuras.GetHSVTransition(perc, r1, g1, b1, r2, g2, b2)
   --get hsv color for colorA
-  hsvFrame:SetColorRGB(r1,g1,b1)
+  hsvFrame:SetColorRGB(r1, g1, b1)
   local h1, s1, v1 = hsvFrame:GetColorHSV() --radius, saturation, luminance
   --get hsv color for colorB
-  hsvFrame:SetColorRGB(r2,g2,b2)
+  hsvFrame:SetColorRGB(r2, g2, b2)
   local h2, s2, v2 = hsvFrame:GetColorHSV() --radius, saturation, luminance
   --calculate new HSV values based on HSV values from colorA and colorB and multiply them by the perc
-  local h3 = floor(h1-(h1-h2)*perc)
+  local h3 = floor(h1 - (h1 - h2) * perc)
   --check if the angle between the two H values is > 180
-  if abs(h1-h2) > 180 then
-      local radius = (360-abs(h1-h2))*perc/100
+  if abs(h1 - h2) > 180 then
+      local radius = (360 - abs( h1 - h2 )) * perc / 100
       --calculate the 360° breakpoint
       if h1 < h2 then
-          h3 = floor(h1-radius)
+          h3 = floor(h1 - radius)
           if h3 < 0 then
-              h3 = 360+h3
+              h3 = 360 + h3
           end
       else
-          h3 = floor(h1+radius)
+          h3 = floor(h1 + radius)
           if h3 > 360 then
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -142,6 +142,45 @@ WeakAuras.function_strings = {
   ]]
 };
 
+local hsvFrame = CreateFrame("Colorselect")
```
InfusOnWoW: The RGB to HSV conversion isn't that complicated, I wonder wheter writing that method instead of using the Colorselect to transform would be easier.

WeakAuras/Prototypes.lua
```diff
@@ -142,6 +142,45 @@ WeakAuras.function_strings = {
...
+          end
+      end
+  end  
+  local s3 = s1-(s1-s2)*perc
+  local v3 = v1-(v1-v2)*perc
+  --get the RGB values of the new color
+  hsvFrame:SetColorHSV(h3, s3, v3)
+  local r,g,b = hsvFrame:GetColorRGB()
+  --return the new color
+  return r,g,b
```
InfusOnWoW: The alpha values should be interpolated too

WeakAuras/Prototypes.lua
```diff
@@ -142,6 +142,45 @@ WeakAuras.function_strings = {
...
+  --get hsv color for colorA
+  hsvFrame:SetColorRGB(r1,g1,b1)
+  local h1, s1, v1 = hsvFrame:GetColorHSV() --radius, saturation, luminance
+  --get hsv color for colorB
+  hsvFrame:SetColorRGB(r2,g2,b2)
+  local h2, s2, v2 = hsvFrame:GetColorHSV() --radius, saturation, luminance
+  --calculate new HSV values based on HSV values from colorA and colorB and multiply them by the perc
+  local h3 = floor(h1-(h1-h2)*perc)
+  --check if the angle between the two H values is > 180
+  if abs(h1-h2) > 180 then
```
InfusOnWoW: Can you explain what is going on here?

WeakAuras/Prototypes.lua
emptyrivers: Considering that this is lifted almost directly from the url in the comment (with just a few modifications to adapt it to the WeakAuras environment), it certainly wouldn't be *easier* to write a home brew solution to do the same thing, though it'd probably be more transparent.

WeakAuras/Prototypes.lua
emptyrivers: Fair enough.

WeakAuras/Prototypes.lua
emptyrivers: This is intended to pick the shortest arc of the hue circle, since there's always two options for how to transition from one color to another. If you're going from red to green, for example, you can either pass through yellow or blue.

Stanzilla:
1) We should probably target this at `8.0` since we didn't want to do no features in `master` anyway.
2) Is there any need to keep the RGB functions alive? Are there any usecases where RGB would be preferred over HSV? I really don't like sticking acronyms to labels where only 1% of the users know what it is and replacing the default would be better IF there are no cases where RGB is still preferred.

emptyrivers:
@Stanzilla 
> We should probably target this at 8.0 since we didn't want to do no features in master anyway.

I can close this PR and open a new one for the 8.0 branch, but I don't have beta access so I can't do any testing (this is why i opened on master in the first place). I suspect that it would be fine, anyways since I don't think that there's any important changes.

> Is there any need to keep the RGB functions alive? ... 

I personally don't think so. If others agree then I can nuke the RGB interpolations. Particularly if this is going to be packaged with 8.0, where there are so many breaking changes - if this was merged to master then I expect that at least one person would open a bug report that the colors changed.

christinaa:
LGTM, though I had to manually merge the changes into the current 8.0 release 5, River's base branch is weird.

PR #441, opened 2018-06-09
Smallpatches

InfusOnWoW:


WeakAurasOptions/RegionOptions/ProgressTexture.lua
```diff
@@ -407,6 +407,23 @@ local function createThumbnail(parent)
   local foreground = region:CreateTexture(nil, "ART");
   borderframe.foreground = foreground;
 
+  local  OrgSetTexture = foreground.SetTexture;
```
Stanzilla: there's an extra space here

PR #450, opened 2018-06-17
Basic aura update support

emptyrivers:


emptyrivers:
This is still pretty rudimentary (in particular, there is no granularity whatsoever on _what_ is updated if the user clicks `update`). What is supported is the following:

- best-effort id matching. This is intended to assist with the transition, since at the moment I am writing this comment, there are exactly 7 auras with a unique id.
- uid generation/matching. Any aura which is exported to a string will be given a globally unique identifier (essentially a random number in base 64) if it does not already have one, which can be used to unambiguously match in the future.
- best-effort group matching. Upon detecting a match for a group, WeakAuras will attempt to detect how that group has changed (notably, which members have been added, modified, and deleted in the import). If no change is detected, then the user will be notified as such and given the option of essentially just creating a duplicate.

WeakAuras/Types.lua
```diff
@@ -198,7 +198,8 @@ WeakAuras.race_types = {
   HighmountainTauren = LBR("Highmountain Tauren"),
   Nightborne = LBR("Nightborne"),
   DarkIronDwarf = LBR("Dark Iron Dwarf"),
-  ZandalariTroll = LBR("Zandalari Troll")
+  ZandalariTroll = LBR("Zandalari Troll"),
+  MagharOrc = LBR("Mag'har Orc"),
```
emptyrivers: Mmm, this isn't really supposed to be in the PR really. But I suppose it would've needed to be added eventually anyways. Note that as of today, it's a no-op since LBR doesn't have maghar orc data.

WeakAuras/Transmission.lua
```diff
@@ -458,12 +469,12 @@ function StringToTable(inString, fromChat)
   return deserialized;
 end
 
-function WeakAuras.DisplayToString(id, forChat)
+function WeakAuras.DisplayToString(id, forChat, unique)
```
InfusOnWoW: Unique seems to be true for all callers. If so remove it.

WeakAuras/Transmission.lua
```diff
@@ -1109,13 +1234,6 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
 
     local RegularGetData;
     if(children) then
-      data.controlledChildren = {};
-      for index, childData in pairs(children) do
-        if(compressed) then
-          DecompressDisplay(childData);
-        end
-        data.controlledChildren[index] = childData.id;
```
InfusOnWoW: This no longer happens.

WeakAuras/Transmission.lua
```diff
@@ -762,15 +809,6 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
       else
         tinsert(tooltip, {1, L["No Children:"], 1, 1, 1});
       end
-    elseif(data.controlledChildren) then
```
InfusOnWoW: Can you explain why you removed this code? It seems used to me.

WeakAuras/Transmission.lua
```diff
@@ -934,6 +987,9 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
             data.parent = nil;
             findUnusedId(data);
             importData(data);
+            if matchedData then
+              data.uid = nil
```
InfusOnWoW: Can you move those two lines above the importData. It makes it cleared that the uid is cleared in the imported data.

WeakAuras/Transmission.lua
```diff
@@ -942,27 +998,25 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
                 childData.parent = data.id;
                 tinsert(data.controlledChildren, childData.id);
                 importData(childData);
+                if matchedChildren then
+                  childData.uid = nil
+                end
```
InfusOnWoW: Same here.

WeakAuras/Transmission.lua
```diff
@@ -972,13 +1026,81 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
...
+                optionsFrame = WeakAuras.OptionsFrame();
+              end
+
+              if not(WeakAuras.IsOptionsOpen()) then
+                WeakAuras.ToggleOptions();
+              end
+
+              local function importUpdatedData(data, oldData)
+                if oldData then
+                  WeakAuras.Update(data, oldData, "important");
```
InfusOnWoW: I'm a bit suprised that the modified check and the importing aren't symetric. That is a aura can be modifed, yet the importing doesn't actually update anything. I would have expected WeakAuras.Update and WeakAuras.diff to use the same criteria.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+        if not oldChildren[i] then
+          modified = true
+          info.added = info.added + 1
+        else
+          if WeakAuras.diff(oldChildren[i], child) then
+            modified = true
+            info.modified = info.modified + 1
+            oldTotal = oldTotal - 1
+          else
+            child.same = true
```
InfusOnWoW: I don't like that this meta-data about the data is stored inside the data, I would have expected that in info.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+        replaceFields(tf1, tf2, val)
+      end
+    elseif val then
+      t1[field] = t2[field] or t1[field]
+    end
+  end
+end
+
+function WeakAuras.Update(oldData, newData, mode)
+  WeakAuras.DeleteOption(oldData)
```
InfusOnWoW: I'm not sure if that is the right thing to do if the id is unchanged.

WeakAuras/Transmission.lua
```diff
@@ -458,12 +469,12 @@ function StringToTable(inString, fromChat)
...
 end
 
-function WeakAuras.DisplayToString(id, forChat)
+function WeakAuras.DisplayToString(id, forChat, unique)
   local data = WeakAuras.GetData(id);
   if(data) then
     local transmitData = CompressDisplay(data);
-    transmitData.controlledChildren = nil;
     local children = data.controlledChildren;
+    transmitData.uid = unique and (transmitData.uid or GenerateUniqueID()) or nil
```
InfusOnWoW: You actually need to set the uid in data, so that on next export the uid is the same.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+  if data.uid then uids[data.uid] = 0 end
+  local oldChildren
+  if children then
+    oldChildren = {}
+    for i, child in ipairs(children) do
+      if child.uid then uids[child.uid] = i end
+      oldChildren[i] = db.displays[child.id]
+    end
+  end
+  for id, existingData in pairs(db.displays) do
```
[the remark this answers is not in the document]
InfusOnWoW: Those can't happen.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+        return true
+      end
+    end
+  end
+end
+
+local ignoredForDiffChecking = {
+  init_started = true,
+  yOffset = true,
+  xOffset = true,
```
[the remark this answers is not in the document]
emptyrivers: I agree, hence the unused arguments and seemingly over-engineered functions that you’ve commented on. But this is intended to be a first functional pass which works for the primary use case, where the user only imports, and then maybe repositions the aura but doesn’t make any other changes whatsoever.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+  width = true,
+  frameStrata = true,
+  anchorFrameType = true,
+  selfPoint = true,
+  height = true,
+  fixedWidth = true,
+  controlledChildren = true,
+}
+
+function WeakAuras.diff(ours, theirs, ignored)
```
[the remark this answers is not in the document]
emptyrivers: Laying groundwork for further iteration on the concept.

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
 
+local recurseDiff
+function recurseDiff(ours, theirs, ignored)
+  for key, ourVal in pairs(ours) do
+    local ignoredVal = type(ignored) == 'table' and ignored[key]
+    if ignoredVal ~= true then
+      if type(ourVal) == 'table' then
+        local theirVal = theirs[key]
+        if type(theirs[key]) ~= 'table' 
+        or recurseDiff(ourVal, theirVal, ignoredVal or nil) then
```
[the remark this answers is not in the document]
emptyrivers: What if I wanted to ignore that an animation is used or not, but not ignore the contents of said animation? I would need to have a deeper ignoredTable. This is groundwork for further iteration.

WeakAuras/Transmission.lua
emptyrivers: More groundwork for future iteration. If you don’t believe that there’s a use case for exporting a non-unique display, then I’d be happy to hear it.

WeakAuras/Transmission.lua
emptyrivers: Can you clarify? Do you mean to say that the code never executed anyways (that feedback wouldn’t make sense since this hunk was removed)?

WeakAuras/Transmission.lua
emptyrivers: I couldn’t see a path to make this code execute, and it’s otherwise a duplication of the branch just above it.

WeakAuras/Transmission.lua
emptyrivers: Update could, depending on user preferences, be less zealous than Diff. Diff is intended to find the differences (mainly so that an unmonitored version can be detected and skipped in the update process), while Update is supposed to (in a future iteration) change behavior based on user preferences.

WeakAuras/WeakAuras.lua
emptyrivers: Yes, that’s a hack that I forgot to improve before opening they PR.

WeakAuras/WeakAuras.lua
emptyrivers: Frankly, aura modifying in general needs to be refactored. I can skip DeleteOption and instead Rename here, but we really need a less brutish method of modifying data.

WeakAuras/Transmission.lua
InfusOnWoW: I believe there's no use case for that. And even if, then adding the parameter is easy.

WeakAuras/Transmission.lua
InfusOnWoW: Take a close look at all callers of ShowDisplayTooltip, some call it with data but no children (and compressed == nil)

WeakAuras/Transmission.lua
InfusOnWoW: Well, in the future we'll solve all kinds of problems. But not updating a change that diff detects sounds wrong to me.

WeakAuras/Transmission.lua
InfusOnWoW: The actual line I commented on was not moved above. I'm not sure whether that is actually a problem.

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
InfusOnWoW: Well, don't try to guess where the next iteration will take you. That's premature generalization that often doesn't work out anyway. Case in point, you passed the wrong variable and as I pointed out, even if you passed on the right variable, it would have the wrong semantics.

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
InfusOnWoW: So do you intend to implement that additional stuff for the first version?

WeakAuras/WeakAuras.lua
```diff
@@ -1642,9 +1642,127 @@ do
...
+    oldChildren = {}
+    for i, child in ipairs(children) do
+      if child.uid then uids[child.uid] = i end
+      oldChildren[i] = db.displays[child.id]
+    end
+  end
+  for id, existingData in pairs(db.displays) do
+    local uindex = existingData.uid and uids[existingData.uid]
+    if uindex == 0 then
+      match = true
```
Stanzilla: `match` is global at this point

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
emptyrivers: I can, but that will have to wait for a while.

WeakAuras/Transmission.lua
```diff
@@ -972,13 +1026,81 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
...
+              
+              data.parent = nil
+              importUpdatedData(data, matchedData)
+              coroutine.yield();
+
+              if children then
+                for index, childData in ipairs(children) do
+                  if not childData.same then
+                    local oldChildData =  matchedChildren[index]
+                    childData.parent = data.id
```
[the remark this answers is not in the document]
emptyrivers: I must be blind, because I don’t see it.

WeakAuras/Transmission.lua
```diff
@@ -972,13 +1024,81 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
...
+                    importUpdatedData(childData, oldData)
```
Stanzilla: oldData is not defined

WeakAuras/Transmission.lua
emptyrivers: It was a no-op, since both import and update handlers will ensure that controlledChildren has the correct children.

WeakAuras/Transmission.lua
```diff
@@ -1109,13 +1234,6 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
 
     local RegularGetData;
     if(children) then
```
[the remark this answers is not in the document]
emptyrivers: Can you explain what you mean? Do you want me to merge this with the loop that I moved the Decompression of the children to?

Stanzilla:
So, do you want to handle basic updating first and then cover options for it in a different PR or do you want to do all in one?

emptyrivers:
My plan was to get the basic behavior that exists in this PR to be merged, and then work on more sophisticated options in a different PR, for a couple reasons. First, the majority of users won’t ever touch the options, so a simple interface with no customizability works fine. Second, the sooner that the uid gets merged in, the sooner that the people who *make* the kinds of auras which get imported can make a new version of their aura which has the uid.

WeakAuras/Transmission.lua
```diff
@@ -1057,20 +1057,20 @@ function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compr
...
                     importUpdatedData(childData, oldChildData)
                     tinsert(data.controlledChildren, childData.id);
                     coroutine.yield()
                   else
-                    childData.same = nil
-                    -- parent may have changed, ensure it's still correct
-                    childData.parent = data.id
+                    -- skip importing
+                    matchedChildren[index].parent = data.id
+                    tinsert(data.controlledChildren, matchedChildren[indes].id)
```
Stanzilla: `indes` typo

emptyrivers:
@InfusOnWoW @Stanzilla, I believe this PR is in its final state, barring any bug fixes or requests to change things and a final squash into one commit. Feel free to leave a review, though I haven't done any rigorous testing (or really any testing yet at all).

emptyrivers:
See [this video](https://www.youtube.com/watch?feature=youtu.be&v=-qzQmfGkmDA) for a preview.

WeakAuras/WeakAuras.lua
```diff
@@ -3143,6 +3142,74 @@ function WeakAuras.CancelAnimation(region, resetPos, resetAlpha, resetScale, res
   end
 end
 
+function WeakAuras.GenerateDataStub(regionType, id)
```
InfusOnWoW: Given that two of the usages are read-only, wouldn't it better to have that table constant somewhere? The one place that needs a copy can make a copy.

WeakAuras/WeakAuras.lua
```diff
@@ -3143,6 +3142,74 @@ function WeakAuras.CancelAnimation(region, resetPos, resetAlpha, resetScale, res
...
+      subeventPrefix = "SPELL",
+      subeventSuffix = "_CAST_START",
+      debuffType = "HELPFUL",
+      names = {},
+      spellIds = {},
+      event = "Health",
+      unit = "player"
+    },
+    untrigger = {},
+    load = {
```
InfusOnWoW: Do we need all of that in the DataStub? Keeping that up to date with other places is unlikely to happen.

WeakAuras/Transmission.lua
```diff
@@ -160,86 +160,34 @@ function tableSubtract(minuend, subtrahend)
...
+  -- ends but YouTube doesn't seem to be having problems
+  local s = {}
+  for i=1,11 do
+    table.insert(s, bytetoB64[math.random(0, 63)])
+  end
+  return table.concat(s)
 end
 
 function CompressDisplay(data)
-  local copiedData = {};
```
InfusOnWoW: Why did you remove the copying before removing custom code?

Only the exported data should be cleaned, the actual auras should not be changed on export.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
 ChatFrame_AddMessageEventFilter("CHAT_MSG_INSTANCE_CHAT_LEADER", filterFunc)
 
-function ShowTooltip(content)
-  if(ItemRefTooltip.WeakAuras_Tooltip_Thumbnail) then
-    ItemRefTooltip.WeakAuras_Tooltip_Thumbnail:Hide();
-    ItemRefTooltip.WeakAuras_Tooltip_Thumbnail = nil;
-  end
-  if(ItemRefTooltip.WeakAuras_Tooltip_Thumbnail_Frame) then
-    ItemRefTooltip.WeakAuras_Tooltip_Thumbnail_Frame:Hide();
+local buttonAnchor =  CreateFrame("FRAME", "WeakAurasTooltipAnchor", ItemRefTooltip)
```
InfusOnWoW: Two sapces after =

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateAnimateCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"),
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateActionCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"),
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateLoadCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"),
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateConditionsCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"),
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateTriggerCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"),
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateDisplayCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"), 
+  CreateFrame("CHECKBUTTON", "WeakAurasUpdateTrivialCheckButton", buttonAnchor, "ChatConfigCheckButtonTemplate"), 
+}
+
+_G[checkButtons[1]:GetName().."Text"]:SetText(L["Make a copy"])
```
InfusOnWoW: Why can't the text be set in the same way as for the other checkboxes?

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+checkButtons[1].func = function() WeakAuras.UpdateTooltipButtons() end
+local buttonLabels = {
+  nil,
+  L["Animations"],
+  L["Actions"],
+  L["Load Conditions"],
+  L["Conditions"],
+  L["Trigger Settings"],
+  L["Display Settings"],
+  L["Trivial Settings"],
```
InfusOnWoW: Trivial settings is a bad name.

WeakAuras/WeakAuras.lua
emptyrivers: The stub should be be absolute minimum data which is common to all auras. I’ll admit I was a bit overzealous when writing it, mostly to stamp down on false positives in WeakAuras.diff. I’ll see how much I can trim it down without adversely affecting diff.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+    if import.diffs then
+      -- position checkButtons
+      local tabsChanged = import.diffs.tabsChanged
+      local update_preferences = WeakAurasSaved.update_preferences
+      for i, button in ipairs(checkButtons) do
+        button:Show()
+        if i ~= 1 then
+          if tabsChanged[button] then
+            button:Enable()
+            button:SetChecked(update_preferences and update_preferences[i] or true)
```
InfusOnWoW: Making update_preferences index via the int will make future changes to e.g. insert a checkbox pretty hard.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+buttonAnchor.import:SetScript("OnClick", function(self)
+    if not self.data then return end
+    -- build closure for DynFrame to use
+    HideUIPanel(ItemRefTooltip)
+    WeakAuras.LoadOptions()
+    WeakAuras.SetImporting(true)
+    buttonAnchor.showCode.codes = nil
+    buttonAnchor.thumbnail.currentThumbnail:Hide()
+    buttonAnchor.thumbnail.currentThumbnail = nil
+    -- TODO: find a way to do this better because uh
```
InfusOnWoW: I agree with you here.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+      local num = 2;
+      while WeakAurasSaved.displays[id] do
+        id = id.." "..num;
+        num = num + 1;
+      end
+      return id
+    end
+    -- upvalues for our closure, and also clear data so that the next import is not corrupted
+    local incomingData, incomingChildren, diffs = self.data, self.children, self.diffs
+    self.data, self.children, self.diffs = nil, nil, nil
```
InfusOnWoW: Clearing here should not be needed, and the comment seems to indicate you thought it was necessary here.

If you only clear in import, then not importing won't clear, and thus likely result in the same corrupted state.

WeakAuras/Transmission.lua
Stanzilla: "Size & Position"

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+            -- clear fields which the user did not want to update
+            local oldData, patch = unpack(diffs[0])
+            if patch then
+              for k in pairs(patch) do
+                local checkbutton = keyToButton[k]
+                if checkbutton and not checkbutton:GetChecked() then
+                  patch[k] = nil
+                end
+              end
+              if patch.id then -- ensure that the id is changed
```
InfusOnWoW: TODO: I don't understand this.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+          WeakAuras.PickDisplay(installedData.id)
+          WeakAuras.CloseImportExport()
+          WeakAuras.SetImporting(false)
+        end
+      )
+    )
   end
+)
+
+function ShowTooltip(info, ...)
```
InfusOnWoW: I think given the extra parameters names here does make the code more readable.

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
 
-function WeakAuras.ShowDisplayTooltip(data, children, icon, icons, import, compressed, alterdesc)
-  if(type(data) == "table") then
-    if(compressed) then
-      DecompressDisplay(data);
-      data.controlledChildren = nil;
+local ignoredForDiffChecking = {
+  -- note for future: it may be the case that we want
+  -- to ignore things not at the surface of a table
+  -- in which case recuseDiff will need to also recurse
```
InfusOnWoW: typo: recuseDiff

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+  init_started = true,
+  internalVersion = true,
+  controlledChildren = true,
+  parent = true,
+  expanded = true,
+}
+
+local deleted = {} 
+--[[
+    This is a filthy hack.
```
InfusOnWoW: That's not a filthy hack. deleted is just a magic value. :)

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+
+local recurseDiff
+function recurseDiff(ours, theirs)
+  local diff, seen, same = {}, {}, true
+  for key, ourVal in pairs(ours) do
+    if not ignoredForDiffChecking[key] then
+      seen[key] = true
+      local theirVal = theirs[key]
+      if type(ourVal) == "table" and type(theirVal) == "table" then
+        local diffVal = recurseDiff(ourVal, theirVal, circleRefs)
```
InfusOnWoW: circleRefs is not used

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+  end
+  local match = oldData ~= nil
+  if data.uid then uids[data.uid] = 0 end
+  local oldChildren
+  if children then
+    oldChildren = {}
+    for i, child in ipairs(children) do
+      if child.uid then uids[child.uid] = i end
+      local oldChild = WeakAuras.GetData(child.id)
+      if oldchild and (not child.uid or child.uid == oldChild.uid) then
```
InfusOnWoW: oldchild vs oldChild

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
-        tinsert(tooltip, {2, " ", childData.id, 1, 1, 1, 1, 1, 1});
-        scamCheck(codes, childData);
-        local childVersion = childData.internalVersion or 1;
-        highestVersion = max(highestVersion, childVersion);
-      end
-      if(#tooltip > 3) then
-        tooltip[4][2] = L["Children:"];
-      else
-        tinsert(tooltip, {1, L["No Children:"], 1, 1, 1});
+function WeakAuras.MatchData(data, children)
```
InfusOnWoW: I think this function should search differently for matching children. It currently can match children in different groups, if auras are moved outside of groups, or the id matching is used.

I think the right algorithm, first finds the parent via uid or id match. And then only looks for children in the child list of that match.

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+          info[i], info.added = "added", info.added + 1
+        else
+          info.deleted[oldChild.id] = nil
+          info.deletedCount = info.deletedCount - 1
+          local childDiff= WeakAuras.diff(oldChild, child)
+          info[i] = {oldChild, childDiff}
+          if childDiff then
+            modified = true
+            info.modified = info.modified + 1
+            if tabsChanged < 7 then
```
InfusOnWoW: Ugly magic constant, that's pretty hard to understand. (Also below)

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+    if children then
+      for i, child in ipairs(children) do
+        local oldChild = oldChildren[i]
+        if not oldChild then
+          modified = true
+          info[i], info.added = "added", info.added + 1
+        else
+          info.deleted[oldChild.id] = nil
+          info.deletedCount = info.deletedCount - 1
+          local childDiff= WeakAuras.diff(oldChild, child)
```
InfusOnWoW: Missing space

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
-            if(trigger.event == "Combat Log" and trigger.subeventPrefix and trigger.subeventSuffix) then
-              tinsert(tooltip, {2, L["Message type:"], (WeakAuras.subevent_prefix_types[trigger.subeventPrefix] or L["Undefined"]).." "..(WeakAuras.subevent_suffix_types[trigger.subeventSuffix] or L["Undefined"]), 1, 1, 1, 1, 1, 1});
             end
           else
-            tinsert(tooltip, {2, L["Trigger:"], L["Custom"], 1, 1, 1, 1, 1, 1});
+            info.unmodified = info.unmodified + 1
           end
         end
       end
+      return (modified or next(info.deleted)) and info or false
```
InfusOnWoW: "next(info.deleted)" should be equivalent to "info.deletedCount > 0" which I find easier to understand.

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
-      local size = #tooltip
-      tooltip[26] = {2, " ",  "[...]", 1, 1, 1, 1, 1, 1};
-      local nrOfChildren = children and #children or data.controlledChildren and #data.controlledChildren or 0
-      tooltip[27] = {1, string.format(L["%s total auras"], nrOfChildren), "", 1, 1, 1, 1};
-      for i = 28, size do
-        tooltip[i] = nil;
-      end
+local recurseUpdate
+function recurseUpdate(data, diff)
+  for k,v in pairs(diff) do
```
InfusOnWoW: the singular of diff is "chunk".

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+  end
+end
 
+function WeakAuras.Update(data, diff)
+  -- modifies the installed data such that it matches the pending import, sans user-specified options
+  -- diff is expected to be output generated by WeakAuras.diff
+  if not diff or not data then 
+    return data 
+  end
+  WeakAuras.DeleteOption(data)
```
InfusOnWoW: Doesn't that mean that the auras are added at the end. E.g. if I have a group with children A B C, where I update B C, then the result is B C A ?

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
-            local function findUnusedId(data)
-              local id = data.id
-              local num = 2;
-              while(WeakAurasSaved.displays[id]) do
-                id = data.id.." "..num;
-                num = num + 1;
-              end
-              data.id = id;
+            if match.deletedCount ~= 0 then
+              table.insert(tooltip, {1, L["%d auras deleted"]:format(match.deletedCount), .5, .5, .5})
```
InfusOnWoW: every line other than this uses tinsert

WeakAuras/Transmission.lua
```diff
@@ -730,283 +953,390 @@ local function scamCheck(codes, data)
...
+        end
+        if tabsChanged >= 7 then break end
       end
-      if(#tooltip > 3) then
-        tooltip[4][2] = L["Children:"];
-      else
-        tinsert(tooltip, {1, L["No Children:"], 1, 1, 1});
+    end
+    if oldData then
+      info[0] = {oldData, diff}
```
InfusOnWoW: Hmmmm, I prefer info.oldData[0], info.diff[0] = oldData, diff. As I think that leads to more readable code in other places.

WeakAuras/Transmission.lua
InfusOnWoW: Argh, that's too bad. I guess keep this, even if it's ugly.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+  elseif WeakAuras.IsImporting() then
+    import:SetText(L["Import in progress"])
+  elseif WeakAurasSaved.import_disabled then
+    import:SetText(L["Import Disabled"])
+  elseif import.data then  -- no need to enable the button if there's nothing to import
+    import:Enable()
+    if import.diffs then
+      -- position checkButtons
+      local tabsChanged = import.diffs.tabsChanged
+      local update_preferences = WeakAurasSaved.update_preferences
```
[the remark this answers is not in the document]
emptyrivers: Yes, that is a leftover from an earlier iteration. Depending on what happens with the make a copy button, I could make it work cleanly now or simply remove it, pending a future update to add the functionality.

WeakAuras/Transmission.lua
```diff
@@ -316,37 +270,318 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+    desc = false,
+    -- trivial
+    anchorPoint      = checkButtons[8],
+    xOffset          = checkButtons[8],
+    selfPoint        = checkButtons[8],
+    frameStrata      = checkButtons[8],
+    anchorFrameType  = checkButtons[8],
+    anchorFrameFrame = checkButtons[8],
+    yOffset          = checkButtons[8],
+    -- not super convinced that these are trivial
```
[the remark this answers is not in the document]
emptyrivers: Is that not covered by x/yOffset? I thought it was. Or maybe I don’t understand your question.

WeakAuras/Transmission.lua
emptyrivers: So, what would you prefer me to do? Add an OnHide handler to ButtonAnchor to clear the data?

WeakAuras/Transmission.lua
emptyrivers: Holdover from an earlier iteration where I didn’t have deletedCount. I’ll change it.

WeakAuras/Transmission.lua
emptyrivers: I don’t agree. To give an example, my Normalized Stagger aura has been copied into practically every brewmaster aura group on wago. You don’t want two copies of that, so the existing one should be updated.

WeakAuras/Transmission.lua
emptyrivers: That’s fine. I’ll rearrange it.

emptyrivers: That’s fine. I’ll rearrange it.

WeakAuras/Transmission.lua
emptyrivers: Are you talking about where I also assign the unused id to `incomingData`, even though i'm not going to use the `incomingData` to actually import anything? This is because `patch` is too small in scope to reach line 498, where I need the ID of the installed/updated/copied aura so that I can call `GetData` properly - this allows me to only write `GetData` once, instead of copying it into each branch of the if-else chain. As for why I need to call GetData at all, this was a workaround to a pretty nasty bug which caused the parent of an updated aura to think it did not have the updated aura as a child anymore.

WeakAuras/Transmission.lua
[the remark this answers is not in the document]
InfusOnWoW: You want to preserve your changes to the group position, but update any inner group position that the author did. That's maybe the author moved some stuff inside the group to make talents fit better.

emptyrivers:
Rebased to the tip of 8.0.

User-facing changes in this iteration:

- The "Make a copy" checkButton is gone. Instead, when no checkButtons are checked, then importing will create a copy.
- There are two new checkButtons, which will appear only when the user imports an update to a group. One controls whether or not new children will actually be added, and the other controls whether or not existing children with no counterpart in the imported data will be deleted. This is implemented partly due to request, and partly due to the fact that without the "Make a copy" checkButton, there were situations in which 
- Update preferences are saved to the SavedVariables - thus, if you wish to keep Size & Position settings intact, and not delete old children for many separate imports, you may do so (or any other permutation of settings).

For anybody reading this who would like to test, but is not comfortable enough with git to checkout this branch, [here](https://github.com/WeakAuras/WeakAuras2/files/2191365/weakauras_update_branch.zip) is WeakAuras with all of the changes.

InfusOnWoW:
There are a few bugs where apparently the child -> parent relationships aren't properly upated:

* Import this group:
d0YWfaGAeOzsvLmBPUjvj3ws7er2RA3i9tQkddb8BQmuQQsdhGdsvQJHswiIQwmvvXYj6HuvvEk0JrXZjyIiQOPszYeA6IlIOcxg01r1gPQQARiiBgLA7a6JuvP(mqFhH6WsmAeLXJOspJQOBHG60Kopczzuf(lvvyAuvrpRTXY2zuyKCcbowhfhf3glmrbcnPcccLE5eHtIWSiWOko63SgLqq1rnIaahbLooGrLgOKwAcbLcQJojphfvbcYPmrbchlDpkaB2(RtPzK8(mZNx(f5W)TXivqqOCBuGmhX(OJcK5i2h1BMOaHNrkVAsfeekfMeRX2ve3gtPH04X2yL3r3oNXuAinBJvEhD7CgLUgUnw5D0TZzuuzZwz4DiABSY7OBNZOSWa3gR8o625mQuqPJdyBSY7OBNZiaGasPH0Snw5D0TZzKDHjQJUnw5D0TZzKla9dMUiiS)CgHKlGcveko5pJmBtI12yHjkqOjvqqO0lNiCseMfbgJubbHYTrbYCe7JokqMJyFuVzIceEgBxrCBmLgsJhBJvEhD7CgtPH0Snw5D0TZzuuzZwz4DiABSY7OBNZOSWa3gR8o625mkDnCBSY7OBNZiaGasPH0Snw5D0TZzuPGshhW2yL3r3oNr2fMOo62yL3r3oNZOsduslnHGsb1rNKNJIQab5uMOaHJLUhLqq1rnIaahvghDKYRMubbHsHjXAufh9(mjp2glmrbcnPcccLE5eHtIWSiWyKkiiuUnkqMJyF0rbYCe7J6ntuGWZy7kIBJkfu64a2gR8o625mMsdPzBSY7OBNZOOYMTYW7q02yL3r3oNrzHbUnw5D0TZzu6A42yL3r3oNraabKsdPzBSY7OBNZi7ctuhDBSY7OBNZyknKgp2gR8o625CgvAGsAPjeukOo6K8Cucbvh1icaCuzC0rvCK8JuE1KkiiukmjwJIQab5uMOaHJLUptYZTXctuGqtQGGqPxor4KimlcmQ0aL0stiOuqD0j55yKkiiuUnkqMJyF0rbYCe7J6ntuGWZy7kIBJkfu64a2gR8o625mMsdPzBSY7OBNZOOYMTYW7q02yL3r3oNrzHbUnw5D0TZzu6A42yL3r3oNraabKsdPzBSY7OBNZyknKgp2gR8o625mYUWe1r3gR8o625CgfvbcYPmrbchlDpkHGQJAebaoQmo6iLxnPcccLctI1Oko6FNZiWjXYp9G15b

* Duplicate the last child C, name it D.

* Import the group again.
* Use "Delete Group and Children"
Observe that "D" remains.

=== Second bug ===
Import this group:
d8IZjaGAeW2KuPMjkrz2sDtjj3wIDIi7vTBK(PKYWKu1VPYqLeIHdWbLeDmuQfIsKfljOwojpeLO6PqpgfpNuMOKkmvktMunDXfLujxg01r1gLurBfLWMruBhqFusq(mqFhbDyQ6zssnAeY4LeWYqjDleORjjKopc1Pj(RKannjH6zVn63oJAJ1be4yzuFuFB0Zebi0uciiuvLthojcYU(rrFScXEubbfh1igaCeu54agfAGkQVzbuPjo6KQEux0iaNYebiC039OgSjxrCcnJSunZQvflRUQtBmkbeeQ2g1iYryn6OgrocRrRKjcq4zKYlMsabHkTjXESDE9TX4BinSUnw4DKTZzm(gsZ2yH3r2oNrLRHBJfEhz7Cg1fYKfgEhI3gl8oY25mQ8mWTXcVJSDoJcfu54a2gl8oY25mcaiG4BinBJfEhz7Cgj7zI4OBJfEhz7Cg5AWkit710UcFgHvaaEQouFw6mYSnj2BJEMiaHMsabHQQC6Wjrq21pgLaccvBJAe5iSgDuJihH1OvYebi8m2oV(2y8nKgw3gl8oY25mgFdPzBSW7iBNZOUqMSWW7q82yH3r2oNrLNbUnw4DKTZzu5A42yH3r2oNraabeFdPzBSW7iBNZOqbvooGTXcVJSDoJK9mrC0TXcVJSDoNrHgOI6BwavAIJoPQh1fncWPmrach9DpQGGIJAedaokmo6iLxmLaccvAtI9OOpw5zsSUn6zIaeAkbeeQQYPdNebzx)yuciiuTnQrKJWA0rnICewJwjteGWZy786BJcfu54a2gl8oY25mgFdPzBSW7iBNZOUqMSWW7q82yH3r2oNrLNbUnw4DKTZzu5A42yH3r2oNraabeFdPzBSW7iBNZizptehDBSW7iBNZy8nKgw3gl8oY25CgfAGkQVzbuPjo6KQEubbfh1igaCuyC0rrFKLgP8IPeqqOsBsSh1fncWPmrach9DFMu1BJEMiaHMsabHQQC6Wjrq21pk0avuFZcOstC0jv9yuciiuTnQrKJWA0rnICewJwjteGWZy786BJcfu54a2gl8oY25mgFdPzBSW7iBNZOUqMSWW7q82yH3r2oNrLNbUnw4DKTZzu5A42yH3r2oNraabeFdPzBSW7iBNZy8nKgw3gl8oY25ms2ZeXr3gl8oY25Cg1fncWPmrach9DpQGGIJAedaokmo6iLxmLaccvAtI9OOpYYptQI3g9mracnLaccvv50HtIGSRFmkbeeQ2g1iYryn6OgrocRrRKjcq4zSDE9TX4BinSUnw4DKTZzm(gsZ2yH3r2oNrDHmzHH3H4TXcVJSDoJkpdCBSW7iBNZOY1WTXcVJSDoJaaci(gsZ2yH3r2oNrHcQCCaBJfEhz7Cgj7zI4OBJfEhz7CoJcnqf13SaQ0ehDsvpQlAeGtzIaeo67Eubbfh1igaCuyC0rkVykbeeQ0Me7rrFSootQIUn6zIaeAkbeeQQYPdNebzx)yuciiuTnQrKJWA0rnICewJwjteGWZy786BJcfu54a2gl8oY25mgFdPzBSW7iBNZOUqMSWW7q82yH3r2oNrLNbUnw4DKTZzu5A42yH3r2oNraabeFdPzBSW7iBNZizptehDBSW7iBNZy8nKgw3gl8oY25CgfAGkQVzbuPjo6KQEubbfh1igaCuyC0rrFSUgP8IPeqqOsBsSh1fncWPmrach9DFMuDVn6zIaeAkbeeQQYPdNebzx)Oqdur9nlGknXrNu1JrjGGq12OgrocRrh1iYrynALmracpJTZRVnkuqLJdyBSW7iBNZy8nKMTXcVJSDoJ6czYcdVdXBJfEhz7CgvEg42yH3r2oNrLRHBJfEhz7Cgbaeq8nKMTXcVJSDoJX3qAyDBSW7iBNZizptehDBSW7iBNZzux0iaNYebiC039OcckoQrma4OW4OJuEXuciiuPnj2JI(yDEoJaNe7kMv2Nh

This adds a group: G1 with children ABCDEF

* Import this group: 
dWYWfaGAeOzIOQMTu3eLQBlPDIi7vTBK(jvPHHOYVPYqbjXWPQoicYXOkwiivTyqsA5e9uOhJINtWebP0uPmzcnDXfruLldCDuTrqQSveuBgLSDq8rqk6ZG67iuldLYOrugpiP6WsCleWPjDEeYZaj(lifMgiP8E2glBNrHrOfazSokokUnwyIcbysfggiz3jcMeb8qUrvCeAY2OeaRoQrKpyew64(JknajT0egifuhDsqzuufiiNYefcyS09OaOzbvCknJqVxZ8Yo5tEqNngPcddKBJcK5i2lDuGmhXEPeIjkeWzKYRMuHHbsHj5zSDfXTXuAanSTnw5D0TZzmLgqZ2yL3r3oNrPRbBJvEhD7CgfvwSugEhI2gR8o625mklmGTXkVJUDoJkfw64(BJvEhD7Cg9H4NsdOzBSY7OBNZiRctuhDBSY7OBNZixaany6IGWq1ZiaQ7xOIaXH(ZiZ2K8SnwyIcbysfggiz3jcMeb8qUrLgGKwAcdKcQJojOmgPcddKBJcK5i2lDuGmhXEPeIjkeWzSDfXTrLclDC)TXkVJUDoJP0aA2gR8o625mkQSyPm8oeTnw5D0TZzuwyaBJvEhD7CgLUgSnw5D0TZz0hIFknGMTXkVJUDoJP0aAyBBSY7OBNZiRctuhDBSY7OBNZzuufiiNYefcyS09OeaRoQrKpyuzC0rkVAsfggifMKNrvCe6otITTXctuiatQWWaj7orWKiGhYngPcddKBJcK5i2lDuGmhXEPeIjkeWzSDfXTXuAanSTnw5D0TZzmLgqZ2yL3r3oNrrLflLH3HOTXkVJUDoJYcdyBSY7OBNZO01GTXkVJUDoJ(q8tPb0Snw5D0TZzuPWsh3FBSY7OBNZiRctuhDBSY7OBNZzuPbiPLMWaPG6OtckJIQab5uMOqaJLUhLay1rnI8bJkJJos5vtQWWaPWK8mQIJq7zsqzBSWefcWKkmmqYUtemjc4HCJrQWWa52OazoI9shfiZrSxkHyIcbCgBxrCBuPWsh3FBSY7OBNZyknGMTXkVJUDoJIklwkdVdrBJvEhD7CgLfgW2yL3r3oNrPRbBJvEhD7Cg9H4NsdOzBSY7OBNZiRctuhDBSY7OBNZyknGg22gR8o625CgvAasAPjmqkOo6KGYOeaRoQrKpyuzC0rvCK8gP8QjvyyGuysEgfvbcYPmrHaglDFoJqMKhOgBEop

This imports a group G2: with children DEF, which get renamed to D 2, E 2, F 2 on import

* Import the group above again:
Notice that the dialog shows "Update Auras", and somehow the children EDF are moved from G1 into group G2.

Trying to delete all groups will eventually result in a error message. 
*  E.g. Delete G1, which deletes EDF from G2.
* Right click on G2, "attempt to index a nil value".

Those are the bugs I suspected when I made the comment that you should look for children only in the matching group.

InfusOnWoW:
Another bug found by Rehok on the Discord:
*Import https://wago.io/HkQwid3Af
*Import it again
Observe that there's overlapping text in the tooltip.

InfusOnWoW:
And another bug:
* Create a new icon
* Select Set tooltip description from the context menu
* Close the text entry
* Import a aura
Observe the the tooltip description text box is visible. The same bug can be reproduced with "Set URL" on auras tha have a embedded url.

WeakAuras/Transmission.lua
```diff
@@ -730,454 +892,546 @@ local function scamCheck(codes, data)
...
-      else
-        tinsert(tooltip, {1, L["No Children:"], 1, 1, 1});
+  -- match the parent/single aura (if no children)
+  local oldParent = WeakAuras.GetData(data.id)
+  if oldParent then
+    if data.uid and oldParent.uid and data.uid ~= oldParent.uid
+    or (children and (not oldParent.controlledChildren or oldParent.parent)) then
+      oldParent = nil
+    end
+  end
```
InfusOnWoW: This can match a standalone aura with a existing group if the id matches.

That is, create a group "New", import a aura named "New" and the offered dialog will offer to update the group with a standalone aura.

If the import has no children, then oldData should have no children either.

InfusOnWoW:
Import this group: G with two children A and B.
dWsAeaGAeWmruLzd6MuPCBvzNiYELDJ0pPsggcQFtvdLkcdxLCqQOogkzHiqlgrvTCIEk0JrXZrLjsfjtL0Krvtx5IuH0LbUoHnsfkBfbzZOuBxL6JurQpRQ(oc1HvXOrugpve9mevUfvQonLZJqwgvWFPcX0OcvhR0WtAlKl0Pa3HVq(q(0WdZSBGkT)hiDZZdIK7SiCOXh60HsW38uLOlq4x6fxHgDaj9ajeqYzEAKixiVXrabLz2ni8aHHCaiBNWB0fsqxQ6YnYZrDmnCs7)bY0qoY8e7IgYrMNyxuNzMDdAHuXtL2)dKCrIvi0F4td3bcOZH0WNaolT1c3bcOln8jGZsBTqPhcsdFc4S0wlK3yZ2yeWruA4taNL2AHYddin8jGZsBTqJ(LEXvA4taNL2AHx3x7ab0Lg(eWzPTwi7dZmpnn8jGZsBTqbhWryGhoUi)wiWjVouEaFeSfYKgjwPHhMz3avA)pq6MNhej3zr4WjT)hitd5iZtSlAihzEIDrDMz2nOfc9h(0WDGa6Cin8jGZsBTWDGa6sdFc4S0wlK3yZ2yeWruA4taNL2AHYddin8jGZsBTqPhcsdFc4S0wl86(AhiGU0WNaolT1cn6x6fxPHpbCwARfY(WmZttdFc4S0wRfA0bK0dKqajN5PrICH8ghbeuMz3GWdegkbFZtvIUaHgJNgsfpvA)pqYfjwHgFOZTi5qAiVXrabLz2ni8aHHhMz3avA)pq6MNhej3zr4qJoGKEGeci5mpnsKlCs7)bY0qoY8e7IgYrMNyxuNzMDdAHq)Hpn0OFPxCLg(eWzPTw4oqaDPHpbCwARfYBSzBmc4ikn8jGZsBTq5HbKg(eWzPTwO0dbPHpbCwARfEDFTdeqxA4taNL2AHSpmZ800WNaolT1c3bcOZH0WNaolT1AHsW38uLOlqOX4PHgFibdPINkT)hi5IeRwl8osSCChy1ka

Import this group G with one child A:
deJudaGAivTnivMjqKzd1nbcFte2jQYEv7MOFIQAyeQXHk1qjKA4qYbPk6yI05Ok1cjKSyiLA5u5PilJQYZrXeHuYuPmzsmDPlce1TH4zuLCDsTrQcBfvWMjPTduFesrNwyAIOVJk5ZO0HbmAurJhvOld6wuvDncX5bs)wupMG)cPWNUnbSDNyMqli4jKjLjLTjoYxvLb29WFsV5ozsrs9XTyV9XDc0X9Q(tkYeO8vvzGDp8NW7eIKMkwSpVejrc0jw8Q(tMycqObyO5cwwOdezf488NkEkuMqZjhKnYsduuWjwxwJAkKf6KayoaDmrwopVMucg0RLcnadNaW4jgiwv05q2jrX3m(GaKazpSP6cwwOBBIHZmx8LtmCM5IV0tHgGH3jPgXCbll0XmV0jCgqzBQayOS(2Mq04gB37ubWqz3Mq04gB37KlJHBtiACJT7DsjuvdbnUGUnHOXn2U3jhGaCBcrJBSDVtHK1L1O2Mq04gB37ekWOkagk72eIg3y7ENubeAKLBtiACJT7DsZardbmadZO9DcYruasfOYI6ojSnV0TjaHgGHMlyzHoqKvGZZFQ4P6cwwOBBIHZmx8LtmCM5IV0tHgGH3jCgqzBQayOS(2Mq04gB37ubWqz3Mq04gB37KsOQgcACbDBcrJBSDVtoab42eIg3y7ENCzmCBcrJBSDVtOaJQayOSBtiACJT7DkKSUSg12eIg3y7ENubeAKLBtiACJT79ofYcDsamhGoMilNNxtkbd61sHgGHtay8KdYgzPbkk4uiKLtsnI5cwwOJzEPtHYKN37e45LM0x69b

In the dialog choose "Keep unmatched children"
Observe: B the unmatched child is no longer part of the group.

WeakAuras/Transmission.lua
```diff
@@ -977,43 +974,47 @@ function WeakAuras.MatchData(data, children)
...
       end
     end
 
     -- confirm uid matches, find tentative matches
-    local tentativeMatches = {}
+    local tentativeMatches, confirmedNewMatches, confirmedOldMatches = {}, {}, {}
     for index, newChild in ipairs(children) do
       if newChild.uid then
         local matchedID = UIDtoID[newChild.uid]
         if matchedID then
```
InfusOnWoW: Even if newChild has uid, if there's no uid match, you probably want to try a id-match. That is the else branch of the newchild.uid branch should be run if there's no matchedId. (Though maybe check if the tentaitive old match has a uid and then reject that.)

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,152 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+        data.uid = nil
+      end
+      data.id = WeakAuras.FindUnusedId(data.id)
+      WeakAuras.Add(data)
+    end
+  elseif not data then -- unmatched old child which is being kept
+    data = oldData
+  else
+    patch = patch or {}
+      for key in pairs(patch) do
```
InfusOnWoW: Indentation error.

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,152 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+        local checkButton = keyToButton[key]
+        if checkButton and not checkButton:GetChecked() then
+          patch[key] = nil
+        end
+      end
+    if patch.id then
+        patch.id = WeakAuras.FindUnusedId(patch.id)
```
InfusOnWoW: Indentation error. Also there's a corner case here, which I think we should do nothing about and just ignore, but which I'll try to explain anyway:

Imagine a group:
G A B with uids 0 1 2.

Importing a group:
G B A with uids 0 1 2

That is in the new version of the group the author renamed A to B, and B to A. For both A and B, patch.id will be set.

On importing B, the code will run FindUnusedId("B"), which finds the not yet renamed old B, and thus chooses "B 2" as the new name.

Again, I think there are limits to what should be supported, and this falls outside of it.

WeakAuras/Transmission.lua
```diff
@@ -730,454 +889,570 @@ local function scamCheck(codes, data)
...
     end
 
-    if(hasDescription) then
-      tinsert(tooltip, {1, "\""..data.desc.."\"", 1, 0.82, 0, 1});
+    -- confirm uid matches, find tentative matches
+    local tentativeMatches, confirmedNewMatches, confirmedOldMatches = {}, {}, {}
+    for index, newChild in ipairs(children) do
+      if newChild.uid then
+        local matchedID = UIDtoID[newChild.uid]
+        if matchedID then
```
InfusOnWoW: Even if a new child has a uid, if there's no uid match, there should be a fallback to a id match.

Essentially the code of the else branch of the check 3 lines up must be done if there's no matchedId.

emptyrivers:
Rebased on top of master to fix a conflict.

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,250 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+  WeakAuras.NewDisplayButton(installedData[0])
+  coroutine.yield()
+
+  -- handle children, if there are any
+  if #imports > 0 then
+    local parentData = installedData[0]
+    local preserveOldOrder = checkButtons["childorder"]:GetChecked()
+
+    local i = 1
+    local map = preserveOldOrder and info.oldToNew or info.newToOld
```
Stanzilla: accessing undefined variable 'info'

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,250 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+        local nextInsert, highestInsert, toInsert, newToOld = 1, 0, {}, indexMap.newToOld
+        for newIndex, data in ipairs(info.newData) do
+          local oldIndex = newToOld[newIndex]
+          if oldIndex then
+            nextInsert = oldIndex + 1
+          else
+            if nextInsert > highestInsert then
+              highestInsert = nextInsert
+            end
+            toinsert[newIndex] = toInsert[newIndex] or {}
```
Stanzilla: mutating non-standard global variable 'toinsert'

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,250 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
+              tinsert(installedData, index)
+              coroutine.yield()
+            end
+          end
+        end
+      elseif keepDeleted then -- we have existing children which need migrating
+        -- same algorithm as before, but mirror image
+        local nextInsert, highestInsert, toInsert, newToOld = 1, 0, {}, indexMap.oldToNew
+        for oldIndex, data in ipairs(info.oldData) do
+          local newIndex = oldToNew[oldIndex]
```
Stanzilla: accessing undefined variable 'oldToNew'

WeakAuras/Transmission.lua
```diff
@@ -730,454 +987,599 @@ local function scamCheck(codes, data)
...
-                if(#trigger.names > 0) then
-                  if(#trigger.names > 1) then
-                    left = L["Auras:"];
-                  else
-                    left = L["Aura:"];
-                  end
-                end
-              end
+    end
+    return true
```
Stanzilla: unreachable code

WeakAuras/Transmission.lua
```diff
@@ -730,454 +987,599 @@ local function scamCheck(codes, data)
...
+      if oldIndex then
+        local oldChild = info.oldData[oldIndex]
+        if lastMatch and lastMatch > oldIndex then
+          lastMatch = nil
+          info.activeCategories[checkButtons["childorder"]] = true
+        else
+          lastMatch = oldIndex
+        end
+      end
+      local childDiff = WeakAuras.diff(oldChild, newChild)
```
Stanzilla: accessing undefined variable 'oldChild'

WeakAuras/Transmission.lua
```diff
@@ -730,454 +987,599 @@ local function scamCheck(codes, data)
...
+      if childDiff then
+        info.modified = info.modified + 1
+        info.diffs[index] = childDiff
```
Stanzilla: accessing undefined variable 'index'

WeakAuras/Transmission.lua
```diff
@@ -730,454 +987,599 @@ local function scamCheck(codes, data)
...
     if(received.m == "d") then
+      -- prepare data to be shown as a tooltip and slotted into pendingData
       tooltipLoading = nil;
-      local data = received.d;
-      WeakAuras.ShowDisplayTooltip(data, received.c, received.i, received.a, "unknown", true)
+      local data, child, icon, icons = received.d, received.c, received.i, received.a
+      DecompressDisplay(data)
+      WeakAuras.validate(data, regionTypes[data.regionType].default)
+      WeakAuras.Modernize(data)
+      if children then
```
Stanzilla: accessing undefined variable 'children'

Stanzilla:
We need to agree on the UX flow here, what I've been saying from the start is that the user has to have the SAME experience with this PR merged as they currently have by default.

This means:
* No checkbox is checked
* The import button says "Import" (or "Import as Copy", not sure here)
* The text is worded in a way that communicates that everything you can check there is optional, so "Please choose which settings you want to KEEP" not "you want to UPDATE" since that would be different from current behaviour, the default will overwrite everything as before.

emptyrivers:
So, wait. I’m confused. The current behavior is that (by default) all checkboxes are left unchecked, the import button functions says “Import As Copy”, and it functions exactly as before - importing the group will not attempt to merge the two groups. This matches your specifications, as best as I can tell.

You want the wording of the tooltip to suggest that enabling a checkbox will cause that category to not be updated, but how exactly does that jive with your other two requirements? If enabling a checkbox would cause it to not be updated, then surely the default should be every thing checked causes the group to not be merged, no?

Stanzilla:
You're right, mixed thoughts :)

WeakAuras/Transmission.lua
```diff
@@ -316,39 +270,287 @@ ChatFrame_AddMessageEventFilter("CHAT_MSG_BN_WHISPER_INFORM", filterFunc)
...
-    ItemRefTooltip.WeakAuras_Desc_Box:Hide();
+end
+
+-- Ideally, we would not add an __index method to this
+-- But that would greatly bloat the update_category for display
+-- A better solution would be to refactor the data
+-- Such that all of the display tab settings are in their own subtable
+setmetatable(keyToButton,{__index = function(_, key) return key and checkButtons["display"] end,})
+
+local function install(data, oldData, patch, forceCopy)
```
InfusOnWoW: This functions should adjust hybridTables as needed, shouldn't it?

WeakAuras/Transmission.lua
```diff
@@ -730,454 +1023,602 @@ local function scamCheck(codes, data)
...
-      for i = 28, size do
-        tooltip[i] = nil;
-      end
-    end
+  if not oldParent then return nil end
 
-    local hasDescription = data.desc and data.desc ~= "";
-    local hasUrl = data.url and data.url ~= "";
-    local hasVersion = data.version and data.version ~= "";
+    info.oldData[0] = oldParent
```
InfusOnWoW: Whitespace error.

WeakAuras/Transmission.lua
emptyrivers: Can you explain what you mean?

PR #453, opened 2018-06-18
fix CheckPvpTalentByIndex nil index error

mrbuds:
check C_SpecializationInfo.GetPvpTalentSlotInfo doesn't return nil

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -495,11 +495,14 @@ function WeakAuras.CheckPvpTalentByIndex(index)
     local checkTalentId = talentSlotInfo.availableTalentIDs[index];
     return talentSlotInfo.selectedTalentID == checkTalentId;
   else
-    local checkTalentId = C_SpecializationInfo.GetPvpTalentSlotInfo(2).availableTalentIDs[index - 3];
-    for i = 2, 4 do
-      local talentSlotInfo = C_SpecializationInfo.GetPvpTalentSlotInfo(i);
-      if talentSlotInfo and (talentSlotInfo.selectedTalentID == checkTalentId) then
-        return true;
+    local checkTalentSlotInfo = C_SpecializationInfo.GetPvpTalentSlotInfo(2)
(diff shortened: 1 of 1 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -491,13 +491,14 @@ end
 function WeakAuras.CheckPvpTalentByIndex(index)
   if (index <= 3) then
     local talentSlotInfo = C_SpecializationInfo.GetPvpTalentSlotInfo(1);
+    if not talentSlotInfo return false
     local checkTalentId = talentSlotInfo.availableTalentIDs[index];
     return talentSlotInfo.selectedTalentID == checkTalentId;
   else
     local checkTalentId = C_SpecializationInfo.GetPvpTalentSlotInfo(2).availableTalentIDs[index - 3];
```
InfusOnWoW: The error message seems to point to line 497. Which is this line. Seems GetPvpTalentSlotInfo(2) returned nil here.

emptyrivers: Additionally, the added line is malformed - it should be `if not talentSlotInfo then return false end`.

mrbuds: i sent a patch to fix this

WeakAuras/Prototypes.lua
```diff
@@ -491,13 +491,14 @@ end
 function WeakAuras.CheckPvpTalentByIndex(index)
   if (index <= 3) then
     local talentSlotInfo = C_SpecializationInfo.GetPvpTalentSlotInfo(1);
+    if not talentSlotInfo then return false end
     local checkTalentId = talentSlotInfo.availableTalentIDs[index];
     return talentSlotInfo.selectedTalentID == checkTalentId;
   else
     local checkTalentId = C_SpecializationInfo.GetPvpTalentSlotInfo(2).availableTalentIDs[index - 3];
```
InfusOnWoW: This needs to be checked too.

Issue #456, opened 2018-06-21
Dynamic Group - Individual Element Positioning Resetting


PersonalProxy:
**Describe the bug**
A dynamic group is collapsing the position of individual elements to it's frame center.
They should be dynamically positioning in a circle as the options are set.

**Do you have an error log of what happened?**
No errors reported. BugSack installed and empty.

**To Reproduce**
Immediately upon login, the WeakAura misbehaves.
I do not recall a specific point in creating it where it started malfunctioning.

**Expected behavior**
The individual elements should align to the set option, which is currently "circle" with "Animated expand and collapse" set. As it stands, they do not. Other similarly setup dynamic groups operate without issue.

**Screenshots**
[GIF link](https://i.imgur.com/nDzg3mo.gifv). Seeing is believing.,

**Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?**
Of course, yes.

**Which version of WeakAuras are you using?**
2.5.12, on WoW Legion (Live)

**Was it working in a previous version? If yes, which was the last good one?**
Did not use on a previous version.

**Additional context**
I completely removed my WeakAura saved variables and then imported the offending string as the only WA. It still misbehaves.

**WeakAura String:**

[...]

PR #457, opened 2018-06-21
ticket #456

emptyrivers:
If a child of a dynamic group has anchorFrameType of SELECT and anchorFrameFrame refers to a frame which doesn't exist, then WeakAuras will get stuck in a loop trying to anchor, which eventually causes the group to fail to control its children. We can prevent this by forcing GetAnchorFrame to detect a dynamicgroup parent.


InfusOnWoW:
Can you change the commit message to be more descriptive? E.g.
"Fix anchored auras in dynamic groups"

emptyrivers:
Done.

Sent from Mail for Windows 10

PR #465, opened 2018-06-27
Add functions commonly used by custom WA authors

asakawa-k:
WeakAuras.UnitAura, WeakAuras.UnitBuff, WeakAuras.UnitDebuff - mimicking functionality from pre-8.0
WeakAuras.GroupMembers - Very popular in Legion.

This would be a bit of a step off the normal path so I totally understand if providing stuff for custom WA authors isn't something you want to get into, or if the functions I added simply aren't strong enough that you want them to be carried by your addon. 
However, I thought that if I throw a PR up then it at least sparks the discussion and you can always reject the PR if you're not keen.

### DIFF
--- a/WeakAuras/HelperFunctions.lua
+++ b/WeakAuras/HelperFunctions.lua
@@ -1,0 +1,50 @@
+local UnitAura = UnitAura
+-- Unit Aura functions that return info about the first Aura matching the spellName or spellID given on the unit.
+local WA_GetUnitAura = function(unit, spell, filter)
+    for i = 1, 40 do
+        local name, _, _, _, _, _, _, _, _, spellId = UnitAura(unit, i, filter)
+        if not name then return end
+        if spell == spellId or spell == name then
+            return UnitAura(unit, i, filter)
+        end
+    end
+end
+local WA_GetUnitBuff = function(unit, spell, filter)
+    return WA_GetUnitAura(unit, spell, filter)
+end
+local WA_GetUnitDebuff = function(unit, spell, filter)
+    filter = filter and filter.."|HARMFUL" or "HARMFUL"
+    return WA_GetUnitAura(unit, spell, filter)
+end
+
+-- Function to assist iterating group members whether in a party or raid.
+local WA_IterateGroupMembers = function(reversed, forceParty)
+   local unit  = (not forceParty and IsInRaid()) and 'raid' or 'party'
+   local numGroupMembers = (forceParty and GetNumSubgroupMembers()  or GetNumGroupMembers()) - (unit == "party" and 1 or 0)
+   local i = reversed and numGroupMembers or (unit == 'party' and 0 or 1)
+   return function()
+      local ret
+      if i == 0 and unit == 'party' then
+         ret = 'player'
+      elseif i <= numGroupMembers and i > 0 then
+         ret = unit .. i
+      end
+      i = i + (reversed and -1 or 1)
+      return ret
+   end
+end
+
+-- Wrapping a unit's name in its class colour is very common in custom Auras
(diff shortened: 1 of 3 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -4756,3 +4756,37 @@ function WeakAuras.PrintProfile()
   end
   print("--------------------------------");
 end
+
+-- Unit Aura functions that return info about the first Aura matching the spellName or spellID given on the unit. 
+function WeakAuras.UnitAura(unit, spell, filter)
```
InfusOnWoW: Apparently you wanted local function GetUnitAura here.

InfusOnWoW:
I think I prefer to keep them in a separate namespace, as not everything in WeakAuras should be called by the user. (Or actually most shouldn't.)

Stanzilla:
I'd probably want them in an extra file.

WeakAuras/HelperFunctions.lua
```diff
@@ -0,0 +1,39 @@
...
+          local name, _, _, _, _, _, _, _, _, spellId = UnitAura(unit, i, filter)
+          if not name then return end
+          if spell == spellId or spell == name then
+              return UnitAura(unit, i, filter)
+          end
+      end
+  end
+  ,
+  GetUnitBuff = function(unit, spell, filter)
+      return GetUnitAura(unit, spell, filter)
```
emptyrivers: Ensure that the filter either doesn't exist or contains HELPFUL. Also, GetUnitAura here refers to a nonexistent global. Instead, use `return WeakAuras.helperFunction.GetUnitAura`.

WeakAuras/HelperFunctions.lua
```diff
@@ -0,0 +1,39 @@
...
+  end
+  ,
+  GetUnitDebuff = function(unit, spell, filter)
+      filter = filter and filter.."|HARMFUL" or "HARMFUL"
+      return GetUnitAura(unit, spell, filter)
```
emptyrivers: GetUnitAura is a nonexistent global here as before

WeakAuras/HelperFunctions.lua
```diff
@@ -0,0 +1,39 @@
...
+          end
+      end
+  end
+  ,
+  GetUnitBuff = function(unit, spell, filter)
+      return GetUnitAura(unit, spell, filter)
+  end
+  ,
+  GetUnitDebuff = function(unit, spell, filter)
+      filter = filter and filter.."|HARMFUL" or "HARMFUL"
```
emptyrivers: Note that this is vulnerable to someone going `GetUnitDebuff("player", "some spell name",  "HELPFUL")`. I have no idea if this is worth guarding against, or if it should just be garbage in garbage out.

emptyrivers:
It may well be worth moving all of the `aura_env` support code into `HelperFunctions.lua` at some point, in part of the ongoing battle to keep these files at a manageable size.

InfusOnWoW:
As TheDanW confirmed that he'll add something to wow after all, we'll wait a few more days to see what exactly they added.

asakawa-k:
Yeah. Fingers tightly crossed. 

If they do put the functionality back in to UnitAura, I'd like to keep the discussion alive regarding adding useful functions for custom WA authors. But at that point it will be significantly less important/urgent and, if you were open to it, there may be a better way to implement it.

asakawa-k:
Yeah, that's the kind of thing that could be done properly if there isn't the urgency of the pre-patch requiring something in place.

InfusOnWoW:
Can you squash all the commits? I think something went wrong and you duplicated a commit or so.

InfusOnWoW:
Thanks.

PR #472, opened 2018-07-06
Add Scale option to groups

sragia:
Basically allows user to change groups scale without need of going to each of groups elements and sizing/positioning them
Useful for WA groups that are too large and only needs sizing down

I guess implements this [issue](https://www.wowace.com/projects/weakauras-2/issues/827) too


WeakAurasOptions/RegionOptions/Group.lua
```diff
@@ -492,6 +492,23 @@ local function createOptions(id, data)
...
+      name = L["Group Scale"],
+      order = 45,
+      softMin = 0.05,
+      softMax = 2,
+      bigStep = 0.05,
+      get = function()
+        return data.scale or 1
+      end,
+      set = function(info, v)
+        data.scale = v
```
InfusOnWoW: You might want to move the group's anchor point here, so that the anchor point appears to not move.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3481,9 +3484,9 @@ function WeakAuras.PositionOptions(id, data, hideWidthHeight, disableSelfPoint)
         end
         return not data.anchorPoint:find("LEFT")
       end,
-      get = function() return data.xOffset end,
+      get = function() return data.xOffset*GetScale() end,
       set = function(info, v)
-        data.xOffset = v;
+        data.xOffset = v*(1/GetScale());
```
InfusOnWoW: I'm curious why you did this? I would have expected the unscaled values here.

emptyrivers:
What about dynamic groups?

WeakAurasOptions/WeakAurasOptions.lua
sragia: Problem here was that as you scale something, their coordinate system change. So 100px offset at 0.8 scale would look like 80px. So if you want align 2 groups (1 scaled and other isnt) you wouldn't be able to just use same offsets (assuming same anchor point). 
Also movers used scaled values so a bit confusing there when mover is showing one result, while options another.
Basically just to keep using same coordinate system for user.

sragia:
Didn't add this to dynamic groups as personally don't see it being as useful there. Most dynamic groups usually are of the same region type so can be quickly resized by changing all region sizes. Also from the quick look dynamic group elements doesn't seem to be actually parented to it, so just :SetScale() on group wouldn't work.
But definitely can look into adding it to them too if you think it should be there.

emptyrivers:
> Didn't add this to dynamic groups as personally don't see it being as useful there.

I don't think lack of personal experience is a good reason to not include a feature.

> Also from the quick look dynamic group elements doesn't seem to be actually parented to it, so just :SetScale() on group wouldn't work.

Well, children of a regular group need not be parented to the group aura, and that's a user-facing option. You'll need to handle that somehow as well.

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW: But the sizes of the auras use the changed coordinate system, right? So aligning in a group is now harder? That feels like the wrong solution.

The mover can be fixed.

sragia:
Reworked way to scale groups and also added option to Dynamic Groups. Instead of using :SetScale() I am changing every children's offset,size,text size,animations. All the region properties that are being changed are stored in WeakAuras.region_scale_elements in Types.lua. It might not be the best place for it so feel free to point me to a better place. 
So scaling thing "manually" gets rid of problems with coordinates and children that aren't parented to group
One thing that I could see problems with is with adding/removing auras from group. Basically should aura be sized to group's scale/reverted to original scale? Right now it stays the same. I am more inclined to leave it like it is because if user is creating new aura and is changing it's size, animations or what not before adding it to group, all of that would be lost. Sure it's scaled but I can't imagine user setting size of aura with groups scale in mind. If user is removing aura it's probably already the size user wants it to be so scaling it back to original would only be annoying. Basically would love input to that

emptyrivers:
Hmm, this last commit degrades performance whilst changing the scale noticeably, which is a shame, since I can't see any way around faking SetScale so that it works without needing auras to be parented to the group, without also piling a load of kludges into the addon as well. This may just end up being a necessary evil.

InfusOnWoW:
So to expand a bit on to that, the feature isn't as trivial to implement as it first looked like.

I think the questions that drive the design are:

Should a group scaling option apply to all children of a group. And how does that mesh with auras that are attached to a different parent? Or with auras that are in different groups but attached to the group? 

Ideally the group's scaling should apply to all children even if they are attached to a different parent. But, after considering the many edge cases, my opinion is that it should simply use wow's scaling. That means a few edge cases might work suprisingly, but it's far, far simpler to implement. I think trying to apply the scaling to all children individually has too many edge cases.

The second big question is, should the settings for scaled auras be unscaled or scaled values? Imho I'd rather keep the complexity down and not show the scaled values, but show the unscaled ones.

emptyrivers:
Perhaps the option should only appear when a group has no fake children? Then the original gunctionality would suffice (at least for non dynamic groups, anyways).

InfusOnWoW:
People do attach to auras in the same group. That case would work with group scaling, but hide the option in your suggestion.

emptyrivers:
Hmmm, true. Maybe the condition to hide the scale option could be, if there exists a member of the group which is not a descendant of the group aura, rather than just a child? If people are silly with their parents, then this might cause unexpected auras to be scaled along with the group. For example, A is a group, with B as a child. C is not a member of A’s group. B has C as a parent, and C has A as a parent. Then scaling A’s group will also scale B, even though it’s not part of the group.

What a finangly question, honestly.

emptyrivers:
So, yeah, I agree with infus. The edge cases are too numerous and too finicky to bother dealing with.

InfusOnWoW:
I fixed your whitespace, there were a few extra spaces in empty lines. (You can probably configure your editor to automatically remove them on save.) And fixed one small bug that I found in testing, setting the scale to 0, was a) possible and b) lead to two different errors: divide by zero and scale factor needs to be > 0.

Thanks for the contribution.

PR #476, opened 2018-07-09
OnHide code was run on frame creation, fix #474

### LINKED ISSUE #474: on show trigger on options loading or when select aura + lua error
last update 8.0 from git
clean WeakAuras.lua save variable

make new icon
trigger1: custom - status - event(s) "SOMETHING"
custom trigger: function() return false end
Action - on show - custom - print("hello")

close options
/reload
/wa => chat: hello
click on the aura => chat: hello x2

=====

Expected: nothing
What we get: on show is triggered
note: same with on show - print in chat instead of custom code
=====

Now if i make another icon and select it i get this error in loop while options are open:

```
8x WeakAuras\WeakAuras-@project-version@.lua:3724: attempt to index field '?' (a nil value)
WeakAuras\WeakAuras-@project-version@.lua:3724: in function `GetTriggerStateForTrigger'
WeakAuras\BuffTrigger.lua:296: in function `SetAuraVisibility'
WeakAuras\BuffTrigger.lua:671: in function `ScanAuras'
WeakAuras\BuffTrigger.lua:1293: in function <WeakAuras\BuffTrigger.lua:1289>

Locals:
id = "New"
triggernum = 0
(*temporary) = nil
(*temporary) = nil
(*temporary) = "attempt to index field '?' (a nil value)"
triggerState = <table> {
 test = <table> {
 }
 test123 = <table> {
 }
}
```

mrbuds:
The whole frame:SetScript("onHide", ...) block was run when WeakAuras.CreateFrame() is called

An other solution if the frame:Hide() is needed, i don't see why it's there, is to move the block after the frame:Hide() call

### DIFF
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -185,7 +185,6 @@ function WeakAuras.CreateFrame()
     yOffset = (492 - GetScreenHeight()) / 2;
   end
   frame:SetPoint("TOPRIGHT", UIParent, "TOPRIGHT", xOffset, yOffset);
-  frame:Hide();
 
   local width, height;
   if(db.frame) then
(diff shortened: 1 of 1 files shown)


emptyrivers:
Would be preferable if you actually do hide the frame.

mrbuds:
last commit re add frame:Hide() and move the OnHide block after it

emptyrivers:
LGTM

PR #478, opened 2018-07-11
add a static popup to confirm deletion of auras

emptyrivers:
Implements [wowace issue 644](https://www.wowace.com/projects/weakauras-2/issues/644).

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -249,32 +249,16 @@ function WeakAuras.MultipleDisplayTooltipMenu()
       text = L["Delete all"],
       notCheckable = 1,
       func = function()
+        local toDelete = {};
+        local parents = {};
         for index, id in pairs(tempGroup.controlledChildren) do
-          local toDelete = {};
-          local parents = {};
-          for index, id in pairs(tempGroup.controlledChildren) do
-            local childData = WeakAuras.GetData(id);
-            toDelete[index] = childData;
-            if(childData.parent) then
-              parents[childData.parent] = true;
-            end
-          end
-          for index, childData in pairs(toDelete) do
-            WeakAuras.DeleteOption(childData);
-          end
-          for id, _ in pairs(parents) do
-            local parentData = WeakAuras.GetData(id);
-            local parentButton = WeakAuras.GetDisplayButton(id);
-            WeakAuras.UpdateGroupOrders(parentData);
-            if(#parentData.controlledChildren == 0) then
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
Modern UI Design shuns message boxes for Delete confirmation. 

It has been replaced by undo, either limited to one action: For example, Windows's trashcan for deleted files, or how e.g. gmail handles deleted emails. 

Or in other applications by just being part of a global undo & redo.

I actually had a half-finished patch for undoing deletion at one point, but the user interface didn't work out, and it didn't feel that great, that's why we didn't take it on.

So, we can probably merge this (after a review), but leave the ticket open

Stanzilla:
If you touch an ace widget you have to bump the minor version at the top of the file

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -1256,6 +1240,47 @@ function WeakAuras.DeleteOption(data)
...
+    self.data = nil
+  end,
+  showAlert = true,
+  whileDead = true,
+  preferredindex = STATICPOPUP_NUMDIALOGS,
+}
+
+function WeakAuras.ConfirmDelete(toDelete, parents)
+  if toDelete then
+    StaticPopup_Show("WEAKAURAS_CONFIRM_DELETE", "", "", {toDelete = toDelete, parents = parents})
```
InfusOnWoW: The message should tell the user what they are deleting, that this the group name, aura name, or for any other case properly somthing like "deleting x auras". 

The buttons in message boxes should never be "yes"/"no", but always verbs indicating the action that is about to be taken.

In this case: "Keep" or "Cancel" and "Delete".

emptyrivers:
I have improved the warning text.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -1256,6 +1240,49 @@ function WeakAuras.DeleteOption(data)
...
+}
+
+function WeakAuras.ConfirmDelete(toDelete, parents)
+  if toDelete then
+    local warningForm = L["You are about to delete %d aura(s). |cFFFF0000This cannot be undone!|r Would you like to continue?"]
+    StaticPopupDialogs["WEAKAURAS_CONFIRM_DELETE"].text = warningForm:format(#toDelete)
+    StaticPopup_Show("WEAKAURAS_CONFIRM_DELETE", "", "", {toDelete = toDelete, parents = parents})
+  end
+end
+
```
Stanzilla: lgtm minus those newlines here

PR #482, opened 2018-07-12
Equipment Slot Name

### LINKED ISSUE #479: Equipment Slot Cooldown Trigger not passing Name
What steps will reproduce the problem?
Create a new aura with Status - Cooldown Progress (Equipment Slot)  (I used Trinket 1 slot)

What is the expected output? What do you see instead?

Works fine but itemName isnt passed to %n

Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?
Yes
What version of WeakAuras are you using?
2.5.12
Was it working in a previous version? If yes, which was the last good one?
Yeah, Hoboken reported it before

Do you have an error log of what happened? If you don't see any errors, make sure that error reporting is enabled (`/console scriptErrors 1`) or install [BugSack](https://mods.curse.com/addons/wow/BugSack). 

None

Rehok:
Fixes #479 - Send name of item from Cooldown Progress (Equipment Slot) to state.name

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2315,7 +2315,7 @@ WeakAuras.event_prototypes = {
       return duration, startTime + duration;
     end,
     nameFunc = function(trigger)
-      return "";
+      return (GetItemInfo(GetInventoryItemID("player", trigger.itemSlot)));
     end,
     iconFunc = function(trigger)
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
I've slightly changed the code to not have extra (), and put it into 8.0.

PR #486, opened 2018-07-18
Add extendTimer to BigWigs and DBM timers

mrbuds:
This option let you extend duration and expirationTime for bigwigs and dbm timers
I did this because bossmods tend to give timers that end when a boss start casting instead of when they finish their cast

I tested it with single timers and clones

screenshot of gui:
https://i.imgur.com/rOZLEuR.png

ps: before doing this PR i tried to make a TSU to handle this problem, which worked for BW but not for DBM, and i had to go deep into unsupported territory(c) https://wago.io/H1BXDComQ
This solution is much more elegant and work with both bossmod

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ WeakAuras_Main.lua
 WeakAuras_Options.lua
 WeakAuras_Templates.lua
 WeakAuras/Libs/
-.commit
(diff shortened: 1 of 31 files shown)


mrbuds:
There is a bug when using extendTimer and remaining time check, i'm working on a fix

mrbuds:
I don't like how i'm doing this, the trigger should never change data store in "bar"

PR #490, opened 2018-07-18
Add "Adjust Timer" option to DBM and BigWigs timers

mrbuds:
This option let you change the duration & expirationTime for bigwigs and dbm timers

I did this because bossmods tend to give timers that end when a boss start casting instead of when they finish their cast.

Tests done for both bigwigs and dbm 
- with and without remaining time set
- with and without clone per event set
- with positive and negative value
- with negative value > - duration (it skip bar creation)

screenshot of gui:
https://i.imgur.com/XBP7JO4.png

The timer adjustment is done in WeakAuras.CopyBarToState() and  WeakAuras.CopyBigWigsTimerToState()

To avoid that the bar stop prematurely from DBM_TimerStop or BigWigs_StopBar events i made it skip them if extendTimer is set, and use state.autoHide = true instead

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2575,6 +2575,12 @@ WeakAuras.event_prototypes = {
           local triggerSpellId = nil;
         ]];
       end
+
+      local ret2 = [[
+        local extendTimer = %s;
+      ]]
+      ret = ret .. ret2:format(trigger.use_extend and tonumber(trigger.extend or 0) or 0)
+      
       local copyOrSchedule;
       if (trigger.use_remaining) then
         local ret2 = [[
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -2618,7 +2624,7 @@ WeakAuras.event_prototypes = {
             end
 
           elseif (event == "DBM_TimerStop") then
-            if (states[id]) then
+            if (states[id] and states[id].extend == 0) then
```
InfusOnWoW: That's incorrect. the TimerStop event comes in two cases. One if the timer is finished, and secondly if the timer is obsolete, e.g. phase transitions.

mrbuds: Indeed, "manually" hiding the bar on TimerStop event should be done if
- we didn't extend timer `states[id].extend == 0` (autoHide = false)
- remainingTime on dbm's timer is not 0 (obselete timer case)

In practice remainingTime is never exactly 0 when we get the TimerStop event
results

So instead of comparing the bw/dbm timer's remainingTime with 0, i have chosen in the next commit to account for a 0.2s latency.
I don't like using this made up max latency but i don't see a better solution here.

PR #494, opened 2018-07-19
adds group zone ID

Causese:


WeakAuras/Prototypes.lua
```diff
@@ -876,6 +876,16 @@ WeakAuras.load_prototype = {
        end,
       test = "WeakAuras.CheckNumericIds([[%s]], zoneId)"
     },
+    {
+      name = "groupzoneId",
+      display = L["Group Zone ID(s)"],
+      type = "string",
+      init = "arg",
+      desc = function()
+         return L["Supports multiple entries, separated by commas\n"] .. L["Current Zone ID: "] .. (C_Map.GetMapGroupID(C_Map.GetBestMapForUnit("player")) or "none")
```
Stanzilla: *Current Zone Group ID

WeakAuras/Prototypes.lua
```diff
@@ -876,6 +876,16 @@ WeakAuras.load_prototype = {
        end,
       test = "WeakAuras.CheckNumericIds([[%s]], zoneId)"
     },
+    {
```
InfusOnWoW: So which of those is the more useful one? The Zone Group or the Zone Id?

That one should be first.

Causese: zoneId can always return something so it's probably more useful? (an ID that is)

PR #495, opened 2018-07-19
add buff iterators

emptyrivers:
Feel free to close this if you don't believe it's useful. The additions are simple iterators which allow you to dig through the buff tray in a more flexible manner than WA_GetUnitBuff(unit, spellname, filter).

Usage:
```lua
for name, icon, count, duration, expireTime in WA_IterateBuffs('player', 'PLAYER') do
  -- block
end
```

Also supports using an offset to discard the first few values:
```lua
for duration, expireTime, _, _ , _, spellID in WA_IterateDebuffs('target', 'PLAYER', 5) do
  -- do something using the debuffs on target from the player
end
```

### DIFF
--- a/WeakAuras/HelperFunctions.lua
+++ b/WeakAuras/HelperFunctions.lua
@@ -43,10 +43,40 @@ local WA_ClassColorName = function(unit)
   return RAID_CLASS_COLORS[class]:WrapTextInColorCode(UnitName(unit))
 end
 
+local function makeIterator(api, unit, filter, offset)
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
@asakawa-k did convice me, that these don't add too much, since iterating buffs isn't as complex as iterating group members.

Maybe, that will change once we have a better idea what kind of helpers we want.

emptyrivers:
That’s fine.

PR #502, opened 2018-07-20
 Actually fix trigger.itemSlot nil

Rehok:


Stanzilla:
https://github.com/WeakAuras/WeakAuras2/commit/b807532b1eceac6687d8481edc67971190508e0f

Issue #467, opened 2018-06-27
Hybrid Sorting doesn't remember what aura was selected when moving it within a dynamic group


Causese:
**Describe the bug**
Hybrid sorting doesn't remember selected weakaura when its index is changed

**To Reproduce**
1.Put two auras in a dynamic group
2. set sorting to hybrid
3. select one of the auras to be listed first (it's an option within hybrid sorting)
4. swap the index(?) of the weakauras (basically put one of the weakauras elsewhere within the same dynamic group)

**Expected behavior**
I expected the selected aura to be remembered even when it's put elsewhere within the dynamic group

**Screenshots**
1. Selecting Co-Tank to be the aura that's displayed first
https://gyazo.com/04441d8cff743f7f3e5c067d80860a78.png
2. I moved Co-Tank below "01 Laser", which caused "01 Laser" to be selected instead of remembering that Co-Tank was selected.
https://gyazo.com/3ccca32feb2a406aa8cfc72f943698a6.png

**Which version of WeakAuras are you using?**
beta-8

**Additional context**
I basically made a weakaura that shows your co-tanks debuffs and wanted it to always be displayed first for more consistency

emptyrivers:
Fixed in 2.6.3

PR #514, opened 2018-07-23
Use id instead of index in hybrid sorting

emptyrivers:
Github issue #467


Stanzilla:
https://cdn.discordapp.com/attachments/172440238717665280/470944958765989908/Schermata_2018-07-23_alle_15.26.41.png

PR #521, opened 2018-07-23
Add SetJustifyH/V to icon text positioning

asakawa-k:
Adds SetJustifyH/V to put the text more firmly in the specified place on the icon


InfusOnWoW:
I updated the commit message to explain the fix and pushed it.

PR #530, opened 2018-07-24
Add Mythic+ Affix load option

sragia:
Useful load condition for affix specific auras. For example my [explosive orbs aura](https://wago.io/NyIGCeCnf) or all necrotic/sanguine/quaking etc auras.

edit: also at the moment description tooltip show affixes how they will be in [BFA ](https://i.imgur.com/q63MKfU.png). IDs doesnt change and being so close to release it made more  sense to keep it that way so there's no need to update it for actual launch

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -79,6 +79,56 @@ local function get_encounters_list()
   return encounter_list:sub(1, -3) .. L["\n\nSupports multiple entries, separated by commas\n"]
 end
 
+WeakAuras.mythic_plus_affixes = {
+  {
+    level = 2,
+    ids = {
+      9,  -- Tyrannical
+      10, -- Fortified
+    }
+  },
+  {
+    level = 4,
+    ids = {
+      5,  -- Teeming
+      6,  -- Raging
+      7,  -- Bolstering
+      8,  -- Sanguine
+      11, -- Bursting
+    }
+  },
+  {
+    level = 7,
+    ids = {
+      2,  -- Skittish
+      3,  -- Volcanic
+      4,  -- Necrotic
+      12, -- Grievous
+      13, -- Explosive
+      14, -- Quaking
+    }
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -79,6 +79,56 @@ local function get_encounters_list()
...
+    ids = {
+      16, -- Infested
+    }
+  }
+}
+
+local function get_affix_list()
+  local affix_list = ""
+  for _,info in ipairs(WeakAuras.mythic_plus_affixes) do
+    affix_list = string.format("%s\124cffffce00%s %i+\124r\n",affix_list,L["Level"],info.level)
```
InfusOnWoW: spaces after commas. Here and in lots of other places.

WeakAuras/Prototypes.lua
```diff
@@ -929,6 +991,14 @@ WeakAuras.load_prototype = {
       values = "role_types",
       init = "arg"
     },
+    {
+      name = "affixes",
+      display = L["Mythic+ Affix ID(s)"],
+      type = "string",
```
InfusOnWoW: This should be multiselect, see the talent option.

InfusOnWoW:
I've quickly rewritten that load condition to use a multi-selection instead of a simple line edit. See here:
https://github.com/InfusOnWoW/WeakAuras2/commit/08610576fca653db1dce457847e0c82d3226b278

There are a few issues that still need clean up, but that looks and feel much better.

InfusOnWoW:
And with a bit more clean up: https://github.com/InfusOnWoW/WeakAuras2/commit/1d077f4788e1d778f71f886bb70a56be65406fcb

Stanzilla:
Commited as https://github.com/WeakAuras/WeakAuras2/commit/e1414b2845e4348ccdbd67e03469235a96322076 thanks for your contribution :)

PR #535, opened 2018-07-25
adds trinkets, class specific traits and enchants to templates

Causese:
made some notes for tank/heal specific trinkets in case you want/want me to change stuff


InfusOnWoW:
I fixed your one extra whitespace and pushed it directly.

InfusOnWoW:
Fixed your whitespace and pushed it directly.

PR #541, opened 2018-07-26
fix hiding adjusted timers on obselete timers

mrbuds:
follow up #490

I made the mistake to not use topic branch before i did #490 (as recommended in CONTRIBUTING.md) so it created a new PR.

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2674,7 +2674,7 @@ WeakAuras.event_prototypes = {
       else -- no clones
         ret = ret .. [[
           if (event == "DBM_TimerStart") then
-            if extendTimer ~= 0 then
+            if (extendTimer ~= 0) then
               if (WeakAuras.DBMTimerMatches(id, triggerId, triggerMessage, triggerOperator, triggerSpellId)) then
                 local bar = WeakAuras.GetDBMTimerById(id);
                 WeakAuras.ScheduleDbmCheck(bar.expirationTime + extendTimer);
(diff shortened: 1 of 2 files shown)


mrbuds:
The no-clone code needs a fix
It use WeakAuras.GetDBMTimer() to get the matching bar with the shortest remainingTime, but WeakAuras.DBMTimerMatches won't return a match for a timer which was extended and the bar doesn't exists anymore.

mrbuds:
On this last commit what changed for no clone:
- on new bar event, schedule a check for updating with the modified timer
- if a bar is already show, check that the bar with the shortest remainingTime given by WeakAuras.GetDBMTimer() will finish before what is in states[""] before updating
- WeakAuras.GetDBMTimer() check that the bar with adjusted timer will not finish in the past (for negative value)

WeakAuras/Prototypes.lua
```diff
@@ -2673,12 +2673,23 @@ WeakAuras.event_prototypes = {
...
+                local bar = WeakAuras.GetDBMTimerById(id);
+                WeakAuras.ScheduleDbmCheck(bar.expirationTime + extendTimer);
+              end
+            end
+          end
+          local bar = WeakAuras.GetDBMTimer(triggerId, triggerMessage, triggerOperator, triggerSpellId, extendTimer);
           local id = "";
           if (bar) then
+            if (not (states[""] and states[""].show)
+              or (states[""] and states[""].show and states[""].expirationTime > (bar.expirationTime + extendTimer))) then
```
InfusOnWoW: Please only add this check if extendTimer is used.

WeakAuras/Prototypes.lua
```diff
@@ -2891,12 +2902,23 @@ WeakAuras.event_prototypes = {
...
+              if (WeakAuras.BigWigsTimerMatches(id, triggerAddon, triggerSpellId, triggerTextOperator, triggerText)) then
+                local bar = WeakAuras.GetBigWigsTimerById(id);
+                WeakAuras.ScheduleBigWigsCheck(bar.expirationTime + extendTimer);
+              end
+            end
+          end
+          local bar = WeakAuras.GetBigWigsTimer(triggerAddon, triggerSpellId, triggerTextOperator, triggerText, extendTimer);
           local id = "";
           if (bar) then
+            if (not (states[""] and states[""].show)
```
InfusOnWoW: Same here.

Issue #413, opened 2018-05-12
Custom trigger "autoHide" option


asakawa-k:
_What steps will reproduce the problem?_

New Aura : 
custom - event - `PLAYER_STARTED_MOVING` 
trigger: `function() return true end`
Hide = custom
duration info: `function() return 5, GetTime()+5 end`

_What is the expected output? What do you see instead?_

Expected is that the duration runs when the event fires. Then the trigger untriggers at the completion of the custom duration. (autoHide)
What happens is the timer ends and the Aura continues to show. 

This was discussed in #wa-alpha-testing. 
Ideally, in my view, we would have the option to set a custom trigger to be autoHide or not. 
Infus suggested that this option might fit into the functionalities of status and event type triggers, with event type triggers always working with autoHide and status ones never autoHide.

InfusOnWoW:
Is now implemented.

PR #551, opened 2018-07-28
autoHide for custom triggers

asakawa-k:
Sets the state.autoHide to true for "event" custom triggers and false for "status" custom triggers, falling back to data.automaticAutoHide on other trigger types.

Should address https://github.com/WeakAuras/WeakAuras2/issues/413

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -460,6 +460,20 @@ function WeakAuras.ActivateEvent(id, triggernum, data, state)
         state.expirationTime = arg2;
         changed = true;
       end
+      local autoHide
+      if data.trigger.custom_type then
+        if data.trigger.custom_type == "event" then
+          autoHide = true
+        elseif data.trigger.custom_type == "status" then
+          autoHide = false
+        end
+      else
+        autoHide = data.automaticAutoHide
+      end
+      if (state.autoHide ~= autoHide) then
+          state.autoHide = autoHide
+          state.changed = true
+      end
       if (state.value or state.total) then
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
So let's figure this out, how does autoHide even work in WeakAuras...

For BuffTriggers:
Those don't set autoHide.
Except for multi, that one sets autoHide.

For GenericTriggers, the prototype data is copied into some internal data automaticAutoHide

By default, automaticAutoHide is true, with trigger opting out via automaticAutoHide in the prototypes.[1]
This also applies to custom trigger. Those do have automaticAutoHide set to true.

automaticAutoHide is then used in WeakAuras.ActivateEvent to decide on whether to set state.autoHide to true. BUT, only if internal data duration is used.

internal data.duration is set to the duration if either:
- It's a custom timed event trigger [2]
- Or the it's a timed event/status trigger. [3]


[1] It seems all triggers that set automaticAutoHide to false in the Prototypes due that unncessarily.
[2] This is done twice...
[3] There should be no timed status trigger.

I'll continue looking at this in the coming days.

InfusOnWoW:
I've implemented what I think is a good solution: https://github.com/WeakAuras/WeakAuras2/pull/617

That adds a checkbox on selection event/timed to select between a static hide duration or a dynamic hide duration. If a dynamic duration is selected, a custom duration function can be entered and is used for the autoHide information.

Note, that these triggers have no custom unhide function. (Though there's technically notthing that would prevent that)

emptyrivers:
#617 is merged now, so this is probably obsolete.

PR #555, opened 2018-07-28
Refactor

emptyrivers:
More backend changes.

### DIFF
--- a/WeakAuras/BuffTrigger.lua
+++ b/WeakAuras/BuffTrigger.lua
@@ -292,15 +292,15 @@ do
 end
 WeakAuras.aura_cache = aura_cache;
 
-function WeakAuras.SetAuraVisibility(id, triggernum, cloneId, showOn, unitExists, active, unit, duration, expirationTime, name, icon, count, index, spellId, unitCaster)
+function WeakAuras.SetAuraVisibility(id, triggernum, cloneId, buffShowOn, unitExists, active, unit, duration, expirationTime, name, icon, count, index, spellId, unitCaster)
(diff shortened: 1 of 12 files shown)


WeakAurasOptions/Cache.lua
```diff
@@ -126,20 +126,13 @@ local function Lev(str1, str2)
...
-  for key, value in pairs(cache) do
-    if(nearkey:lower() == key:lower()) then
-      return key;
-    end
-  end
   local bestKey = "";
   local bestDistance = math.huge;
   local partialMatches = {};
   for key, value in pairs(cache) do
+    if(nearkey:lower() == key:lower()) then
```
InfusOnWoW: That's not the same algorithm anymore.

emptyrivers: What specifically breaks, though?

InfusOnWoW: The old algorithm prefer matches with the same case to matches without the same. E..g if there's a all upper case match, and that one is before the correct match, then this will return the key uppercased.

emptyrivers: Hmm, I see.

WeakAurasOptions/Cache.lua
```diff
@@ -129,15 +129,22 @@ function spellCache.BestKeyMatch(nearkey)
   local bestKey = "";
   local bestDistance = math.huge;
   local partialMatches = {};
-  for key, value in pairs(cache) do
+  local lowerMatch
+  for key in pairs(cache) do
+    if key == newkey then
+      return key
+    end
```
InfusOnWoW: That iteration is not needed, after all. This can directly check cache[nearkey].

PR #558, opened 2018-07-29
range check condition for spell and item cooldown

mrbuds:
Internal events doesn't work (yet) so i used PLAYER_TARGET_CHANGED for now


InfusOnWoW:
I implemented the missing parts and this was directly merged.

Issue #556, opened 2018-07-29
Buffed condition for aura triggers


sragia:
**Is your feature request related to a problem? Please describe.**
Basically at the moment if you create Aura trigger that is always active (Show On - Always)  there's no real condition that you can use to determine if unit is actually buffed or not. 
**Describe the solution you'd like**
Condition Buffed that is true/false, similar how we have On Cooldown for CD Progress triggers.
**Describe alternatives you've considered**.
Can try using remaining time but when not buffed it seems to be math.huge or something, basically can't use remaining time <= 0. so have to do something like remaining time > buff max duration that you type in yourself.

asakawa-k:
I agree this would be a nice addition ("show on" as added to aura triggers after Conditions went in) but a workaround for now is to use `total duration = 0`

sragia:
good point, hadn't tried that out. However that doesnt work for auras that don't have duration, for example Consecration

asakawa-k:
Yeah. In those cases I guess the workaround is imply using 2 triggers, one for buffed and one for missing.

InfusOnWoW:
Implemented by MrBuds

PR #562, opened 2018-07-30
buffed/unbuffed condition on auras with show=always #556

mrbuds:
Add a condition when onShow=always in trigger tab #556 

There is still a problem with the condition list not updated when the trigger option onShow is changed.
If i select another condition, the list is updated.


InfusOnWoW:
Thanks, I'll fix a few omissions in it and merged it. (The check for when to add the property was wrong, and there are buffs without a duration, thus I made a separate active variable in state.)

PR #571, opened 2018-08-01
Fix text shaking animation bug

Lothrik:
Ticket Nr: 1160

https://www.wowace.com/projects/weakauras-2/issues/1160


InfusOnWoW:
I removed the comment from the code, as checking if we need to actually do something is the right thing to do, and thus the comment might confuse more than it helps to explain. Thanks for the patch.

PR #573, opened 2018-08-01
littlepatches

emptyrivers:


WeakAuras/Transmission.lua
```diff
@@ -556,7 +561,7 @@ local function importPendingData()
   if not WeakAuras.IsOptionsOpen() then
     WeakAuras.ShowOptions()
   end
-  WeakAuras.PickDisplay(installedData[0].id)
+  WeakAuras.PickAndEditDisplay(installedData[0].id)
```
InfusOnWoW: This also opens the rename, that seems wrong.

WeakAuras/Transmission.lua
```diff
@@ -1282,18 +1283,19 @@ function WeakAuras.MatchData(data, children)
...
-    info.activeCategories[checkButtons["oldchildren"]] = true
+    info.activeCategories[checkButtons.oldchildren] = true
   end
 
   if info.modified ~= 0 then
     for i, diff in pairs(info.diffs) do
       for key in pairs(diff) do
         local button = keyToButton[key]
         if button then
+          button = button == checkButtons.anchor and checkButtons.arrangement or button
```
InfusOnWoW: What?

WeakAuras/Transmission.lua
```diff
@@ -372,6 +372,7 @@ local function install(data, oldData, patch, mode)
     if patch then -- if there's no result from Diff, then there's nothing to do
       for key in pairs(patch) do
         local checkButton = keyToButton[key]
+        checkButton = (not isParent and checkButton == checkButtons.anchor) and checkButtons.arrangement or checkButton
```
InfusOnWoW: I kinda prefer:
if (not isParent and checkButton == checkButtons.anchor) then
  checkButton = checkButtons.arrangement;
end

Even if longer, that makes imho the intent of changing the checked button in one specific case cleare.

PR #578, opened 2018-08-02
Add group unit options

asakawa-k:
The goal is to provide options in the group=unit settings for users to specify raiding roles to watch and also exclude themselves from results.
 
The use case I had in mind for this was be to track "tank swap" debuffs on other tanks. These settings would let users only see debuffs on other tanks which is a very common thing to want to see without a very satisfactory method currently. 
Currently while raiding as tank I would *always* focus my cotank and use the focus unit for tank monitoring Auras. This gets messy when there are 3 tanks and removes any other uses I might have for my focus. Other people have made complex TSU auras to track the other tank. 

Of course I expect that the setting could be useful for other stuff too.

### DIFF
--- a/WeakAuras/BuffTrigger.lua
+++ b/WeakAuras/BuffTrigger.lua
@@ -652,7 +652,7 @@ function WeakAuras.ScanAuras(unit)
                 end
               end
             end
-            
+
             local satisfies_role = true
             if data.group_role then
               satisfies_role = data.group_role == "ANY" or UnitGroupRolesAssigned(unit) == data.group_role
@@ -696,6 +696,9 @@ function WeakAuras.ScanAuras(unit)
               if(data.group_count) then
                 -- Query count from aura cache
                 local aura_count, max = aura_object:GetNumber(id, triggernum, data), aura_object:GetMaxNumber();
+                if (data.ignoreSelf) then
+                  max = max - 1;
+                end
                 local satisfies_count = data.group_count(aura_count, max);
 
                 if(data.hideAlone and not IsInGroup()) then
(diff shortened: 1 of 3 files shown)


asakawa-k:
This is a little under-cooked still actually

asakawa-k:
Okay, done much more testing on this now and fixed the issue.

WeakAuras/BuffTrigger.lua
```diff
@@ -688,7 +697,7 @@ function WeakAuras.ScanAuras(unit)
                 -- Query count from aura cache
                 local aura_count, max = aura_object:GetNumber(id, triggernum, data), aura_object:GetMaxNumber();
                 local satisfies_count = data.group_count(aura_count, max);
-
+
```
Stanzilla: 🤔

WeakAuras/Types.lua
```diff
@@ -1302,6 +1302,13 @@ WeakAuras.group_aura_stack_info_types = {
   stack = L["Aura Stack"]
 }
 
+WeakAuras.group_roles = {
```
Stanzilla: we already have role_types, can we re-use and maybe add "any" to it?

WeakAurasOptions/BuffTrigger.lua
```diff
@@ -756,6 +756,20 @@ function WeakAuras.GetBuffTriggerOptions(data, trigger)
       end,
       values = group_aura_stack_info_types
     },
+    ignoreSelf = { 
+      type = "toggle", 
+      name = "Ignore player",
```
Stanzilla: name = L["Ignore self"],

WeakAurasOptions/BuffTrigger.lua
```diff
@@ -756,6 +756,20 @@ function WeakAuras.GetBuffTriggerOptions(data, trigger)
...
     },
+    ignoreSelf = { 
+      type = "toggle", 
+      name = "Ignore player", 
+      order = 47.7, 
+      hidden = function() return not (trigger.type == "aura" and trigger.unit == "group"); end, 
+    }, 
+    group_role = { 
+      type = "select", 
+      name = "Group Member Role",
```
Stanzilla: L["Group Member Role"],

WeakAuras/Types.lua
asakawa-k: I thought about that but we don't actually want it to show "any" in the load conditions.

PR #587, opened 2018-08-03
add conditions to templates

mrbuds:
Add support for condition in templates

Added some template types and changed how some are show

* abilitybuff & abilitydebuff : 
aura trigger+cooldown spell trigger
show buff when active, otherwise show cooldown, and desaturate on cooldown

* abilitytarget : 
cooldown spell trigger
show always, turn red if out of range and have target, desaturate on cooldown or insufficientResources

* ability : 
show always, desaturate on cooldown or insufficientResources

* item :  
show always, desaturate on cooldown 

* debuff : 
show always, desaturate if not active

I only used the new templates types on some spells for disc priest for now

ps: There is a bug with "aura - debuff - show always" template, the visibility is changed to "on active" by WeakAuras.Modernize().
Infus have a patch close to be ready for this.


mrbuds:
* abilitycharge
cooldown spell trigger
show always, desaturate if charge=0

mrbuds:
It's now possible to use more than one type of template per ability/item/buffs/..
The "type" field was changed to "types" and is now a list.
GUI was updated to show a new panel when there is more than one template available.

Templates titles and descriptions needs to be improved.
The GUI code can certainly be improved

mrbuds:
I made table to have a better overview on what the templates types are doing, this will help avoid inconsistency between them.

https://docs.google.com/spreadsheets/d/1XYmpwUdF_TzPyq5tJp8YSrNUuCkp1xty_f8zs6kpvcw/edit#gid=0

Things that needs improvement:

- conditions with range (red) + resource (blue) will conflict when both are true

- conditions with aura + cooldown doesn't have good visual when aura is active, i avoided to use glow, maybe it is still the best solution ?

- when first testing a basic UI made only from templates, the way icons where changing color too much was overwhelming, so i removed the use of desaturate on cooldown.
The result is better, but i'm not sure what to do of spells with charges=0, to differentiate "spell cooldown" of "spell not usable".
Making them blue when charges=0 seems good to me, but then what to do with spells with charges that use resource, to differentiate no resource from no charge.

Stanzilla:
I like it sans the active thing, maybe should leave that out for now?

In my own UI I set charges=0 the same as on cooldown

Stanzilla:
I don't want to have the final say about the active thing, wait for what the others have to say, please. 

For on cooldown and charges=0 I use #7f7f7f (127,127,127)

mrbuds:
Last commit added possibility to use a table in the "talent" field for multi-talents
This open possibility to have 2 (or more) sets of templates for a same spell, which is useful if a talent add charges to a spell for example.

mrbuds:
last commit fix the template button in the trigger tab
add trigger template will add conditions using correct trigger position

New templates for all classes is not finished yet, it takes much more time to do it with zero knowledge of a spec.

I think it's ready for review now

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -23,6 +23,560 @@ AceGUI:RegisterLayout("WATemplateTriggerLayoutFlyout", function(content, childre
   flowLayout(content, children);
 end);
 
+local changes =
+{
```
Stanzilla: move the { to the same line as = please

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -23,6 +23,560 @@ AceGUI:RegisterLayout("WATemplateTriggerLayoutFlyout", function(content, childre
...
+    property = "inverse",
+  },
+  glow = {
+    value = true,
+    property = "glow",
+  }
+}
+
+local checks =
+{
```
Stanzilla: same as above

PR #593, opened 2018-08-04
Extend event coverage for cooldown checks

jgersti:
Register `ACTIONBAT_UPDATE_COOLDOWN` for `cdReadyFrame`.

This change allows for example to accurately track stacks of _Spinning Crane Kick_. Previously a decrease of the stack count would not have been registered in the aura.


InfusOnWoW:
That's a bit strange. Seems the event fires even if SCK is not on any action bar, and it is indeed the only event that fires for that.

PR #601, opened 2018-08-06
Fixing the Luminous Barrier CD template

gnumast:
The Luminous Barrier cooldown template sets Evangelism as a required talent in order to load the aura which will never happen since both talent share a row.


InfusOnWoW:
Thanks.

PR #603, opened 2018-08-06
misc

emptyrivers:


WeakAuras/Types.lua
```diff
@@ -1589,8 +1597,6 @@ WeakAuras.internal_fields = {
   parent = true,
   internalVersion = true,
   regionType = true,
```
InfusOnWoW: regionType is probably also problemeatic.

emptyrivers: Ah, you're probably right on that one. I think Display would be the best category for regionType, given just how much of the display tab data is governed by the regionType.

InfusOnWoW: I agree

emptyrivers: on further thought, sortHybridTable (and much of the dynamic group options such as growth direction) should be categorized as arrangement

WeakAuras/Types.lua
```diff
@@ -1538,7 +1538,7 @@ WeakAuras.update_categories = {
   },
   {
     name = "display",
-    fields = {},
+    fields = {"regionType"},
```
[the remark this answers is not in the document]
emptyrivers: True, I don't need to, but if it's alright with you I think I'll leave it.

PR #606, opened 2018-08-06
Integration of group role options

asakawa-k:
Gather role count info on GROUP_ROSTER_UPDATE
Handle role counts as potential max counts when those settings are used.


InfusOnWoW:
Thanks, pushed that directly.

PR #608, opened 2018-08-07
Fix false negative on sortHybridTable change

### LINKED ISSUE #607: Changing only the sortHybridTable, in a dynamic group results in a false negative from MatchData.
This will require a special case algorithm, akin to the special case algorithm that already exists for updating sortHybridTable.

emptyrivers:
As it turns out, I had more time this evening than I thought. Fixes #607.


emptyrivers:
Second commit fixes an issue reported by Nightwarden on discord.

PR #614, opened 2018-08-08
Rewrite SetFrameLevel code to be more robust for clones

InfusOnWoW:
The group/dynamic group now simply record the desired frame levels
in their modify function. And get the desired frame level in the
expand function.

Github-Issue: 613


jgersti:
When closing the options or clicking on the aura/group in the options, the group and all its children get frameLevel of 0 (or 1 for text) assigned. Additionally the `WeakAuras.frameLevels` table is empty.


_On a side note:_  
_This is quite interesting since its mean that for many cases the frame level is actually irrelevant._

InfusOnWoW:
I can reproduce that and fixed it.

PR #619, opened 2018-08-10
Templates: Fixes specified unit for Lingering Sporepods trinket

Causese:


InfusOnWoW:
Thanks, fixed the commit message and pushed it.

PR #624, opened 2018-08-11
add securecall to blocked functions, remove ForceQuit

lqnrd:
`securecall` can be used to break out of the environment. `ForceQuit` has been protected for some time now and can only be called from Blizzard code.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -402,7 +402,7 @@ local blockedFunctions = {
   SetBindingMacro = true,
   GuildDisband = true,
   GuildUninvite = true,
-  ForceQuit = true,
+  securecall = true,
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
How can secucecall be abused?

lqnrd:
`myOwnGlobalTable = securecall("loadstring", "return _G")`

emptyrivers:
Frankly I’m just impressed that securecall can even be called from a tainted execution path.

lqnrd:
It just stays tainted afterwards :). The problem is that the first argument does not need to be a function, but as shown can be a string.

InfusOnWoW:
Ah that's news to me. Looks good then.

Stanzilla:
Thanks!

Issue #660, opened 2018-08-20
Supercharged Templates Design Dicussions/TODO List


InfusOnWoW:
~~* The last page needs icons.~~

Stanzilla:
I want to work on the descriptions some more

Stanzilla:
item.buff and item.requiresTarget and item.debuff and item.requiresTarget have insufficientResources, that should be isUsableBlue, right?

What do totem types do with insufficientResources?
Do we make color items blue when they ARE usable and spells when they ARE NOT usable or is that just weird naming?

I made some wording changes, can you check if that all matches what it actually does? @mrbuds

InfusOnWoW:
The templates don't work very well with non-icon types.

mrbuds:
Except "Basic Show On Cooldown", I made everything use either insufficientResources or isUsableBlue (if "usable = true"), including totems, because most shaman totems cost mana.

It is a bit overkill and the condition will not be use on some ability with no cost indeed

Yes it's a weird name, isUsableBlue, make the icon blue when NOT usable.

Stanzilla:
okay we need to rename that then :D

mrbuds:
I made a PR to fix non-icon types and rename isUsableBlue.
Instead of the glow effect i used the color yellow for non-icon.

I'm using a color code to show the state of a spell
- blue = not usable
- grey = on cooldown
- red = out of range
- yellow / glow = active

Now, the 4 types of progress bar we give have different default color :
- red
- red
- yellow
- green

I think we should give only 1 default color possibility, which would not conflict with the color code.
White is the easiest solution, what do you think?

mrbuds:
btw i think we should only give 2 possibilities for progress bar, horizontal & vertical

mrbuds:
![cooldowntext](https://i.imgur.com/bq5aR84.png)
"Show Only On Cooldown" is bad naming

mrbuds:
"Default" icon have cooldown and inverse = false instead of true for all other icons

InfusOnWoW:
* Changing the first screen to be more like this:
https://media.discordapp.net/attachments/273144328447197186/481179051445452820/wa.jpg

mrbuds:
I made isBuffedGlow() use color yellow for regionTypes which don't support glow but support color.
As a result, the description of templates with glow is wrong for these regionTypes.

mrbuds:
Added icons in #664 
[...]

PR #661, opened 2018-08-20
Fix templates for all regionTypes

mrbuds:
=> use correct field for color
 => for aura and totem, use yellow color instead of glow for all regionTypes except "icon"
 => model : no conditions
Rename isUsableBlue => isNotUsableBlue

#660

### DIFF
--- a/WeakAurasTemplates/TriggerTemplates.lua
+++ b/WeakAurasTemplates/TriggerTemplates.lua
@@ -23,54 +23,173 @@ AceGUI:RegisterLayout("WATemplateTriggerLayoutFlyout", function(content, childre
   flowLayout(content, children);
 end);
 
+local grey = {
+  0.5,
+  0.5,
+  0.5,
+  1,
+};
+
+local blue = {
+  0.5,
+  0.5,
+  1,
+  1,
+};
+
+local red = {
+  0.8,
+  0.1,
+  0.1,
+  1,
+};
+
+local white = {
+  1,
+  1,
+  1,
+  1,
+};
+
+local yellow = {
+  1,
+  1,
+  0,
+  1,
+};
+
 local changes = {
   grey = {
-    value = {
-        0.5,
-        0.5,
-        0.5,
-        1,
+    icon = {
+      value = grey,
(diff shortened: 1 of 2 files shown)


Stanzilla:
So we have

```
onCdCheckGrey (isOnCdGrey better?)
spellInRangeRed (isSpellNotInRangeRed better?)
isBuffedGlow 
insufficientResourcesBlue
totemActiveGlow
isNotUsableBlue
buffGreyed (missingBuffGreyed better? or is it "hasbuff"?)
```

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -255,7 +402,7 @@ local function createAbilityAndDebuffTrigger(triggers, item)
   createAbilityTrigger(triggers, 1, item, "showAlways");
 end
 
-local function subTypesFor(item)
+local function subTypesFor(item, regionType)
```
InfusOnWoW: This function should filter which options are available, so that for models only basic ones are offered. Offering some choices which don't actually work is not great.

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -133,44 +252,72 @@ local function buildCondition(trigger, check, properties)
   return result;
 end
 
-local function buffGreyed(conditions, trigger)
-  tinsert(conditions, buildCondition(trigger, checks.buffedFalse, {changes.grey}));
+local function missingBuffGreyed(conditions, trigger, regionType)
+  if regionType ~= "model" then
```
InfusOnWoW: There is the possibility of adding region types via plugins, and in fact there's a plugin: StopMotion

So any check for regionTypes needs to test the supported ones. Testing for not model does not work.

See the later comment on where to move the check though.

mrbuds: done in #664 

To check if the property exists for the regionType i did this:
WeakAuras.regionTypes[regionType].default[property] 

I'm not 100% sure this is the right way to test it.

Issue #662, opened 2018-08-21
If you check Auto-clone but don't uncheck it "Show On: ___" stays hidden


amiyuy:
**Describe the bug**
If you enable Full Scan and then "Show all matches (Auto-clone)", "Show On: Buffed/Debuffed" is hidden.  If you then only uncheck Full Scan but not Auto-clone too, the Show On dropdown stays hidden.

**To Reproduce**
1. On an Aura trigger check Full Scan
2. Check "Show all matches (Auto-clone)"
3. Uncheck Full Scan
4. Show On dropdown is missing

**Screenshots**
![wowscrnshot_082118_100925](https://user-images.githubusercontent.com/3958649/44417564-c52bed80-a52a-11e8-9916-7c46e0d86026.jpg)

**Which version of WeakAuras are you using?**
2.7.2

InfusOnWoW:
Should be fixed now.

PR #663, opened 2018-08-21
Conditions: Add AND/OR combinations to conditions

InfusOnWoW:

PR #667, opened 2018-08-24
Ensure that Options Frame exists before commiting to an import job

emptyrivers:
If a user clicks on a link in chat, then the import tooltip will appear and allow them to import an aura. This is fine, but importPendingData calls WeakAuras.CloseImportExport, which assumes that the options frame has been created. So, Load WeakAurasOptions and Show the options Frame, and if that fails then back out of the import job.

All of this would be solved if we weren't abusing the tooltip like we are.
Issue reported on Discord.

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -381,12 +381,16 @@ local function importPendingData()
   local indexMap = pendingData.indexMap
 
   -- cleanup the mess
-  WeakAuras.CloseImportExport()
   HideUIPanel(ItemRefTooltip) -- this also wipes pendingData as a side effect
   buttonAnchor:Hide()
   thumbnailAnchor.currentThumbnail:Hide()
   thumbnailAnchor.currentThumbnail = nil
-  if not imports then return end
+  if imports and WeakAuras.LoadOptions() then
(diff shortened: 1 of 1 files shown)


emptyrivers:
Second commit fixes #662.

InfusOnWoW:
That second commit is incomplete, compare to https://github.com/WeakAuras/WeakAuras2/pull/663/commits/249308f115e336eb680567e780f96386f9404bee

WeakAuras/Transmission.lua
```diff
@@ -381,12 +381,16 @@ local function importPendingData()
...
 
   -- cleanup the mess
-  WeakAuras.CloseImportExport()
   HideUIPanel(ItemRefTooltip) -- this also wipes pendingData as a side effect
   buttonAnchor:Hide()
   thumbnailAnchor.currentThumbnail:Hide()
   thumbnailAnchor.currentThumbnail = nil
-  if not imports then return end
+  if imports and WeakAuras.LoadOptions() then
+    WeakAuras.ShowOptions()
```
InfusOnWoW: Don't we want to ensure that the options are always shown after importing?

Issue #678, opened 2018-08-27
[Suggesstion] Streamlining creation of multiple template-based auras


eyriewow:
This is somewhat of an extension to https://github.com/WeakAuras/WeakAuras2/issues/647

**Is your feature request related to a problem? Please describe.**
Creating several auras at a time that are meant to be in a (dynamic) group is a very repetitive and for me at times frustrating process. 

- Create group
- Create aura from template
- Find created aura in your (often very long list of auras)
- Put aura in group
- Repeat 2-4 depending on how many spells you want to monitor

Another issue with the current template based creation process arises if users intend on monitoring a lot of traits or cooldowns. It's very easy to lose track of the auras that have already been created (since users can't see the existing auras while in the template screen). 

**Describe the solution you'd like**
In the screen where users select what type of aura they would like to create there should be an option to create a group or a dynamic group of icons or progress bars. Here's a mockup:
![](https://i.imgur.com/OFDthiJ.png)

The user should now be asked to put in the name of the group they would like to create.
After naming the group the user should face a similar screen to the regular aura creation screen which should allow the user to select multiple templates before the addon creates the icons.
Mockup of this screen:
![](https://i.imgur.com/8v965EC.png)
(At the bottom next to "back" and "cancel" should now be a button to "create group" - I forgot to add it)

**Describe alternatives you've considered**
Can't think of any alternatives to this. Something like a web tool that would create import strings for users as an alternative to the template system would be possible but probably even more effort.

mrbuds:
I like your suggestion, i have started something which would work with any regionType
- Hold ctrl for multi selection, a new "batch" button show
- Click "batch" when done
- Select template for each spell (to do)
- Add all templates, without group, or in selected group (to do)

![mockup](https://i.imgur.com/Wfq1FQN.png)

InfusOnWoW:
This is part of the next release.

Issue #647, opened 2018-08-14
When creating auras, add aura to a group automatically


amiyuy:
**Is your feature request related to a problem? Please describe.**
When creating an aura, it would be very useful to have that aura go into the group I have selected automatically.

**Describe the solution you'd like**
If I have a group selected and expanded, add any new auras that I create (from New or From Template) to that group.

**Alternative**
Allow sorting by date created/updated/imported so they can be near each other for easier adding to group.

**Additional context**
When creating large groups of auras (like through Templates for Azerite Trait tracking), it would save a lot of time to not have to find them where they fall alphabetically and intstead have them go directly into the group I have for them.

InfusOnWoW:
The problem is that on selecting new, the group is technically no longer selected.

My preferred solution would be to make new part of a "toolbar", next to the search button. (Same with import, and thus it would be capable of inserting into whatever is selected.)

As to your context, @mrbuds was interested into investigating adding "creating groups" directly from templates. I think being able to create into a group is a generally useful idea.

emptyrivers:
Something along the lines of this has been implemented in 2.7.8 and later.

PR #680, opened 2018-08-27
Templates Batch creation + make new auras in a group

mrbuds:
Allow multi-selection with CTRL and create a batch of auras at once

screen1
A "Batch" button appear when you select labels with CTRL.

"subType" selection screen
screen2

Click "Create Auras" will make all of them at once
screen3

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -488,7 +488,11 @@ local methods = {
             end
           end
         else
-          WeakAuras.PickDisplay(data.id);
+          if (WeakAuras.IsDisplayPicked(data.id)) then
+            WeakAuras.ClearPicks(data.id);
+          else
+            WeakAuras.PickDisplay(data.id);
+          end
           self:ReloadTooltip();
         end
       end
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -849,6 +849,11 @@ function WeakAuras.CreateFrame()
     else
       WeakAuras.PauseAllDynamicGroups();
       finishPicking();
+      if (data.controlledChildren and #data.controlledChildren == 0) then
+        WeakAurasOptions.pickedDisplay = data.id;
+        WeakAurasOptions:PickOption("New");
+        WeakAurasOptions.pickedDisplay = data.id;
+      end
     end
   end
 
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -671,6 +671,12 @@ function WeakAuras.CreateFrame()
 
   frame.PickOption = function(self, option)
     local previousPick = self.pickedDisplay;
+    if (type(self.pickedDisplay) == "table") then
+      previousPick = {}
+      WeakAuras.DeepCopy(tempGroup.controlledChildren, previousPick);
+    elseif (type(self.pickedDisplay) == "string") then
+      previousPick = self.pickedDisplay;
+    end
     self:ClearPicks();
     self.moversizer:Hide();
     self.pickedOption = option;
@@ -724,10 +730,37 @@ function WeakAuras.CreateFrame()
           local new_id = WeakAuras.FindUnusedId("New")
           local data = {id = new_id, regionType = regionType}
           WeakAuras.DeepCopy(WeakAuras.data_stub, data)
+          local parentData;
+          if (not data.controlledChildren and previousPick) then
+            if (type(previousPick) == "table") then
+              previousPick = previousPick[1];
+            end
+            local previousData = WeakAuras.GetData(previousPick);
+            if (previousData) then
+              if (previousData.controlledChildren) then
+                parentData = previousData;
+              else
+                parentData = WeakAuras.GetData(previousData.parent);
+              end
+              if (parentData) then
+                data.parent = parentData.id;
+                tinsert(parentData.controlledChildren, new_id);
+                WeakAuras.Add(parentData);
+              end
+            end
+          end
           data.internalVersion = WeakAuras.InternalVersion();
           WeakAuras.validate(data, WeakAuras.regionTypes[regionType].default)
           WeakAuras.Add(data);
           WeakAuras.NewDisplayButton(data);
+          if (parentData) then
+            local parentButton = WeakAuras.GetDisplayButton(parentData.id);
+            parentButton.callbacks.UpdateExpandButton();
+            WeakAuras.UpdateDisplayButton(parentData);
+            WeakAuras.ReloadGroupRegionOptions(parentData);
+            WeakAuras.SortDisplayButtons();
+            parentButton:Expand();
+          end
           WeakAuras.PickAndEditDisplay(new_id);
         end);
         containerScroll:AddChild(button);
@@ -793,6 +826,7 @@ function WeakAuras.CreateFrame()
     else
       error("An options button other than New or Addons was selected... but there are no other options buttons!");
     end
+    --self:ClearPicks();
   end
 
   frame.PickDisplay = function(self, id)
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -670,6 +670,7 @@ function WeakAuras.CreateFrame()
   end
 
   frame.PickOption = function(self, option)
+    local previousPick = self.pickedDisplay;
     self:ClearPicks();
     self.moversizer:Hide();
     self.pickedOption = option;
@@ -694,7 +695,7 @@ function WeakAuras.CreateFrame()
         button:SetDescription(L["Offer a guided way to create auras for your class"])
         button:SetIcon("Interface\\Icons\\INV_Misc_Book_06");
         button:SetClick(function()
-          WeakAuras.OpenTriggerTemplate();
+          WeakAuras.OpenTriggerTemplate(nil, previousPick);
         end);
         containerScroll:AddChild(button);
 
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -730,6 +730,8 @@ function WeakAuras.CreateFrame()
           local new_id = WeakAuras.FindUnusedId("New")
           local data = {id = new_id, regionType = regionType}
           WeakAuras.DeepCopy(WeakAuras.data_stub, data)
+          data.internalVersion = WeakAuras.InternalVersion();
+          WeakAuras.validate(data, WeakAuras.regionTypes[regionType].default)
           local parentData;
           if (not data.controlledChildren and previousPick) then
             if (type(previousPick) == "table") then
@@ -749,8 +751,6 @@ function WeakAuras.CreateFrame()
               end
             end
           end
-          data.internalVersion = WeakAuras.InternalVersion();
-          WeakAuras.validate(data, WeakAuras.regionTypes[regionType].default)
           WeakAuras.Add(data);
           WeakAuras.NewDisplayButton(data);
           if (parentData) then
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -695,12 +695,17 @@ function WeakAuras.CreateFrame()
 
   frame.PickOption = function(self, option)
     local parentData = GetPreviousGroupData(self.pickedDisplay);
+    local parentButton;
+    if (parentData) then
+      parentButton = WeakAuras.GetDisplayButton(parentData.id);
+    end
     self:ClearPicks();
+    if (parentButton) then
+      parentButton:Pick();
+    end
     self.moversizer:Hide();
     self.pickedOption = option;
     if(option == "New") then
-      newButton:Pick();
-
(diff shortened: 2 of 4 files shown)


mrbuds:
It is still missing an option to create them directly in a group as asked in #678 and #647

emptyrivers:
How is this workflow communicated to the user? How will the end user (who doesn’t read patch notes) learn that holding the CTRL key will enable this feature?

mrbuds:
i see 2 way to improve this
- add a text somewhere that say "hold CTRL to make more than 1 aura at a time"
- or add a checkbox on the bottom, left side "Create multiple auras at once" (i'm not sure on the correct wording)

emptyrivers:
I just want to make sure that the work you’re doing isn’t just completely ignored because it isn’t advertised enough 😀

emptyrivers:
How about this addition:

- add a button labeled “start batch” or similar. 
- after that button is clicked, or when ctrl is held, interpret button clicks on template options to be part of a batch
- while at least one template is queued in the batch, clicking the start batch button will clear the queue

mrbuds:
I added a new checkbox "Batch mode"
While checked it act like the ctrl key is down
When you ctrl+click it will activate the Batch mode and check the checkbox
When you unselect everything batch mode cancel
The "Batch" button show if batch mode is active and at least one label is selected
batch mode

mrbuds:
I would like that when you click "Create Auras" button, a popup dialog show asking if you want to add auras in a group

and then it works like when you have clicked on this

emptyrivers:
If others agree on that UI, then the code path would look something like:
- Finish interacting with templates
- WeakAuras.Add gets called a bunch of times
- PickDisplayMultiple gets called on the newly created auras
- Dialog box pops up
- OnAccept, set OptionsFrame to grouping mode

But to get this to work, will require that WeakAuras even understands how to start grouping on multiple auras, which it doesn't at the moment. I think the change required to do so would be relatively small, however.

emptyrivers:
(to be clear, I personally have no opinion on whether this UI is the correct way to do it)

mrbuds:
Written like that it does not seems hard to do.

However i will wait for stan and infus input on this before going further on that road.

mrbuds:
last commit create templates in the last selected group (or parent if it was not a group)

Things not finished:
- OptionsFrame.lua/frame.PickOption() use self:ClearPicks() clear last selection witch doesn't help understand the workflow.
- Didn't handle the case where "previousPick" is a table (multi-selection)
- If in "batch mode" and "previousPick" is nil (top level selected) should make a group and/or ask user for a group name and group type (dynamic or not)
- (not template) Make every new non-group regionType use "previousPick" ?

mrbuds:
short demo of what is done
demo

Stanzilla:
That's really cool!

emptyrivers:
If I want to create an aura which isn't part of a group, how can i do that? How can I know which group I'm going to be adding the new aura into (if e.g. I step away from the computer and forget what I've done)?

Stanzilla:
I thought he had CTRL pressed in the video when he clicked on stuff? I mean that's not really obvious to anyone but that's how it works, right?

emptyrivers:
That would be fine, I guess, but there's no indication in the gif of which group is being extended. I want that indicator (and a way to cancel or modify it).

Stanzilla:
I agree

mrbuds:
that's why i wrote:

**Things not finished** / Ideas

 -   OptionsFrame.lua/frame.PickOption() use self:ClearPicks() clear last selection witch doesn't help understand the workflow.

mrbuds:
And it also needs a better way to clear selection and select top level

mrbuds:
And yes as you said, maybe a big arrow pointing on where it's going to be added when you click on "New"

InfusOnWoW:
So, there are quite a few details that I don't like:
* Batch Mode being a checkbox feels off.
* The Batch button should be named "next" and be on the right, not on the left
* Single Selection vs Multi Selection behaves differntly. I think that's unavoidable but feels a bit strange
* The Batch Mode checkbox was already checked if the previous wizard run had it enabled.
* If you have only groups, creating a top level aura is impossible
* I got a error on creating a batch with no group selected

* Clicking new deselecting the group loses the focus. 

I do have two suggestions for a user interface:


Instead of having "New" as a entry in the list, have "New" as a toolbar button above the list. Thus, naturally the group would stay selected on clicking new. One problem, is that it's unclear how to create top level auras if you only have groups.

Or adding a "New" entry to all groups, which would look like this:
wowscrnshot_082918_230157

InfusOnWoW:
wowscrnshot_082918_225555

mrbuds:
How it looks with a button instead of the checkbox, with LockHighlight() and UnlockHighlight()

i1
i2
i2
i2

It's ok, but when you move your cursor on it it use LockHighlight() so you don't see that the click had an effect before you move the cursor out of the frame

mrbuds:
When you click on "New", it set the selection to where the aura will be done instead of selecting "New".
And if you click on a group or aura already selected, it remove the selection, allowing to make auras with parent = top level.

Without making big changes on the UI, this solve these problems:
- If you have only groups, creating a top level aura is impossible
- Clicking new deselecting the group loses the focus.

Now i'd like that if an aura is selected, new auras are created at this position, and instead of auto-selecting the parent aura it keep previous selection.

mrbuds:
demo with last commits: https://cdn.discordapp.com/attachments/350339895622893569/484741116885860359/2018-08-30_17-06-33.gif

Stanzilla:
Like it! That's how imagined it to work like from the beginning tbh

mrbuds:
last commit, a suggestion from Causese, show the "New" screen when an empty group is selected
https://cdn.discordapp.com/attachments/350339895622893569/484784032157138957/2018-08-30_19-57-15.gif

mrbuds:
off
on
on+selected

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -488,7 +488,11 @@ local methods = {
             end
           end
         else
-          WeakAuras.PickDisplay(data.id);
+          if (WeakAuras.IsDisplayPicked(data.id)) then
+            WeakAuras.ClearPicks(data.id);
```
InfusOnWoW: So this code is almost the same as the above one for IsCtrlKeyDown(). It might be easier to combine them.

And also I'm not sure how the interactions between normal clicks and ctrl+clicks should be, but I think it should be:

 _ | Normal Click | Ctrl Click
------------ | ------------- |  -------------
Unselected  |  Clear+Select | Add to selection
Single-Selection | Unselect (1) | Unselect
Multi-Selection | Clear + Select (2) | Unselect

Clear+Select = The clicked one is the only selected after the click.

(1) Is the one that is a bit off, because looking at explorer, that's now how it usually works. But we need that, so I'm okay with it.

(2) Is how explorer works, and different from what you have implemented.

But, I'm not sure how it behaved before, and you can convince me that what you have implemented is better or more consistent with how it worked before.

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -839,6 +849,11 @@ function WeakAuras.CreateFrame()
     else
       WeakAuras.PauseAllDynamicGroups();
       finishPicking();
+      if (data.controlledChildren and #data.controlledChildren == 0) then
```
InfusOnWoW: Can you explain the purpose of this code block?

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -912,30 +920,64 @@ function WeakAuras.CreateTemplateView(frame)
...
-              replaceTrigger(newView.data, item, subType);
-              replaceCondition(newView.data, item, subType);
-              newView.data.id = WeakAuras.FindUnusedId(item.title);
-              newView.data.load = {};
-              if (item.load) then
-                WeakAuras.DeepCopy(item.load, newView.data.load);
+          if ((IsControlKeyDown() or newView.BatchMode) and not newView.existingAura) then
+            local isPicked = false;
+            for k,v in pairs(newView.choosenItemBatch) do
+              if v.spell == item.spell then
```
InfusOnWoW: You can just check v == item.

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -912,30 +920,64 @@ function WeakAuras.CreateTemplateView(frame)
...
+              button.frame:UnlockHighlight();
+              for k,v in pairs(newView.choosenItemBatch) do
+                if v.spell == item.spell then
+                  newView.choosenItemBatch[k] = nil;
+                end
+              end
+              newView.choosenItemButtonsBatch[j] = nil;
+            else
+              button.frame:LockHighlight();
+              tinsert(newView.choosenItemBatch, item);
```
InfusOnWoW: If you use chooseItemBatch as a hash instead of a array, the code should get easier.

That is, use: choosenItemBatch[item] = true here;

To clear use:
choosenItemBatch[item] = nil;

And to check whether it is picked use if (choosenItemBatch[item])

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -962,7 +1007,16 @@ function WeakAuras.CreateTemplateView(frame)
       end
       button:SetFullWidth(true);
       button:SetClick(function()
-        if (newView.existingAura) then
+        if newView.batchStep then
+          for x, b in pairs(subTypesButtons) do
```
InfusOnWoW: x is a terrible variable name.

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -962,7 +1007,16 @@ function WeakAuras.CreateTemplateView(frame)
...
+            if (x == k) then
+              b.frame:LockHighlight();
+            else
+              b.frame:UnlockHighlight();
+            end
+          end
+          newView.choosenItemBatchSubType[item.spell] = subType;
```
InfusOnWoW: Index by item, not by item.spell. We can have multiple items with the same spell.

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -1164,6 +1254,17 @@ function WeakAuras.CreateTemplateView(frame)
       else
         newView.backButton:Show();
       end
+
+      -- batchButton
+      newView.batchButton:Hide();
```
InfusOnWoW: You already hide it at the start of the function.

InfusOnWoW:
* After creating a new group, the "New" selection should not offer "Group" nor "Dynamic Group".

* I think the next button should be with the other buttons. Same with "Create Auras"
* My preference for the order is: "Back", "Next/Create Auras", "Cancel". That's what windows would use, and that's also consistent with the button order of wow.
* The slider needs a bit of work. I was thinking something like this: https://cdn-images-1.medium.com/max/1600/1*FKTZMYiRPYmNuQ1pYJ9qSg.png Looking over the api of sliders, there are SetThumbTexture and SetBackdrop to modify it.

InfusOnWoW:
Or it might be better to try to skin a button.

mrbuds:
I'm starting to think the best solution is to remove the "batch mode" button/checkbox/slider and make it the default behavior. It would solve a few problems
- The button/checkbox/slider design => get rid of it
- The click / ctrl+click have a different behavior when batch mode On and Off => act like ctrl is always pressed instead
- 2 different screen with batch mode On or Off => only 1

Instead of showing the "Next" button only when >= 1 item selected, we can show it always and disable it when nothing is selected, and put the "!" in there.

Seems good?

WeakAurasTemplates/TriggerTemplates.lua
mrbuds: you are a genius

WeakAurasTemplates/TriggerTemplates.lua
mrbuds: Renamed both for `for index, subTypesButton in`
Btw, newView.batchStep was called like that because at the time i did it, i though i would make more screens after the selection is done (for group creation), and was intended to be of the type number later.

InfusOnWoW:
That would be a solution. For consistency sakes we then should have select on all wizard pages.

Or maybe instead of having a button/checkbox/slider to enable it, just have a text telling users that ctrl+click multi selects

WeakAurasTemplates/TriggerTemplates.lua
mrbuds: good point

WeakAurasTemplates/TriggerTemplates.lua
mrbuds: indeed, fixed for the "back" button too

mrbuds:
demo with last commits : https://i.imgur.com/MJf4xtI.gif

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -1179,6 +1281,62 @@ function WeakAuras.CreateTemplateView(frame)
     end
   end
 
+  local batchModeLabel = CreateFrame("FRAME", "batchModeLabel", newView.frame);
+  batchModeLabel:SetSize(300, 20);
+  local batchModeLabelString = batchModeLabel:CreateFontString(nil, "ARTWORK");
+  batchModeLabelString:SetFont(STANDARD_TEXT_FONT, 10); -- "OUTLINE"
+  batchModeLabelString:SetTextColor(1,1,1,1);
+  batchModeLabelString:SetText('|TInterface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon:0|t' .. L["Hold CTRL to make multiple auras at once."]);
```
Stanzilla: "Hold CTRL to create multiple auras at once"

mrbuds: amended with "create"

InfusOnWoW:
I think we should just go with that solution. It's not perfect, but I like it more than how the others looked. I'll look over the code a bit later.

PR #691, opened 2018-08-31
Templates fixes

mrbuds:
> fix totem glow
> typo in a description
> new "overlayGlow" subType for spells with procs
> added and fixed a few spells

### DIFF
--- a/WeakAurasTemplates/TriggerTemplatesData.lua
+++ b/WeakAurasTemplates/TriggerTemplatesData.lua
@@ -154,15 +154,14 @@ templates.class.WARRIOR = {
       icon = 464973
     },
     [3] = {
-      title = L["Cooldowns"],
+      title = L["Abilities"],
       args = {
         { spell = 100, type = "ability", requiresTarget = true, talent = {5,6}}, -- Charge
         { spell = 100, type = "ability", charges = true, requiresTarget = true, talent = 4, titleSuffix=" (2 Charges)"}, -- Charge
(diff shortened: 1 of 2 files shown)


WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -158,22 +158,28 @@ templates.class.WARRIOR = {
       args = {
         { spell = 100, type = "ability", requiresTarget = true, talent = {5,6}}, -- Charge
         { spell = 100, type = "ability", charges = true, requiresTarget = true, talent = 4, titleSuffix=" (2 Charges)"}, -- Charge
+        { spell = 121, type = "ability", requiresTarget = true}, -- Slam
```
InfusOnWoW: Unknown spell id: 121 and 129.

mrbuds: oops! I typed macro id there, it's fixed, and i also renamed Cooldowns to Abilities as you said on discord

InfusOnWoW:
I'm not sure whether Abilities is the best, but imho it's better than Cooldowns

Stanzilla:
Question, I noticed some templates have an unchecked "inverse" condition set, what is that used for exactly?

mrbuds:
"inverse" is checked by default because when a cooldown start we can't use it and we show when it will be ready 
cd
But when we show auras duration, we want the icon spiner to show we have it, and when it will be off : 
aura cd

PR #700, opened 2018-09-03
fix regression in multi-adding to dynamic group

emptyrivers:


emptyrivers:
Ah, silly me. I'll drop this then.

PR #703, opened 2018-09-03
multi-selection drag & drop + re-open last selection

mrbuds:
demo: https://i.imgur.com/LIrgzhW.gif

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -1061,40 +1061,21 @@ local methods = {
       self:Enable();
     end
   end,
-  ["SetDragging"] = function(self, data, drop, displayButtons)
+  ["SetDragging"] = function(self, data, drop)
     if data then
       -- self
-      if self.data.id == data.id or (displayButtons and data[self.data.id] == 1) then
+      if self.data.id == data.id or self.multi then
         if drop then
+          --self.frame:Show()
           self:Drop()
-          if displayButtons then
-            for _, button in pairs(displayButtons) do
-              button.frame:SetScript("OnClick", button.callbacks.OnClickNormal)
-              button.frame:EnableKeyboard(false);
-            end
-          else
-            self.frame:SetScript("OnClick", self.callbacks.OnClickNormal)
-            self.frame:EnableKeyboard(false); -- disables self.callbacks.OnKeyDown
-          end
+          self.frame:SetScript("OnClick", self.callbacks.OnClickNormal)
+          self.frame:EnableKeyboard(false); -- disables self.callbacks.OnKeyDown
         else
           Hide_Tooltip()
-          if displayButtons then
-            for _, button in pairs(displayButtons) do
-              button.frame:SetScript("OnClick", nil)
-              button.frame:EnableKeyboard(true);
-            end
-            self.multiTemp = {
-              displayButtons = displayButtons,
-              data = data
-            }
-          else
-            self.frame:SetScript("OnClick", nil)
-            self.frame:EnableKeyboard(true); -- enables self.callbacks.OnKeyDown
-          end
+          self.frame:SetScript("OnClick", nil)
+          self.frame:EnableKeyboard(true); -- enables self.callbacks.OnKeyDown
           self:Drag()
         end
-        -- skip draged buttons after the first
-      elseif (displayButtons and data[self.data.id]) then
         -- invalid targets
       elseif not self.data.parent and not self:IsGroup()
       then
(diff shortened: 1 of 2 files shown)


emptyrivers:
I don't really like how all of the buttons lift up like that. I could see it getting a bit hairy if e.g. you select a couple dozen display buttons. I think having a dummy display that just attaches to the cursor might be better.

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -268,7 +268,8 @@ local Actions = {
           group.callbacks.UpdateExpandButton();
           group:ReloadTooltip()
         else
-          error("Calling 'Group' with invalid groupId. Reload your UI to fix the display list.")
+          WeakAuras.Add(source.data)
+          -- error("Calling 'Group' with invalid groupId. Reload your UI to fix the display list.")
```
emptyrivers: Commented code that doesn't look likely to ever be uncommented

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -748,8 +749,10 @@ local methods = {
 
     function self.callbacks.OnDragStart()
       if WeakAuras.IsImporting() or self:IsGroup() then return end;
-      WeakAuras.PickDisplay(data.id)
-      WeakAuras.SetDragging(data)
+      if not WeakAuras.tempGroup.controlledChildren then
```
emptyrivers: As far as I know, tempGroup always has controlled children.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4075,8 +4075,19 @@ end
 
 function WeakAuras.SetDragging(data, drop)
   WeakAuras_DropDownMenu:Hide()
-  for id, button in pairs(displayButtons) do
-    button:SetDragging(data, drop)
+  if (frame.pickedDisplay == tempGroup and #tempGroup.controlledChildren > 0) then
+    local datas = {}
```
emptyrivers: the plural of data is still data

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
mrbuds: indeed, good catch !

mrbuds:
I though about your idea of a dummy display, and i think a way to do that would be to only show a subset of the selection and add a transparency mask on the .. mhm .. let's say 3rd item
multi-selection subset
I made this with gimp and i don't know if it is possible with the api

mrbuds:
i tried to implement the mock-up, best i could get for now is this
multi dragging

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -1056,20 +1061,40 @@ local methods = {
       self:Enable();
     end
   end,
-  ["SetDragging"] = function(self, data, drop)
+  ["SetDragging"] = function(self, data, drop, displayButtons)
```
InfusOnWoW: This is the wrong place to implement most of this code.

This is the "SetDraggin" method of a single button. And that one should not care about other buttons.

You should move most of this code to WeakAuras.SetDragging

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -1088,37 +1113,97 @@ local methods = {
...
+        if (self.dragging) then
+          self:Drop(true)
+        end
       end
     end
   end,
   ["ShowTooltip"] = function(self)
   end,
   ["Drag"] = function(self)
-    local uiscale, _, y = UIParent:GetScale(), GetCursorPosition()
```
InfusOnWoW: Same here, this is the Drag method of a button. It should not know anything about other buttons.

So the whole logic should be in SetDragging

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4075,8 +4076,19 @@ end
...
-  for id, button in pairs(displayButtons) do
-    button:SetDragging(data, drop)
+  if (frame.pickedDisplay == tempGroup and #tempGroup.controlledChildren > 0) then
+    local children = {}
+    for index, childId in pairs(tempGroup.controlledChildren) do
+      local data = WeakAuras.GetData(childId);
+      children[data.id] = index
+    end
+    for id, button in pairs(displayButtons) do
+      button:SetDragging(data and children, drop, displayButtons);
```
InfusOnWoW: Here for each button you drag, you call button:SetDragging, which will then iterate over all displayButtons, essentially each call to button:SetDragging doing the same work.

mrbuds: button:SetDragging iterate over all displayButtons only once `(displayButtons and data[self.data.id] == 1)`

mrbuds:
Thanks for your comments, i moved the iteration on selected buttons on WeakAuras.SetDragging, it made the code much easier.
Also, i was moving buttons out of screen because Hide() removed the drop event if i didnt select the first item from the list, this issue is now solved by using the selected button instead of the 1st from the list.

The last issue i see is with how i handle the temporary "WA" icon, if the button i'm dragging already have an icon, it will have both icon at the same time while dragging

mrbuds:
Last commit fix a regression when closing WeakAurasOptions with more than one button selected

It also implement a suggestion from @Stanzilla on discord to re-select previous button(s) when re-opening WeakAurasOptions

Stanzilla:
merged as https://github.com/WeakAuras/WeakAuras2/pull/744

PR #707, opened 2018-09-05
Refactor

emptyrivers:


InfusOnWoW:
Yeah, we should try combine a few cleanups, so that we don't do that again in 30 locals.

emptyrivers:
Total of locals in WeakAuras.lua is now down to about 130.

WeakAuras/WeakAuras.lua
```diff
@@ -2528,7 +2528,7 @@ function WeakAuras.Modernize(data)
   -- Version 6 was introduced July 30, 2018 in BFA
   -- Changes were entirely within triggers, so no code runs here
 
-  -- Version 7 was introduced %DATE% in BFA
+  -- Version 7 was introduced September 1 in BFA
```
InfusOnWoW: The exact day is not relevant, the year though is.

WeakAuras/WeakAuras.lua
```diff
@@ -1755,15 +1208,15 @@ function WeakAuras.UnloadAll()
   end
   wipe(timers);
 
-  for id in pairs(conditionChecksTimers.recheckTime) do
```
InfusOnWoW: Instead of making conditionChecksTimers and checkdConditions globally available, I'd prefer if there would be functions:

Rename/UnloadAll/Unload/Delete prefixed with Conditions (either in its own table or just in the method name).

WeakAuras/AuraEnvironment.lua
```diff
@@ -50,3 +54,138 @@ WeakAuras.helperFunctions = {
...
+  securecall = true,
+}
+
+local overrideFunctions = {
+  ActionButton_ShowOverlayGlow = WeakAuras.ShowOverlayGlow,
+  ActionButton_HideOverlayGlow = WeakAuras.HideOverlayGlow,
+}
+
+local aura_environments = {};
+WeakAuras.AuraEnvironments = aura_environments
```
InfusOnWoW: Same here. I would really prefer if that's never available via WeakAuras. People will try to do funny things with it otherwise.

WeakAuras/Profiling.lua
```diff
@@ -0,0 +1,162 @@
...
+
+local function PrintOneProfile(name, map, total)
+  if (map.count ~= 0) then
+    print(name, " ERROR: count is not zero:", map.count);
+  end
+  local percent = "";
+  if (total) then
+    percent = ", " .. string.format("%.2f", 100 * map.elapsed / total) .. "%";
+  end
+  print(name, " ", string.format("%.2f", map.elapsed), " ms", percent);
```
Stanzilla: "ms"

PR #725, opened 2018-09-08
Create an Aura Environment for group regions as well.

emptyrivers:
This is necessary because the Animated Expand & Collapse feature animates the dynamic group region itself.
This eventually calls ActivateAuraEnvironment, which tries to get the aura environment for the dynamic group.
Before the aura_env support was split off, this would fallback to an empty table, which is rather wasteful since a new extra table would be
created every frame for every dynamic group which is animating an expand or collapse.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -2583,6 +2583,7 @@ local function pAdd(data)
     error("Improper arguments to WeakAuras.Add - id not defined");
   elseif (data.controlledChildren) then
     db.displays[id] = data;
+    WeakAuras.CreateAuraEnvironment(id)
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
To add some more context: This broke in 99ea1fef877ad15f1cc639cb6d7beb34f2e3d7bd, where instead of the old code  WeakAuras.ActivateAuraEnvironment:
 -- Reset the environment if we haven't completed init, i.e. if we add/update/replace a WeakAura
aura_environments[id] = {};

the new code was:
 -- Reset the environment if we haven't completed init, i.e. if we add/update/replace a WeakAura
aura_environments[id] = {};

PR #729, opened 2018-09-08
revert back to lazy aura environment creation

emptyrivers:
Unfortunately, pAdd does still need to flag the aura as not initialized (since every change to aura data which would require a re-initialization goes through Add), so we do still need a method to do that, so right now it kind of does both.

### DIFF
--- a/WeakAuras/AuraEnvironment.lua
+++ b/WeakAuras/AuraEnvironment.lua
@@ -105,7 +105,7 @@ function WeakAuras.IsEnvironmentInitialized(id)
   return environment_initialized[id]
 end
 
-function WeakAuras.CreateAuraEnvironment(id)
+function WeakAuras.ResetAuraEnvironment(id)
   aura_environments[id] = {}
   environment_initialized[id] = false
 end
@@ -132,6 +132,7 @@ function WeakAuras.ActivateAuraEnvironment(id, cloneId, state)
   else
     if environment_initialized[id] then
       -- Point the current environment to the correct table
+      aura_environments[id] = aura_environments[id] or {}
       current_aura_env = aura_environments[id]
       current_aura_env.cloneId = cloneId
       current_aura_env.state = state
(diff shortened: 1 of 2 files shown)


WeakAuras/AuraEnvironment.lua
```diff
@@ -105,7 +105,7 @@ function WeakAuras.IsEnvironmentInitialized(id)
   return environment_initialized[id]
 end
 
-function WeakAuras.CreateAuraEnvironment(id)
+function WeakAuras.ResetAuraEnvironment(id)
   aura_environments[id] = {}
```
InfusOnWoW: That line is unnecessary.

WeakAuras/WeakAuras.lua
```diff
@@ -2583,12 +2583,14 @@ local function pAdd(data)
     error("Improper arguments to WeakAuras.Add - id not defined");
   elseif (data.controlledChildren) then
     db.displays[id] = data;
-    WeakAuras.CreateAuraEnvironment(id)
+    WeakAuras.ResetAuraEnvironment(id)
```
InfusOnWoW: you don't need to call this here.

WeakAuras/WeakAuras.lua
```diff
@@ -2583,12 +2583,14 @@ local function pAdd(data)
...
     WeakAuras.SetRegion(data);
   else
+
     if (not data.triggers.activeTriggerMode or data.triggers.activeTriggerMode > #data.triggers) then
       data.triggers.activeTriggerMode = WeakAuras.trigger_modes.first_active;
     end
+    WeakAuras.ResetAuraEnvironment(id)
```
InfusOnWoW: or here

WeakAuras/WeakAuras.lua
InfusOnWoW: As in do it once for both types

PR #731, opened 2018-09-08
Smallpatches

InfusOnWoW:


WeakAuras/WeakAuras.lua
```diff
@@ -2579,18 +2579,19 @@ end
 
 local function pAdd(data)
   local id = data.id;
+  db.displays[id] = data;
+
+  WeakAuras.ClearAuraEnvironment(id);
   if not(id) then
```
emptyrivers: This doesn't make sense, I think. If pAdd is going to assume that the id isn't nil, then this check shouldn't exist.

InfusOnWoW: Indeed.

PR #739, opened 2018-09-10
Add a popup frame for profiling prints

### LINKED ISSUE #702: [Profiling] Print into a frame
We should print the results from profiling into a frame instead of chat so it is easier to copy paste it

Stanzilla:
Adds a popup frame that we print the profiling data into instead of spamming the chat.

Currently looks like this

image

the frame is draggable via right clicks.

Pending an answer from @kesava-wow if I can use his popup code since KuiNameplates is All Rights Reserved. Edit: Got the ok!

Fixes #702

### DIFF
--- a/WeakAuras/Profiling.lua
+++ b/WeakAuras/Profiling.lua
@@ -95,13 +95,13 @@ end
 
 local function PrintOneProfile(name, map, total)
   if (map.count ~= 0) then
-    print(name, " ERROR: count is not zero:", map.count)
+    print(name, "ERROR: count is not zero:", map.count)
   end
   local percent = ""
   if (total) then
     percent = ", " .. string.format("%.2f", 100 * map.elapsed / total) .. "%"
   end
-  print(name, " ", string.format("%.2f", map.elapsed), "ms", percent)
+  print(name, string.format("%.2f", map.elapsed) .. "ms" .. percent)
 end
 
 local function SortProfileMap(map)
@@ -138,11 +138,11 @@ function WeakAuras.PrintProfile()
 
   print("--------------------------------")
   prettyPrint(L["EXPERIMENTAL Profiling Data:"])
-  PrintOneProfile("Total Time:    ", profileData.systems.time)
-  PrintOneProfile("Time inside WA:", profileData.systems.wa)
-  print("% Time spent inside WA:", string.format("%.2f", 100 * profileData.systems.wa.elapsed / profileData.systems.time.elapsed))
+  PrintOneProfile("|cff9900FFTotal Time:|r", profileData.systems.time)
+  PrintOneProfile("|cff9900FFTime inside WA:|r", profileData.systems.wa)
+  print("|cff9900FF% Time spent inside WA:|r", string.format("%.2f", 100 * profileData.systems.wa.elapsed / profileData.systems.time.elapsed))
   print("")
-  print("Systems:")
+  print("|cff9900FFSystems:|r")
 
   for i, k in ipairs(SortProfileMap(profileData.systems)) do
     if (k ~= "time" and k~= "wa") then
@@ -151,9 +151,9 @@ function WeakAuras.PrintProfile()
   end
 
   print("")
-  print("Auras:")
+  print("|cff9900FFAuras:|r")
   local total = TotalProfileTime(profileData.auras)
-  print("Total Time attributed to auras: ", total)
+  print("Total Time attributed to auras: ", floor(total) .."ms")
   for i, k in ipairs(SortProfileMap(profileData.auras)) do
     PrintOneProfile(k, profileData.auras[k], total)
   end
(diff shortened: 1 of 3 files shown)


krazyito65:
frame resizeable?

Stanzilla:
Nope, figured most people would click in and copy anyway.

krazyito65:
Fair enough.   Another idea I had would to make it look nicer:

time `\t\t` percent `\t\t` aura_name

That way they line up in nice columns.

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+
+      str = (str and str .. "|cff999999,|r " or "|cff999999{|r ") .. "|cffffff99[" .. tostring(k) .. "]|r |cff999999=|r |cffffffff" .. tostring(v) .. "|r"
+    end
+  end
+  return (str or "{ ") .. " }"
+end
+
+local debugpopup
+local function CreateDebugPopup()
+  local p = CreateFrame("EditBox", "WADebugEditBox", UIParent)
```
InfusOnWoW: I think p is a awful variable name.

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+  end
+
+  p:SetScript("OnEscapePressed", function(self)
+    self:ClearFocus()
+    self:Hide()
+    self.ScrollFrame:Hide()
+    self.Background:Hide()
+  end)
+
+  local s = CreateFrame("ScrollFrame", "WADebugEditBoxScrollFrame", UIParent, "UIPanelScrollFrameTemplate")
```
InfusOnWoW: Same with s.

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+  s:SetMovable(true)
+  s:SetFrameStrata("DIALOG")
+  s:SetSize(450, 300)
+  s:SetPoint("CENTER")
+  s:SetHitRectInsets(-4, -4, -4, -4)
+  s:SetScrollChild(p)
+  s:Hide()
+
+  s:SetScript("OnMouseDown", function(self, button)
+    if button == "RightButton" and not self.is_moving then
```
InfusOnWoW: That's rather non-standard for our dialogss.

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+  debugpopup = p
+end
+
+WeakAuras.DebugPopup = function()
+  -- create/get and return reference to debug EditBox
+  if not debugpopup then
+    CreateDebugPopup()
+  end
+
+  debugpopup:Hide()
```
InfusOnWoW: That's a bit strange place to call hide.

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+        k = '"' .. k ..'"'
+      end
+
+      str = (str and str .. "|cff999999,|r " or "|cff999999{|r ") .. "|cffffff99[" .. tostring(k) .. "]|r |cff999999=|r |cffffffff" .. tostring(v) .. "|r"
+    end
+  end
+  return (str or "{ ") .. " }"
+end
+
+local debugpopup
```
InfusOnWoW: Maybe call this: profilePopup?

WeakAuras/Profiling.lua
```diff
@@ -6,6 +6,132 @@ local profileData = {}
...
+  -- create/get and return reference to debug EditBox
+  if not debugpopup then
+    CreateDebugPopup()
+  end
+
+  debugpopup:Hide()
+  return debugpopup
+end
+
+local popup = WeakAuras:DebugPopup()
```
InfusOnWoW: Isn't debugpopup in the same scope anyway?

Stanzilla: Yes but I wanted to keep it this way in case we may want to use it somewhere else as well. Can change it for now though.

WeakAuras/Profiling.lua
Stanzilla: Why exactly? Also remember, I took this directly from KNP.

WeakAuras/Profiling.lua
Stanzilla: So left click for drag then?

WeakAuras/Profiling.lua
Stanzilla: Fixed

WeakAuras/Profiling.lua
Stanzilla: Fixed

WeakAuras/Profiling.lua
InfusOnWoW: Yeah, we have a titlebar for that.

PR #740, opened 2018-09-10
migrate condition for aura buffed/debuffed

mrbuds:
fix templates with auras buffed/debuffed


Stanzilla:
merged with https://github.com/WeakAuras/WeakAuras2/pull/745

PR #749, opened 2018-09-12
Fix WA_IterateGroupMembers to iterate over all  party members.

### LINKED ISSUE #748: WA_IterateGroupMembers does not iterate over all members.
https://github.com/WeakAuras/WeakAuras2/blob/4640b90713d9109f26cdd8e3fe9ad786abbc1b31/WeakAuras/AuraEnvironment.lua#L29

When I was using rivers snippet of this, I found that forcing a party, it would always be missing party4.  This is because `GetNumSubgroupMembers()` already returns a 4 and you don't need to -1.

krazyito65:
Fixes #748


InfusOnWoW:
Please write a better commit message

krazyito65:
Done

PR #755, opened 2018-09-15
BuffTrigger Rewrite

InfusOnWoW:
Introudce a new buff/debuff tracking system

Instead of each trigger iterating over all auras, keep maps of names/spellids to triggers and use those to cut down on the amount of checking that needs to happen.

Also combine fullscan and normal auras, so that the new trigger has a superset of both modes.
Enable all of those features for group triggers, and add two more variants of group triggers for nameplates and arena units.

Use the same infrastructure also for multi-target tracking, which allows e.g. a combined view of all matches.


WeakAuras/GenericTrigger.lua
```diff
@@ -829,9 +829,21 @@ function GenericTrigger.Add(data, region)
             else
               error("Improper arguments to WeakAuras.Add - no event prototype can be found for event type \""..trigger.event.."\" and default prototype reset failed.");
             end
-          elseif(trigger.event == "Combat Log" and not (trigger.subeventPrefix..trigger.subeventSuffix)) then
```
InfusOnWoW: This cannot ever happen. If we succuesfully concanated prefix and suffix, then that value can never be nil.

WeakAuras/GenericTrigger.lua
```diff
@@ -829,9 +829,21 @@ function GenericTrigger.Add(data, region)
...
-            error("Improper arguments to WeakAuras.Add - event type is \"Combat Log\" but subevent is not defined");
           else
+            if (trigger.event == "Combat Log") then
```
InfusOnWoW: This used to be done in special set methods for those two properties. Moving the code here, makes the code around the trigger options not have to care about that special case of the combat log trigger.

WeakAurasOptions/GenericTrigger.lua
```diff
@@ -10,9 +10,6 @@ local custom_trigger_types = WeakAuras.custom_trigger_types;
 local eventend_types = WeakAuras.eventend_types;
 
 function WeakAuras.GetGenericTriggerOptions(data, trigger, untrigger)
-  if (not trigger) then
-    return {};
-  end
```
InfusOnWoW: This was never possible, trigger was never nil.

WeakAurasOptions/GenericTrigger.lua
```diff
@@ -67,6 +64,38 @@ function WeakAuras.GetGenericTriggerOptions(data, trigger, untrigger)
   end
 
   local options = {
+    typedesc = {
```
InfusOnWoW: This moved from the general trigger options to the generic trigger options, as selection between status and event is generic trigger specific

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -350,7 +350,7 @@ AceGUI:RegisterLayout("ButtonsScrollLayout", function(content, children, skipLay
   end
 end)
 
-function WeakAuras.ConstructOptions(prototype, data, startorder, subPrefix, subSuffix, triggernum, triggertype, unevent)
+function WeakAuras.ConstructOptions(prototype, data, startorder, triggernum, triggertype, unevent)
```
InfusOnWoW: subPrefix and subSuffix were unused. I didn't dive into the histroy to figure when they were last used, as I can't even remember when this code last changed.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -2629,17 +2632,14 @@ function WeakAuras.ReloadTriggerOptions(data)
   local iconCache = spellCache.Get();
   WeakAuras.EnsureOptions(id);
 
-  local trigger, untrigger, appendToTriggerPath, appendToUntriggerPath;
+  local appendToTriggerPath, appendToUntriggerPath;
```
InfusOnWoW: This is a bit more complex.

For multi-selection we construct the options twice:
* Once for each aura.
* Once for the whole selection

The whole selection options is then cleaned of all "functions", and those functions are replaced by functions that forward the call to each individual aura with some intelligent merging.

The trigger and untrigger values for the options creation are only used inside the functions, so while they are crucial for the per aura options creation, they aren't needed for creating options for the whole selection

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3003,47 +3002,13 @@ function WeakAuras.ReloadTriggerOptions(data)
       end,
       order = 2
     },
-    typedesc = {
```
InfusOnWoW: This moved to the generic trigger, as it is generic trigger specific

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3003,47 +3002,13 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      end
-    },
   };
 
   local function hideTriggerCombiner()
     return not (data.triggers.disjunctive == "custom")
   end
   WeakAuras.AddCodeOption(trigger_options, data, L["Custom"], "custom_trigger_combination", 0.1, hideTriggerCombiner, {"triggers", "customTriggerLogic"}, false);
 
-  trigger_options = union(trigger_options, WeakAuras.GetGenericTriggerOptions(data, trigger, untrigger));
```
InfusOnWoW: This adds the common generic trigger options and custom trigger options. This should only done for generic triggers, but needs a few more steps to reach the right place.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
       WeakAuras.ReloadTriggerOptions(data);
     end
 
-    removeFuncs(displayOptions[id]);
```
InfusOnWoW: All code paths, whether for buff or generic trigger functions called this a second time after adding the trigger options, so just do it once after adding the trigger options

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-
+    trigger_options = union(trigger_options, WeakAuras.GetGenericTriggerOptions(data));
```
InfusOnWoW: This isn't yet the right place, but it's closer.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
     if(optionTriggerChoices[id] >= 1 and getAll(data, {"trigger", "type"}) == "aura") then
-      local aura_options = WeakAuras.GetBuffTriggerOptions(data, trigger);
-      displayOptions[id].args.trigger.args = union(trigger_options, aura_options);
```
InfusOnWoW: The overall code got simpler by assinging to trigger_options here and doing the assignment for displayOptions[id].args.trigger.args below.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      removeFuncs(displayOptions[id].args.trigger);
```
InfusOnWoW: This is now done once for all different code paths.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      displayOptions[id].args.trigger.args.type.set = options_set;
```
InfusOnWoW: This was done below again, there was no point in doing it here.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
+      local aura_options = WeakAuras.GetBuffTriggerOptions(data);
+      trigger_options = union(trigger_options, aura_options);
     elseif(optionTriggerChoices[id] >= 1 and (getAll(data, {"trigger", "type"}) == "event" or getAll(data, {"trigger", "type"}) == "status")) then
       local event = getAll(data, {"trigger", "event"});
       local unevent = getAll(data, {"trigger", "unevent"});
       if(event and WeakAuras.event_prototypes[event]) then
-        local trigger_options_created;
```
InfusOnWoW: This code block fetched subeventPrefix and subeventSuffix. But those arguments to ConstructFunction were ignored anyway, so that whole code block was unnecessary.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-            displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[event], data, 10, subeventPrefix, subeventSuffix, optionTriggerChoices[id], nil, unevent));
-          end
-        end
-
-        if not(trigger_options_created) then
-          displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[event], data, 10, nil, nil, optionTriggerChoices[id], nil, unevent));
-        end
-      else
-        displayOptions[id].args.trigger.args = union(trigger_options, {});
-        removeFuncs(displayOptions[id].args.trigger);
```
InfusOnWoW: This is now done below once for all trigger types.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      end
-      removeFuncs(displayOptions[id].args.trigger);
-      replaceNameDescFuncs(displayOptions[id].args.trigger, data);
-      replaceImageFuncs(displayOptions[id].args.trigger, data);
-      replaceValuesFuncs(displayOptions[id].args.trigger, data);
```
InfusOnWoW: These wer done again for all trigger types below and thus doing it in the generic trigger code was superflous

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-
-      if(displayOptions[id].args.trigger.args.unevent) then
```
InfusOnWoW: The generic trigger options setter is fine for those options (after adding ReloadTriggerOptions to it)

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      if(displayOptions[id].args.trigger.args.subeventSuffix) then
-        displayOptions[id].args.trigger.args.subeventSuffix.set = options_set;
-      end
-
-      if(displayOptions[id].args.trigger.args.type) then
-        displayOptions[id].args.trigger.args.type.set = options_set;
+        trigger_options = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[event], data, 10, optionTriggerChoices[id], nil, unevent));
       end
-      if(displayOptions[id].args.trigger.args.event) then
-        displayOptions[id].args.trigger.args.event.set = options_set;
```
InfusOnWoW: This was done below anyway.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      end
-    else
-      displayOptions[id].args.trigger.args = trigger_options;
-      removeFuncs(displayOptions[id].args.trigger);
     end
+    displayOptions[id].args.trigger.args = trigger_options;
+
+    removeFuncs(displayOptions[id]);
```
InfusOnWoW: This was done in each code path individually, doing it once for all is better.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3056,59 +3021,23 @@ function WeakAuras.ReloadTriggerOptions(data)
...
+    replaceNameDescFuncs(displayOptions[id], data);
```
InfusOnWoW: Those 3 calls, merely moved. The code between the old and new place does not affect those functions, and keeping them all together makes the code easier to understand

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3187,14 +3112,16 @@ function WeakAuras.ReloadTriggerOptions(data)
...
+    displayOptions[id].args.load.args = WeakAuras.ConstructOptions(WeakAuras.load_prototype, data, 10, optionTriggerChoices[id], "load");
     removeFuncs(displayOptions[id].args.load);
     replaceNameDescFuncs(displayOptions[id].args.load, data);
     replaceImageFuncs(displayOptions[id].args.load, data);
     replaceValuesFuncs(displayOptions[id].args.load, data);
 
     WeakAuras.ReloadGroupRegionOptions(data);
   else -- One aura selected
+    local triggerChoice = optionTriggerChoices[id];
+    local trigger, untrigger = data.triggers[triggerChoice].trigger, data.triggers[triggerChoice].untrigger;
```
InfusOnWoW: This moved from the top to here.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3203,40 +3130,21 @@ function WeakAuras.ReloadTriggerOptions(data)
...
     end
+    trigger_options = union(trigger_options, WeakAuras.GetGenericTriggerOptions(data, trigger, untrigger));
     local triggernum = optionTriggerChoices[id];
     if(trigger.type == "aura") then
       local aura_options = WeakAuras.GetBuffTriggerOptions(data, trigger);
-      displayOptions[id].args.trigger.args = union(trigger_options, aura_options);
+      trigger_options = union(trigger_options, aura_options);
     elseif(trigger.type == "event" or trigger.type == "status") then
       if(WeakAuras.event_prototypes[trigger.event]) then
-        if(trigger.event == "Combat Log") then
```
InfusOnWoW: Same reasoning as in the multi selection code path. ConstructOption does not care about subeventPrefix nor subEventSuffix

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3203,40 +3130,21 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-          displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[trigger.event], data, 10, (trigger.subeventPrefix or ""), (trigger.subeventSuffix or ""), optionTriggerChoices[id]));
-        else
-          displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[trigger.event], data, 10, nil, nil, optionTriggerChoices[id]));
-        end
-        if(displayOptions[id].args.trigger.args.unevent) then
-          displayOptions[id].args.trigger.args.unevent.set = options_set;
-        end
-        if(displayOptions[id].args.trigger.args.subeventPrefix) then
```
InfusOnWoW: This sepcial setter function code is now in GenericTrigger.Add() t

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3203,40 +3130,21 @@ function WeakAuras.ReloadTriggerOptions(data)
...
-      displayOptions[id].args.trigger.args = union(trigger_options, aura_options);
+      trigger_options = union(trigger_options, aura_options);
     elseif(trigger.type == "event" or trigger.type == "status") then
       if(WeakAuras.event_prototypes[trigger.event]) then
-        if(trigger.event == "Combat Log") then
-          displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[trigger.event], data, 10, (trigger.subeventPrefix or ""), (trigger.subeventSuffix or ""), optionTriggerChoices[id]));
-        else
-          displayOptions[id].args.trigger.args = union(trigger_options, WeakAuras.ConstructOptions(WeakAuras.event_prototypes[trigger.event], data, 10, nil, nil, optionTriggerChoices[id]));
-        end
-        if(displayOptions[id].args.trigger.args.unevent) then
```
InfusOnWoW: The generic setter function is fine.

WeakAuras/Types.lua
```diff
@@ -1440,6 +1446,33 @@ WeakAuras.bufftrigger_progress_behavior_types = {
...
 }
 
+WeakAuras.bufftrigger_2_progress_behavior_types = {
+  showOnActive = L["Aura(s) Found"],
+  showOnMissing = L["Aura(s) Missing"],
+  showAlways= L["Always"]
+}
+
+WeakAuras.bufftrigger_2_preferred_match_types =
+{
```
emptyrivers: opening brace here is on a newline, which is inconsistent with your other additions (and the rest of the file)

WeakAurasOptions/GenericTrigger.lua
```diff
@@ -66,56 +63,8 @@ function WeakAuras.GetGenericTriggerOptions(data, trigger, untrigger)
...
-      hidden = function() return not (trigger.type == "event" and trigger.event == "Combat Log" and subevent_actual_prefix_types[trigger.subeventPrefix]); end
-    },
-    spacer_suffix = {
-      type = "description",
-      name = "",
-      order = 9.1,
-      hidden = function() return not (trigger.type == "event" and trigger.event == "Combat Log"); end
-    },
+  local customOptions =
+  {
```
emptyrivers: also here

WeakAuras/WeakAuras.toc
```diff
@@ -29,6 +29,7 @@ Profiling.lua
 WeakAuras.lua
 Transmission.lua
 BuffTrigger.lua
+BuffTrigger2.lua
```
emptyrivers: You have translation phrases in this file, so babelfish needs to know about the file. Same with the Options file.

Issue #753, opened 2018-09-15
Pattern matching: text box can't handle some special characters


mrbuds:
**Describe the bug**
I want to use a tiny regex with "pattern match" option on bigwigs timer trigger `^[^<]`

When i press enter it raise an error
`[string "--[[ Error in 'aura name']:4: unexpected symbol near ']'"`

This pattern is a valid lua regex pattern:
```
> print(("<hello"):match("^[^<]"))
nil

> print(("world"):match("^[^<]"))
w
```

I tried to escape [ ] with \ with no success

To give a bit more context, i want to show timer for a spell but bigwigs can also show a castbar for same spellid with message "<casting: spellname>", which i want to filter.
So i need my trigger to match a spell id and ignore messages starting with "<"

InfusOnWoW:
Fixed for the next release.

PR #757, opened 2018-09-15
fix DBM trigger

mrbuds:
Use "[=[" instead of "[[" as string interpolation.

linked too #753 #754

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2782,10 +2782,10 @@ WeakAuras.event_prototypes = {
 
       if (trigger.use_message) then
         local ret2 = [=[
-          local triggerMessage = [[%s]]
+          local triggerMessage = %s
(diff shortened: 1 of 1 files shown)


emptyrivers:
Why not use %q instead of trying to add quotes to %s? This is what %q is for, last I checked.

Stanzilla:
Yup, should probably use that everywhere

InfusOnWoW:
Point is, I just pushed a fix for the BigWigs trigger where I didn't use %q either. So I think we can merge this, and @mrbuds can decide whether that's something he wants to do or not.

emptyrivers:
Well, I wasn’t suggesting to replace %s with %q wholesale. %q formats strings such that when they’re interpreted again by lua via loadstring or similar, they remain as string values, so it only makes sense to use %q when you want a string literal in the triggerfunction.

Stanzilla:
In any case, a thing for a new PR

PR #759, opened 2018-09-15
templates: fix 2 wrong spellids

mrbuds:
oops i erased 1st commit


mrbuds:
"wormed" reported on discord that his auras would not load with this error: https://pastebin.com/GFKCunuP
I could not reproduce the problem with https://wago.io/rk1HiNozM but i think this commit fix the problem
I tried to contact him to see if this fix his problem but no answer

PR #761, opened 2018-09-16
ensure that the trigger reference actually refers to a trigger

emptyrivers:
It seems that somehow some people have gotten auras with conditions that refer to nonexistent triggers. No idea how that happened, but we can guard against that in this modernization easily enough.
WoWAce Ticket 1227


emptyrivers:
Ah, @mrbuds already has the exact same fix. I'll drop this.

PR #776, opened 2018-09-22
Improved Health, Power and Cast Templates

mrbuds:
Power
- new type "power"
- inverse = false
- left text and icon removed
- show cost of casts

Health
- new type "health"
- inverse = false
- left text and icon removed
- show incoming heal and absorb

Cast
- new type "cast"
- inverse = false
- condition: uninterruptible = red

Shaman
- removed totem empty category
- added missing mana templates for the 3 specs

### DIFF
--- a/WeakAurasTemplates/TriggerTemplates.lua
+++ b/WeakAurasTemplates/TriggerTemplates.lua
@@ -112,6 +112,10 @@ local checks = {
   overlayGlow = {
     variable = "show",
     value = 1,
+  },
+  uninterruptible = {
+    variable = "interruptible",
+    value = 0,
   };
 }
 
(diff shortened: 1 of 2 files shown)


WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -239,6 +247,48 @@ local function createTotemTrigger(triggers, position, item)
...
+  triggers[position] = {
+    trigger = {
+      type = "status",
+      event = "Power",
+      unevent = "auto",
+      use_unit = true,
+      unit = "player",
+      use_powertype = true,
+      use_showCost = true,
+      powertype = item.powertype
```
emptyrivers: Here, you don't have a comma,

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -239,6 +247,48 @@ local function createTotemTrigger(triggers, position, item)
...
+  triggers[position] = {
+    trigger = {
+      type = "status",
+      event = "Health",
+      unit = "player",
+      use_unit = true,
+      unevent = "auto",
+      use_absorbMode = true,
+      use_showAbsorb = true,
+      use_showIncomingHeal = true,
```
emptyrivers: and here, you do.

mrbuds: indeed when you see it, you can't miss them anymore, i added a bunch

Issue #406, opened 2018-05-08
Early Preview: LibDeflate, pure Lua compressor which compresses much better, with similar speed to LibCompress


SafeteeWoW:
Repository of LibDeflate: https://github.com/SafeteeWoW/LibDeflate
Commit when I wrote this: https://github.com/SafeteeWoW/LibDeflate/tree/c1510af970ccd092235d46f1248181f20bb50f18

Hello WeakAuras developers:
     I would like to give an early preview of my recently implemented pure lua compressor and decompressor, called "LibDeflate". This library uses [DEFLATE](https://tools.ietf.org/html/rfc1951) compression algorithm. The library provides much better compression ratio than LibCompress, but the compress/decompression speed is similar to LibCompress. I believe this library makes WeakAuras transmission in-game much faster.

_I hope you can have a try of the library and give me some feedback, and I want to know if WeakAuras2 is willing to switch to LibDeflate as the compression library._

---
### DISCLAIMER
1. The library is NOT production ready at the moment. This is why the README of in the LibDeflate stills says "NOT working yet." However, I have tested the library through Travis CI with 100% code coverage, so it is at least testable by WeakAuras right now. The functionality and the API is still SUBJECT TO CHANGE. I plan to make a pull request to WeakAuras when I release the Beta version of Curseforge, but not now.

2. I think that using this library breaks all WeakAuras transmission compatibilty., including WeakAuras stuffs on WoWhead and Wago.io, but I think it's a perfect time to break it when the new expansion comes. The compressed data produced by LibDeflate is formatted according to [RFC1951](https://tools.ietf.org/html/rfc1951), and can be decompressed by the popular [Zlib](http://www.zlib.net/) library. Also the Zlib compressed data can be decompressed by LibDeflate, so I hope this won't cause much headache for website developers.

3. The library currently lacks functions similar to LibCompress:GetAddonEncodeTable():Encode(). I plan to implement my own version that does the same thing. But currently you can use LibCompress:GetAddonEncodeTable():Encode() on the LibDeflate's compressed data if you want to transmit the compressed data over the addon channel.
---
### Usage
LibDeflate has only one file "LibDeflate.lua". Load that file in-game, then
```
local LibDeflate = LibStub:GetLibrary("LibDeflate")

local compressed = LibDeflate:Compress(str)  -- compression, current default is level 5
local compressed = LibDeflate:Compress(str, 4)  -- optionally provides compression level, 1=fastest, 8=best compression.
[...]

PR #784, opened 2018-09-25
 implement LibDeflate

emptyrivers:
implement LibDeflate
From my testing, this is approximately a 80% decrease in the size of large string exports, and a further decrease in the size of addon channel transmissions.
This implementation is also backwards compatible with legacy strings, which were compressed and encoded using LibCompress, and older versions without LibDeflate will correctly detect that they cannot decode and decompress the data.
GitHub ticket #406

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 WeakAuras_Main.lua
 WeakAuras_Options.lua
 WeakAuras_Templates.lua
-WeakAuras/Libs/
+WeakAuras/Libs/*
+!WeakAuras/Libs/LibDeflate
 format.bat
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -18,7 +18,6 @@ externals:
  WeakAuras/Libs/LibButtonGlow-1.0: https://repos.wowace.com/wow/libbuttonglow-1-0
  WeakAuras/Libs/LibSpellRange-1.0: git://github.com/ascott18/LibSpellRange-1.0.git
  WeakAuras/Libs/LibRangeCheck-2.0: https://repos.wowace.com/wow/librangecheck-2-0/trunk/LibRangeCheck-2.0
- WeakAuras/Libs/LibDeflate: git://github.com/SafeteeWoW/LibDeflate.git
 
 enable-nolib-creation: no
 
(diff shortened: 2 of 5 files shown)


Stanzilla:
@oratory this will need changes to wago as a consequence

krazyito65:
Based on what the other thread said about the speed of compression, have you tired different levels of compression to test the speed and size?

He was saying even at lvl 8 it was too slow for the amount of compression you get

emptyrivers:
Yes, level 9 is quite slow compared to the lower levels, and in most applications I would not bother with it. But in this situation it makes sense. The difference in compression time between level 1 and level 9 is on the order of one second (about 100x slower) for very large compression jobs, for a size saving of about 40%, relative to the minimum level of compression. Compared to level 8, level 9 is (very roughly) twice as slow, with a relatively small difference in size (about 1-3%, maybe) So, in order to be worth it, the size savings needs to translate into on average a one second improvement on performance.

When sending through an addon channel, you effectively have a bandwidth of about 2kBps, due to all of the limitations imposed by blizzard. I estimate that in this case, there is no real difference between level 9 and level 8, since on the large end, level 9 compression will add a small amount of time to compression, while shaving off a comparable time from the transmission, due to a slightly smaller payload to transmit.

When encoding a printable string which is copied and pasted into e.g. https://wago.io, there is actually a massive slowdown (at least quadratic with the size of the string) from printing the text into an editbox. Why exactly this occurs is unclear, but it means that a 50k char string is rendered in a decent amount of time, a 100k char string has a noticeable delay, while a 150k char string takes over a minute to render, and anything larger than that will make the game appear to have crashed for minutes at a time. The time to serialize, compress, and encode the string is absolutely swamped by this slowdown, so that any decrease in size will ultimately be a boon for export time.

Essentially, blame blizzard for silly inefficiencies and tiny bandwidths ;)

InfusOnWoW:
for the record I did report the quadratic performance of the editbox to blizzard at some point....

And yeah, from my testing all the slowness is in actually pasting the text into the box. Compression time wasn't a big issue.

krazyito65:
I see, makes sense.

Issue #781, opened 2018-09-22
Request: Condition - Attackable Target


Kha-Khan:
**Is your feature request related to a problem? Please describe.**
I would like to have groups of auras that are visible only if you are either: In Combat or have an attackable target selected.  I can get this to work but it takes custom trigger logic a fair amount of the time.
I can get attackable target to work fine using Unit characteristics - target - attackable. I would like to be able to do that in conditions though

**Describe the solution you'd like**
I would like to see a new global condition for Attackable Target - True/False.  possibly replacing "has target"

**Describe alternatives you've considered**
I currently do this 2 different ways. If I only care about an always active trigger (Shield Slam for example). Then I have 2 other triggers, one for "in combat" and one for "Unit Characteristics - Target - Attackable" then I can simply have "Any trigger" required for activation and use current conditions. However, if I have some other logic beyond the one trigger, then I need to use custom trigger logic.

**Additional context**
While I can currently get Weakauras to do what I want, it takes several more steps than I think it needs to.  What matters more for me, since I share so many aura's, it is a pain to change for the end-user, if they don't like the way I did it.  For something that would be really pretty easy to change if I did it in conditions.

I know this doesn't get requested much but I do think quite a few people would actually use this condition if it existed.

InfusOnWoW:
This was implemented by @Vardex

PR #785, opened 2018-09-25
Added Condition:Attackable Target

Vardex:
Added the condition Attackable Target as requested in #781.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -878,6 +878,14 @@ local globalConditions =
       state.hastarget = UnitExists("target");
     end
   },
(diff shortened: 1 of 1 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -878,6 +878,14 @@ local globalConditions =
       state.hastarget = UnitExists("target");
     end
   },
+  ["attackabletarget"] = {
+    display = L["Attackable Target"]
+    type = "bool"
+    events = {"PLAYER_TARGET_CHANGED"},
```
InfusOnWoW: There must be a event for when a unit becomes attackable. Though I have no idea where to test that.

PR #799, opened 2018-09-30
rangecheck: handle group and multi-target

mrbuds:
Not Ready for merge

TODO: 
- throttling
- handle multi using internal list of nameplates made with NAMEPLATE_ADD event instead of iterate blindly
- fix display text tooltip
> %min and %max should only show with single target units
> i don't know if it should show %units since it's a table only usable for custom code?
- move %stacks to %s ? (i'm not sure how to do that)

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -4809,23 +4809,58 @@ WeakAuras.event_prototypes = {
     },
     name = L["Range Check"],
     init = function(trigger)
-      trigger.unit = trigger.unit or "target";
       local ret = [=[
+        -- TODO: handle throttle
+
           local unit = [[%s]];
-          local min, max = WeakAuras.GetRange(unit, true);
-          min = min or 0;
-          max = max or 999;
+          local range = %s;
+          local operator = [[%s]];
           local triggerResult = true;
+          local count = 0;
+          local names = "";
+          local units = {};
+
+          if unit=="multi" then
+            -- TODO or NOT TODO (that is the question) maintain a list of nameplates using events instead of iterate
+            -- increase 40 ?
+            for i = 1, 40 do
+              local u = "nameplate"..i
+              if UnitCanAttack("player", u)
+              and WeakAuras.CheckRange(u, range, operator)
+              then
+                count = count + 1;
+                units[count] = u;
+                names = names .. UnitName(u) .. "\n";
+              end
+            end
+            triggerResult = count > 0;
+          elseif unit=="group" then
+            for u in WA_IterateGroupMembers() do
+              if UnitIsVisible(u)
+              and not UnitIsPlayer(u)
+              and not UnitIsDeadOrGhost(u)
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
%s is a "legacy" text replacement that automtically takes state.stacks, except in a convulted and strange way. I don't know whether it's possible to hide %min %max, I would have to look into that code again.

You can implement throttling by:
* Making prototype.events a function, which depending on whether it's a simple one unit trigger or a multiple unit trigger retrurns a different event. So for single units make it FRAME_UPDATE, and for multiple make it RANGE_CHECK_UPDATE. (or make throttling a option in the trigger, and depending on which is selected, return the event.)

* Add a loadFunc (see other prototypes for those), that calls WeakAuras.InitRangeCheckUpdater()

* InitRangeCheckUpdater then creates a frame, that has a OnUpdate handler, and every 0.2s (or whatever) it sends out a RANGE_CHECK_UPDATE event.

* The problem with that is, that all range checking auras will then update in the same frame, so maybe that should be central then too.

mrbuds:
I added throttling with the OnUpdate method

There is still one thing missing from what i was doing with custom code, I like running the rangecheck trigger only when it is relevant, coupled with an other trigger(s) check if it's active before spaming rangecheck with `if WeakAuras.triggerState[aura_env.id].triggers[1] then`

But i don't believe this is the place to implement such function

mrbuds:
I changed my mind :D

Added option to run the trigger only when aura is active to save CPU.

If checked it will avoid to run a costly function every 0.2s when there is no need for it, requiring to have another trigger active and the activation function returning true.

option

The text description may need improvement.

WeakAuras/Prototypes.lua
```diff
@@ -4839,49 +4839,60 @@ WeakAuras.event_prototypes = {
           local count = 0;
           local names = "";
           local units = {};
+          local activation = %s;
+          activation = activation and WeakAuras.IsTriggerActive(aura_env.id) or not activation;
```
InfusOnWoW: That doesn't work at all. Triggers can't depend on other triggers.

mrbuds: Mhmm i'm not surprised of reaction.
I still really don't like to run the "group" and "multi" rangecheck for a full encounter.

What could be done is to leave it like that without my last dirty commit, but add a performance warning, and add rangechecks to conditions for triggers where it can be relevant, like aura, bw and dbm timers, and cast trigger.

Do you think this is an acceptable solution ?

mrbuds:
After what was told today on discord, i am convinced that this pull request is a bad idea.

Because the rangecheck trigger for group and multi takes too much resources to be run so frequently during an encounter, it is better to use custom code to be able to run the code only when needed.

I have started to make a group of templates on wago that could be added to a discord !rangecheck macro https://wago.io/RangeCheckTemplates

Stanzilla:
@mrbuds close then?

PR #804, opened 2018-10-02
Change Zoom property by Condition

Irame:
I like to have the zoom property set to 30% for non glowing icons but if they glow it looks better with 0% zoom, so I added an option to also set the zoom property of an icon via the condition tab.

I used the 'UpdateSize' function to apply the change, it seems a bit heavy but my goal was to keep the change as small as possible. I'm sure there is a better (less CPU heavy) way of doing this.

### DIFF
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -121,7 +121,8 @@ local properties = {
     min = 0,
     max = 1,
     step = 0.01,
-    default = 0
+    default = 0,
+    isPercent = true
   },
 };
 
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -475,20 +475,10 @@ local function modify(parent, region, data)
   end
 
   function region:UpdateSize()
-    local mirror_h, mirror_v, width, height;
-    local scalex = region.scalex;
-    local scaley = region.scaley;
-    if(scalex < 0) then
-      mirror_h = true;
-      scalex = scalex * -1;
-    end
-    width = region.width * scalex;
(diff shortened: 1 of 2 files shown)


WeakAuras/RegionTypes/Icon.lua
```diff
@@ -557,6 +566,11 @@ local function modify(parent, region, data)
   function region:SetInverse(inverse)
     cooldown:SetReverse(not inverse);
   end
+  
+  function region:SetZoom(zoom)
+    data.zoom = zoom;
```
InfusOnWoW: That modifies the setting in the actual aura settings.

What needs to happen is to introduce region.zoom, e.g. look at how keepAspectRatio is copied to the region and the Set method only sets that.

You probably should split up UpdateSize() into the parts that modify the widht/height and those that update the tex coordinates, so that you can call the later without the former code.

Irame: Thanks for the quick response! I will look into it and update it as soon as possible.

Irame:
The change to save the zoom in the 'region' table, was a small one.
I thought about splitting the 'UpdateSize' function and I didn't find a nice way to do it, because a lot of values that are calculated for the size are also used for the texture coordinates. So most of the 'UpdateSize' function had to be copied into a 'UpdateTexCoords' function or extracted into a 3rd function.

InfusOnWoW:
I don't agree that splitting up UpdateSize is complicated. 
After all, the only two values used for actually setting the size are width and height.

Everything else is just for setting the tex coords.

Irame:
You actually need the scaling factor as well to set the size and you need the calculated width and height again to set the aspect ratio, but I have split it now.

InfusOnWoW:
Thanks!

Issue #810, opened 2018-10-02
Aura Shift Selecting


Stanzilla:
**Is your feature request related to a problem? Please describe.**
Shift selection is quite common and we should probably support it

**Describe the solution you'd like**
Selecting an aura in a group and then shift clicking on a aura further down or up the list, should select all the auras in between as well.

**Describe alternatives you've considered**

**Additional context**

emptyrivers:
What about shift clicking between groups?

Stanzilla:
Would be a bit odd, no? I'd say only inside OR outside of groups

emptyrivers:
I ask because you can ctrl click to select children in different groups

Stanzilla:
Fixed with https://github.com/WeakAuras/WeakAuras2/pull/812

PR #812, opened 2018-10-04
multi-selection of auras with shift modifier

mrbuds:
It works only if the first aura selected and the second are both in the same group or are top level.

implement #810

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -1316,13 +1316,7 @@ function WeakAuras.ShowOptions(msg)
       for k,v in pairs(tempGroup.controlledChildren) do
         children[k] = v
       end
-      for index,childId in pairs(children) do
-        if (index == 1) then
-          WeakAuras.PickDisplay(childId);
-        else
-          WeakAuras.PickDisplayMultiple(childId);
-        end
-      end
+      frame:PickDisplayBatch(children);
     else
       WeakAuras.PickDisplay(frame.pickedDisplay);
     end
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -4070,7 +4070,7 @@ function WeakAuras.PickDisplayMultipleShift(target)
     -- get first aura selected
     local first;
     if (WeakAuras.IsPickedMultiple()) then
-      first = tempGroup.controlledChildren[1];
+      first = tempGroup.controlledChildren[#tempGroup.controlledChildren];
     else
       first = frame.pickedDisplay;
     end
@@ -4098,7 +4098,7 @@ function WeakAuras.PickDisplayMultipleShift(target)
               break;
             end
           end
-        else
+        elseif (firstData.parent == nil and targetData.parent == nil) then
           -- top-level
           for index, button in ipairs(frame.buttonsScroll.children) do
             local data = button.data;
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -4082,25 +4082,19 @@ function WeakAuras.PickDisplayMultipleShift(target)
       first = frame.pickedDisplay;
     end
     if (first and first ~= target) then
-      -- check that id is in same group as first
+      -- check if target and first are in same group and are not a group
       local firstData = WeakAuras.GetData(first);
-      -- in a group
-      if (firstData.parent) then
-        local sameGroup = false;
-        local group = WeakAuras.GetData(firstData.parent);
-        for index, child in pairs(group.controlledChildren) do
-          if (child == target) then
-            sameGroup = true;
-            break;
-          end
-        end
-        if (sameGroup) then
-          local batchSelection = {};
-          for index, child in pairs(group.controlledChildren) do
+      local targetData = WeakAuras.GetData(target);
+      if (firstData.parent == targetData.parent and not targetData.controlledChildren and not firstData.controlledChildren) then
+        local batchSelection = {};
+        -- in a group
+        if (firstData.parent) then
+          local group = WeakAuras.GetData(targetData.parent);
+          for index, child in ipairs(group.controlledChildren) do
             -- 1st button
             if (child == target or child == first) then
               table.insert(batchSelection, child);
-              for i=index+1,#group.controlledChildren do
+              for i = index + 1, #group.controlledChildren do
                 local current = group.controlledChildren[i];
                 table.insert(batchSelection, current);
                 -- last button: stop selection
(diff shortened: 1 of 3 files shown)


emptyrivers:
This seems to have some scaling issues. Performance drops off quickly, and when selecting a large number of auras (the primary use case i think), the UI can start to glitch out and not render DisplayButtons that it should be rendering.

mrbuds:
Indeed i confirm the problem.

It needs a batch version of frame.PickDisplayMultiple in OptionsFrame.lua

WeakAuras.PickDisplayMultiple() does for each aura:
WeakAuras.EnsureOptions(id);
WeakAuras.ReloadTriggerOptions(tempGroup);
self:FillOptions(displayOptions[tempGroup.id]);

mrbuds:
i'm dumb i did only in 1 part of the code...

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4072,84 @@ function WeakAuras.PickDisplayMultiple(id)
...
+      first = frame.pickedDisplay;
+    end
+    if (first and first ~= target) then
+      -- check that id is in same group as first
+      local firstData = WeakAuras.GetData(first);
+      -- in a group
+      if (firstData.parent) then
+        local sameGroup = false;
+        local group = WeakAuras.GetData(firstData.parent);
+        for index, child in pairs(group.controlledChildren) do
```
InfusOnWoW: You can use GetData(target) here and check .parent of that.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4072,84 @@ function WeakAuras.PickDisplayMultiple(id)
...
+          if #batchSelection > 0 then
+            frame:PickDisplayBatch(batchSelection);
+          end
+        end
+      else
+        -- 1st aura selected is top-level, check if target is also top-level and is not a group
+        local targetData = WeakAuras.GetData(target);
+        if (targetData and not targetData.controlledChildren and not targetData.parent) then
+          local batchSelection = {};
+          for index,button in pairs(frame.buttonsScroll.children) do
```
InfusOnWoW: I think this must be ipairs

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4072,84 @@ function WeakAuras.PickDisplayMultiple(id)
...
+        local group = WeakAuras.GetData(firstData.parent);
+        for index, child in pairs(group.controlledChildren) do
+          if (child == target) then
+            sameGroup = true;
+            break;
+          end
+        end
+        if (sameGroup) then
+          local batchSelection = {};
+          for index, child in pairs(group.controlledChildren) do
```
InfusOnWoW: This must be ipairs.

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW: Or well, it doesn't need to be, but it looks better if it is.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4072,84 @@ function WeakAuras.PickDisplayMultiple(id)
...
+            -- 1st button
+            if (child == target or child == first) then
+              table.insert(batchSelection, child);
+              for i=index+1,#group.controlledChildren do
```
InfusOnWoW: Missing spaces around = and after ,

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW: Also missing space after ,

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4072,84 @@ function WeakAuras.PickDisplayMultiple(id)
...
+        -- 1st aura selected is top-level, check if target is also top-level and is not a group
+        local targetData = WeakAuras.GetData(target);
+        if (targetData and not targetData.controlledChildren and not targetData.parent) then
+          local batchSelection = {};
+          for index,button in pairs(frame.buttonsScroll.children) do
+            local data = button.data;
+            -- 1st button
+            if (data and (data.id == target or data.id == first)) then
+              table.insert(batchSelection, data.id);
+              for i=index+1,#frame.buttonsScroll.children do
```
InfusOnWoW: Missing spaces

WeakAurasOptions/WeakAurasOptions.lua
mrbuds: I could do 
      local firstData = WeakAuras.GetData(first);
      local targetData = WeakAuras.GetData(target);
and then compare firstData.parent == targetData.parent (which is more readable and concise)
But after checking both items are in the same group, i still need to iterate over the group children, so another GetData() on the parent

The way i did it save a GetData() call, i have not checked the complexity of this call, do you think their is not enough performance gain doing what i did for what we loose in readability ?

InfusOnWoW: GetData is cheap, iterating over all childrens is not as cheap. So using that to figure out whether they are the same parent is probably better.

mrbuds:
fixed a bug, and "Duplicate All" and re-opening option with a multi-selection should be faster

mrbuds:
shiiiiit since last commit, when the right click on a multi-selection act as only 1 aura is selected

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4065,71 @@ function WeakAuras.PickDisplayMultiple(id)
   frame:PickDisplayMultiple(id);
 end
 
+function WeakAuras.PickDisplayMultipleShift(target)
+  if (frame.pickedDisplay) then
+    -- get first aura selected
+    local first;
+    if (WeakAuras.IsPickedMultiple()) then
+      first = tempGroup.controlledChildren[1];
```
InfusOnWoW: I think using the last one here is actually better. The controlledChildren of tempGroup are in the order in which they were picked, so using the last one uses the last one that was selected, which matches what's normal for shift selection

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -4072,6 +4065,71 @@ function WeakAuras.PickDisplayMultiple(id)
...
+                -- last button: stop selection
+                if (current == target or current == first) then
+                  break;
+                end
+              end
+              break;
+            end
+          end
+        else
+          -- top-level
```
InfusOnWoW: You might want to check for firstData.parent == nil and targetData.parent == nil here, to confirm that we are handling the case of both being top level. 

The code works without that, but the comment is wrong, the else also currently handles the case of mixed top-levelness

WeakAurasOptions/WeakAurasOptions.lua
mrbuds: Indeed, i wasn't thinking enough about how shift-selection works with some auras already selected with ctrl. I also need to change frame.PickDisplayBatch to not do any PickDisplay()

mrbuds:
Thanks Infus it's much better using the last element of previous selection
example
Just one comment before someone report it as a bug, as you can see it in this example, the shift-selection as it is implemented doesn't show previews of auras selected like left click or ctrl-click selection does.

Stanzilla:
```Just one comment before someone report it as a bug, as you can see it in this example, the shift-selection as it is implemented doesn't show previews of auras selected like left click or ctrl-click selection does.
```

I see this as a problem, they all should behave the same

mrbuds:
wait wait wait .. i said this after watching the (cropped) gif without actually checking in game, the preview of all auras is fine

PR #818, opened 2018-10-04
Tweak dynFrame timings

emptyrivers:
the current usage will let the frame be redrawn after it has been working for 16 milliseconds. But this doesn't imply 60 fps, since it will go frame update -> 16ms of work from weakauras -> next frame update.
So, instead, measure the time difference between last frame, and now. This will ensure that weakauras by itself will not consume too much cpu time per frame, though fps might still drop below 60 if another addon has work to do after weakauras finishes its threads.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -4010,14 +4010,9 @@ do
   dynFrame.frame:Hide();
   local start = debugprofilestop()
   dynFrame.frame:SetScript("OnUpdate", function(self, elapsed)
-    -- Start timing
-    local hasData = true;
-
+    local hasData = false
     -- Resume as often as possible (Limit to 16ms per frame -> 60 FPS)
-    while (debugprofilestop() - start < 16 and hasData) do
-      -- Stop loop without data
-      hasData = false;
-
+    repeat
       -- Resume all coroutines
       for name, func in pairs(dynFrame.update) do
         -- Loop has data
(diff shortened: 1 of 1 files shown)


emptyrivers:
Certainly, in my testing this is a noticeable improvement. The work is much smoother.

InfusOnWoW:
What's the problem that is trying to solve?

Stanzilla:
Yeah the commit and your text needs some more explaining. I tested it in game and it did not feel very different?

emptyrivers:
This is targeted at combating some stuttering that I get when importing very large groups. Essentially, the difference is between counting the time that WeakAuras is taking in its coroutines, and counting the time between one frame and the next.

What currently happens:
`|(0)Frame renders -> (1)various bits and bobs -> (2)WeakAuras coro -> (3)other bits and bobs -> |Next Frame -> ...`

And WeakAuras tries to not let (2) take longer than 16 milliseconds, but this doesn't guarantee that the whole frame takes no more than 16 milliseconds to finish. If instead we measure the time between now and the moment that WeakAuras last paused its coroutines (so, record a time at (3), and compare against that), then we implicitly also measure the various other things which are happening (other events or addons). In a busy environment with busy addons, this might help minimize the stuttering which can result from importing a very large group.

I won't cry if you decide not to make the change :)

InfusOnWoW:
Nothing in that code is optimized, so I'd rather look into what exactly we are doing there and why it is slow.

Issue #665, opened 2018-08-22
Change Glow Color


disc0infern0:
Can a region setting be added to allow the glow color to be specified? 

Changing the glow color could then be done programmatically to indicate different conditions

asakawa-k:
The glow used in WeakAuras comes from the Lib that utilises Blizzard elements that are yellow. 
Various different ways to "glow" Auras have been developed on the WA Discord so I recommend you go there to find/discuss your options.

disc0infern0:
Will do. Thankyou for replying so quickly.

InfusOnWoW:
I'm pretty sure the orignal wow glow that we use, only supports yellow. And that's the only glow we support atm, so that's not trivial to do.

Dooez:
Desaturating glow textures and setting vertex color allows for okay colouring

emptyrivers:
Oh, this has been implemented for a few weeks now.

PR #827, opened 2018-10-07
Better glow options

emptyrivers:
This implements a library written by Deezo, and adds two new glow effects, as well as the ability to change the color. I spoke with Deezo earlier today, and he said that he would look at adding the standard Overlay Glow (the same one that LibButtonGlow provides) to his library, and also (more importantly) provide a solution which can change the color of the standard Overlay Glow. That would take care of #665.


emptyrivers:
This change will add two new fields to the data of icon type auras, glowType and glowColor, two new methods to icon regions (region:SetGlowType and region:SetGlowColor), and (if used) up to two new fields to start and finish actions of auras (again, glow type and glow color).

Stanzilla:
I'd rather not have two glow libs in use, so if he can integrate the standard overlay glow and it performs on par with LibButtonGlow then ok

InfusOnWoW:
I'm not against this, apparently glow options are pretty popular for whatever reason.

emptyrivers:
The whole reason that I decided to go ahead and work on this was because people were starting to write drop-in snippets that they put into actions - on init to change the glow effect.

emptyrivers:
There's just a few kinks left to work out (namely, I'm trying to decide how to let the user decide not to use a color, since the solution for button overlay involves desaturating the textures), and this will be ready to go.

emptyrivers:
@Stanzilla the button glow implementation in LibCustomGlow is the same as in LibButtonGlow, with very minor changes which i don't think will affect performance.

Stanzilla:
Can you move every other place to WeakAuras.newFeatureString as well in the same commit please

Issue #817, opened 2018-10-04
Multi-target unit for status - cast trigger


Causese:
**Is your feature request related to a problem? Please describe.**
1) On council type fights, when there's multiple bossX units, sometimes boss 1/2 swaps right on pull.
Also duplicating a cast trigger for each bossX unit is also an annoying process. 

2) For non boss units aka nameplate units it would also be useful to have a multi-target unit option 
(for dungeons specifically, since "status - cast" auras hide if the unit's cast is interrupted. An event - cleu trigger wouldn't hide for example)

**Describe the solution you'd like**
basically a multi-target unit for status - cast (unit)

**Additional context**
CLEU triggers used to be popular for this but imo status - cast is the new meta because you don't have to worry about setting the right timed duration etc since it's all done by blizzard (unitcastinginfo)
In raids the cast time often changes based on raid difficulty, real horror for event - cleu trigger.

InfusOnWoW:
This is now implemented for the next release with a "Multi" option for unit. There's also a option to check the Caster's target, which dynamically updates wheneve the caster's target changes.

PR #857, opened 2018-10-13
cast trigger: slight rework, add multi, auto-clone and target options

mrbuds:
ready for review :)

solve #817 
also target check in trigger and condition
this can be use for all abilities with late targeting like zek'voz eye beam or zul's shadow barrage

TODO: 
- ~~auto-clone break conditions, ex: condition target=player work only once, then need to target someone else before it work again~~ #875 fixed it
- ~~target condition can't use "specific unit", workaround or remove it from the list ?~~ fixed
- double check that there are no useless events in the list

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -2640,15 +2640,15 @@ end
 do
   local scheduled_scans = {};
 
-  local function doCastScan(fireTime)
-    WeakAuras.debug("Performing cast scan at "..fireTime.." ("..GetTime()..")");
-    scheduled_scans[fireTime] = nil;
-    WeakAuras.ScanEvents("CAST_REMAINING_CHECK");
+  local function doCastScan(args)
+    WeakAuras.debug("Performing cast scan at "..args[1].." ("..GetTime()..")");
+    scheduled_scans[args[1]] = nil;
+    WeakAuras.ScanEvents("CAST_REMAINING_CHECK", args[2]);
   end
-  function WeakAuras.ScheduleCastCheck(fireTime)
+  function WeakAuras.ScheduleCastCheck(fireTime, unit)
     if not(scheduled_scans[fireTime]) then
       WeakAuras.debug("Scheduled cast scan at "..fireTime);
-      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, fireTime);
+      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, {fireTime, unit});
     end
   end
 end
@@ -2983,7 +2983,7 @@ function GenericTrigger.GetTriggerConditions(data, triggernum)
               display = v.display,
               type = v.conditionType
             }
-            if (result[v.name].type == "select") then
+            if (result[v.name].type == "select" or result[v.name].type == "unit") then
               if (v.conditionValues) then
                 result[v.name].values = WeakAuras[v.conditionValues];
               else
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -2640,15 +2640,15 @@ end
 do
   local scheduled_scans = {};
 
-  local function doCastScan(args)
-    WeakAuras.debug("Performing cast scan at "..args[1].." ("..GetTime()..")");
-    scheduled_scans[args[1]] = nil;
-    WeakAuras.ScanEvents("CAST_REMAINING_CHECK", args[2]);
+  local function doCastScan(firetime, unit)
+    WeakAuras.debug("Performing cast scan at "..firetime.." ("..GetTime()..")");
+    scheduled_scans[firetime] = nil;
+    WeakAuras.ScanEvents("CAST_REMAINING_CHECK", unit);
   end
   function WeakAuras.ScheduleCastCheck(fireTime, unit)
     if not(scheduled_scans[fireTime]) then
       WeakAuras.debug("Scheduled cast scan at "..fireTime);
-      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, {fireTime, unit});
+      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, fireTime, unit);
     end
   end
 end
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -1003,7 +1003,7 @@ WeakAuras.load_prototype = {
 };
 
 local function AddUnitChangeEvents(unit, t)
-  if (unit == "player") then
+  if (unit == "player" or unit == "multi") then
 
   elseif (unit == "target") then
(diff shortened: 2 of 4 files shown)


mrbuds:
Last commit add new conditionType="unit", it works the same as "select" but if the value of the check is not in the "values" list it will be considered as "specific unit" (member), and will show a new line with input box.
This conditionType may as well be useful for some other trigger that have optional unit field.

InfusOnWoW:
Revased it and removed all the changes in Types.lua that were accidentally included.

WeakAuras/Prototypes.lua
```diff
@@ -4393,159 +4393,246 @@ WeakAuras.event_prototypes = {
...
+        "UNIT_SPELLCAST_STOP",
+        "UNIT_SPELLCAST_CHANNEL_STOP",
+        "UNIT_SPELLCAST_INTERRUPTED",
+        "UNIT_SPELLCAST_FAILED",
+        "UNIT_SPELLCAST_FAILED_QUIET",
       };
-      AddUnitChangeEvents(trigger.unit, result);
-      return result;
+      AddUnitChangeEvents(trigger.unit, result)
+      if trigger.target ~= "" then
```
InfusOnWoW: I don't undertand this code. What is trigger.target?

WeakAuras/Prototypes.lua
```diff
@@ -4393,159 +4393,246 @@ WeakAuras.event_prototypes = {
...
-      ret = ret:format(trigger.unit, trigger.use_inverse and "true" or "false");
 
-      if(trigger.use_remaining) then
-        local ret2 = [[
-          local expirationTime = endTime and endTime > 0 and (endTime / 1000) or 0;
-          local remaining = expirationTime - GetTime();
-          local remainingCheck = %s;
-          if(remaining >= remainingCheck and remaining > 0) then
-            WeakAuras.ScheduleCastCheck(expirationTime - remainingCheck);
+          if event == "PLAYER_TARGET_CHANGED" then
```
InfusOnWoW: I don't understand this bit, can you explain?

WeakAuras/GenericTrigger.lua
```diff
@@ -2640,15 +2640,15 @@ end
...
+    WeakAuras.debug("Performing cast scan at "..args[1].." ("..GetTime()..")");
+    scheduled_scans[args[1]] = nil;
+    WeakAuras.ScanEvents("CAST_REMAINING_CHECK", args[2]);
   end
-  function WeakAuras.ScheduleCastCheck(fireTime)
+  function WeakAuras.ScheduleCastCheck(fireTime, unit)
     if not(scheduled_scans[fireTime]) then
       WeakAuras.debug("Scheduled cast scan at "..fireTime);
-      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, fireTime);
+      scheduled_scans[fireTime] = timer:ScheduleTimerFixed(doCastScan, fireTime - GetTime() + 0.1, {fireTime, unit});
```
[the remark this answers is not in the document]
mrbuds: fixed locally

WeakAuras/Prototypes.lua
mrbuds: trigger.target reference a field i added that let you specify who is the target of the cast.
This give the possibility to use the generic cast trigger for example to show any cast with you as the target (with unit=multi and target=player) or when Zek'voz eye target's you (unit=boss2 and target=player)
There are quite a few npc spells that doesn't target the right destination when they start casting, and change target during the cast, so in this part of the code i added the event "UNIT_TARGET" for checking the trigger (in uldir the zkvoz' eye beam and zul's shadow barrage for ex)

InfusOnWoW: Ah, hmm, but wouldn't they untrigger as soon as the boss selects a different target?

I'm a bit unsure if the feature will work good enough to not require tons of explanation on why it doesn't work "correctly"

WeakAuras/Prototypes.lua
mrbuds: "PLAYER_TARGET_CHANGED" is registered by AddUnitChangeEvents() for trigger.unit == "target" or with specific unit, for specific unit it is useless in most case except for "target-xxxx", so it might be preferable to register this event only for "target".

When this event trigger, sourceUnit is always nil and the unit that needs to be checked is trigger.unit

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -1143,18 +1143,49 @@ local function addControlsForIfLine(args, order, data, conditionVariable, condit
...
-        order = order + 1;
+        if (currentConditionTemplate.type == "unit") then
+          args["condition" .. i .. tostring(path) .. "_value"] = {
+            type = "select",
+            name = blueIfNoValue(data, conditions[i].check, "value", L["Differences"]),
+            desc = descIfNoValue(data, conditions[i].check, "value", currentConditionTemplate.type),
+            order = order,
+            values = currentConditionTemplate.values,
+            get = function()
+              return currentConditionTemplate.values[check.value] and check.value or "member"
```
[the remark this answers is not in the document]
mrbuds: fixed locally

WeakAuras/Prototypes.lua
```diff
@@ -4393,159 +4393,246 @@ WeakAuras.event_prototypes = {
...
+        enable = function(trigger) return not trigger.use_inverse end
+      },
+      {
+        name = "sourceName",
+        display = L["Caster Name"],
+        store = true,
+        hidden = true
+      },
+      {
+        name = "destUnit",
```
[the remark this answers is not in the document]
mrbuds: Yes indeed this option is obviously not good for pvp, and will not always be correct in pve.
But i think there are enough situations in pve where this is useful and the event UNIT_TARGET make it more accurate. This option will let creators of dungeons/raids packs avoid using (sometime bad) custom code.

WeakAuras/Prototypes.lua
mrbuds: > wouldn't they untrigger as soon as the boss selects a different target?

Yes it will untrigger if boss change target, or trigger after the cast has started if trigger.target is not the first target during the cast, but that's enough to run a sound alert on show or whatever, and it will always show an accurate timer for the end of the cast.

For eyebeam, bigwigs register event "UNIT_TARGET" at the start of the cast and run it's "printTarget()" on next UNIT_TARGET event with the right source. It works for this case but is not generic enough.
For DBM after SPELL_CAST_START it check target every 0.1s

WeakAuras/Types.lua
```diff
@@ -155,6 +155,15 @@ WeakAuras.actual_unit_types_with_specific = {
   member = L["Specific Unit"]
 }
 
+WeakAuras.actual_unit_types_with_specific_and_multi = {
+  player = L["Player"],
+  target = L["Target"],
+  focus = L["Focus"],
+  pet = L["Pet"],
+  member = L["Specific Unit"],
+  multi = L["Multi-target (CLEU)"]
```
emptyrivers: Enough people don't know what CLEU means, that I'd rather write Combat Log in the parenthesis here. We don't use CLEU anywhere else in the codebase, or in the UI phrases.

InfusOnWoW: It's not really using the Combat Log either, so it shouldn't even have that.

PR #866, opened 2018-10-16
hide cooldown swipe & show edge on icons with stacks

lotheac:
Hi, I'd like icons with stacks (eg. spells with charges) to display like the default actionbars show them: the cooldown swipe is visible if there are no stacks remaining, and the swipe edge is visible if the spell is recharging with stacks remaining.

This is a quick and dirty stab at doing that; it might not be desirable for all icons to behave this way.

### DIFF
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -45,6 +45,8 @@ local default = {
   glowColor = {1, 1, 1, 1},
   glowType = "buttonOverlay",
   cooldownTextEnabled = true,
+  cooldownSwipe = true,
+  cooldownEdge = false,
 };
 
 WeakAuras.regionPrototype.AddAlphaToDefault(default);
@@ -133,6 +135,16 @@ local properties = {
     display = L["Inverse"],
     setter = "SetInverse",
     type = "bool"
+  },
+  cooldownSwipe = {
+    display = WeakAuras.newFeatureString .. L["Cooldown Swipe"],
+    setter = "SetCooldownSwipe",
+    type = "bool",
+  },
+  cooldownEdge = {
+    display = WeakAuras.newFeatureString .. L["Cooldown Edge"],
+    setter = "SetCooldownEdge",
+    type = "bool",
   },
   zoom = {
     display = L["Zoom"],
@@ -223,7 +235,6 @@ local function create(parent, data)
   local cooldown = CreateFrame("COOLDOWN", "WeakAurasCooldown"..frameId, region, "CooldownFrameTemplate");
   region.cooldown = cooldown;
   cooldown:SetAllPoints(icon);
-  cooldown:SetDrawEdge(false);
 
   local stacksFrame = CreateFrame("frame", nil, region);
   local stacks = stacksFrame:CreateFontString(nil, "OVERLAY");
@@ -598,6 +609,19 @@ local function modify(parent, region, data)
       cooldown:SetCooldown(cooldown.expirationTime - cooldown.duration, cooldown.duration);
     end
   end
+
+  function region:SetCooldownSwipe(cooldownSwipe)
+    region.cooldownSwipe = cooldownSwipe;
+    cooldown:SetDrawSwipe(cooldownSwipe);
+  end
+
+  function region:SetCooldownEdge(cooldownEdge)
+    region.cooldownEdge = cooldownEdge;
+    cooldown:SetDrawEdge(cooldownEdge);
+  end
+
+  region:SetCooldownSwipe(data.cooldownSwipe)
+  region:SetCooldownEdge(data.cooldownEdge)
 
   function region:SetZoom(zoom)
     region.zoom = zoom;
--- a/WeakAurasOptions/RegionOptions/Icon.lua
+++ b/WeakAurasOptions/RegionOptions/Icon.lua
@@ -3,48 +3,16 @@ local L = WeakAuras.L
 
 local function createOptions(id, data)
   local options = {
-    cooldown = {
-      type = "toggle",
-      name = L["Cooldown"],
-      order = 4,
-      disabled = function() return not WeakAuras.CanHaveDuration(data); end,
-      get = function() return WeakAuras.CanHaveDuration(data) and data.cooldown; end
-    },
-    inverse = {
-      type = "toggle",
-      name = L["Inverse"],
-      order = 6,
-      disabled = function() return not (WeakAuras.CanHaveDuration(data) and data.cooldown); end,
-      get = function() return data.inverse and WeakAuras.CanHaveDuration(data) and data.cooldown; end,
-      hidden = function() return not data.cooldown end
-    },
-    cooldownTextEnabled = {
-      type = "toggle",
-      name = L["Show Cooldown Text"],
-      order = 6.5,
-      disabled = function() return not WeakAuras.CanHaveDuration(data); end,
-      hidden = function() return not (data.cooldown and not IsAddOnLoaded("OmniCC") and GetCVar("countdownForCooldowns") == "1") end,
-      width = "double"
-    },
-    cooldownSpace = {
-      type = "description",
-      name = "",
-      order = 6.6,
-      width = "normal",
-      hidden = function()
-        return data.cooldown and (IsAddOnLoaded("OmniCC") or GetCVar("countdownForCooldowns") ~= "1");
-      end
-    },
     color = {
       type = "color",
       name = L["Color"],
       hasAlpha = true,
-      order = 7
+      order = 1
     },
     auto = {
       type = "toggle",
       name = L["Automatic Icon"],
-      order = 8,
+      order = 2,
       disabled = function() return not WeakAuras.CanHaveAuto(data); end,
       get = function() return WeakAuras.CanHaveAuto(data) and data.auto; end
     },
@@ -52,7 +20,7 @@ local function createOptions(id, data)
       type = "input",
       name = L["Display Icon"],
       hidden = function() return WeakAuras.CanHaveAuto(data) and data.auto; end,
-      order = 12,
+      order = 3,
       get = function()
         return data.displayIcon and tostring(data.displayIcon) or "";
       end,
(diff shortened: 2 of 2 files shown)


emptyrivers:
No, I don’t think this is a good approach. A more elegant solution would be to introduce some options to control the behavior of the swipe.

asakawa-k:
Not that I would discourage a built-in option for this, but it is relatively simple to make an Aura work like this with a Condition and small snippet of code. 

`IF - trigger# - charges > 0 THEN - run custom code`:
```
aura_env.region.cooldown:SetDrawEdge(true)
aura_env.region.cooldown:SetSwipeColor(0,0,0,0)
```
Reversing those settings on another conditions with `charges = 0`

lotheac:
rewrote commit to add option instead.

emptyrivers:
Well, that wasn't really what I had in mind. That's almost too specific of an option. I'd rather expand the cooldown swipe options to be controllable via conditions. So you'd have some region methods along the lines of `region:SetCooldownSwipe(true)` to enable/disable the cooldown swipe, and then let the user specify under exactly which conditions that they want to see a cooldown swipe.

lotheac:
On Tue, Oct 16 2018 15:18:31 -0700, Allen Leigh Faure wrote:
> Well, that wasn't really what I had in mind. That's almost too specific of an option.

I dunno, it makes sense to me to be able to easily replicate how the
default actionbars draw cooldowns on abilities with charges. I don't
think it's too specific (but it *is* badly named, I didn't really know
what to call it)

> I'd rather expand the cooldown swipe options to be controllable via conditions. So you'd have some region methods along the lines of `region:SetCooldownSwipe(true)` to enable/disable the cooldown swipe, and then let the user specify under exactly which conditions that they want to see a cooldown swipe.

This almost sounds like it would be *more* difficult for the user to get
it working "right" (ie. how the default actionbars look) than it is
with custom code like asakawa-k mentioned above :)

-- 
Lauri Tirkkonen | lotheac @ IRCnet

lotheac:
On Wed, Oct 17 2018 06:33:07 +0300, Lauri Tirkkonen wrote:
> This almost sounds like it would be *more* difficult for the user to get
> it working "right" (ie. how the default actionbars look)

not that I got it completely right yet either; the swipe needs to be
there for the global cooldown (if enabled) but the edge should show
charge progress at the same time. maybe I need to consult Blizzard's
interface code on this.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

lotheac:
On Tue, Oct 16 2018 21:20:59 -0700, Allen Leigh Faure wrote:
> Mimicking Blizzard isn't enough of a reason to add an option which works for exactly one use case. If all we did was mimic Blizzard, then WeakAuras wouldn't be half as flexible as it is today.

sure, I wasn't exactly suggesting that... I mentioned Blizzard's code
because it seemed like they must have not one but two cooldown frames
(one for the regular cooldown, and another for the recharge) to be able
to show both the gcd and the recharge spark in different positions. And
I just checked, they do - so that means in order to accomplish what I
want to do here, I also need a new frame. That in itself changes how
this will end up looking, new options or not.

I'll look into this again later, gonna be a bit busy the rest of the
week. I welcome any insight you have into the matter (I'm a bit of a WA
novice; migrating from TMW since they started spending 5ms every frame
even with 0 displays configured...)

-- 
Lauri Tirkkonen | lotheac @ IRCnet

Dooez:
Charge swipe is different to cooldown swipe on the default action bars. If you have GCD and >1 charge both are shown.

emptyrivers:
Hi @lotheac, did you intend to continue working on this?

lotheac:
@emptyrivers yes, but I haven't gotten around to it yet...

emptyrivers:
Can you give me a time frame?

lotheac:
I'll take a quick stab at it now.

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -599,6 +614,14 @@ local function modify(parent, region, data)
     end
   end
 
+  function region:SetCooldownSwipe(drawswipe)
```
emptyrivers:
```suggestion
  function region:SetCooldownSwipe(cooldownSwipe)
    region.cooldownSwipe = cooldownSwipe
```

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -599,6 +614,14 @@ local function modify(parent, region, data)
...
+    cooldown:SetDrawSwipe(drawswipe);
```
emptyrivers:
```suggestion
    cooldown:SetDrawSwipe(cooldownSwipe);
```

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -599,6 +614,14 @@ local function modify(parent, region, data)
...
+  end
+
+  function region:SetCooldownEdge(drawedge)
```
emptyrivers:
```suggestion
  function region:SetCooldownEdge(cooldownEdge)
    region.cooldownEdge = cooldownEdge
```

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -599,6 +614,14 @@ local function modify(parent, region, data)
...
+    cooldown:SetDrawEdge(drawedge);
```
emptyrivers:
```suggestion
    cooldown:SetDrawEdge(cooldownEdge);
```

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -365,6 +378,8 @@ local function modify(parent, region, data)
 
   cooldown:SetReverse(not data.inverse);
   cooldown:SetHideCountdownNumbers(not data.cooldownTextEnabled or IsAddOnLoaded("OmniCC") or false);
+  cooldown:SetDrawSwipe(data.cooldownSwipe)
```
emptyrivers: Instead of calling SetDrawSwipe here directly, can you use the region:SetCooldownSwipe method you wrote?

emptyrivers:
Oh, and before I forget: You based your changes off of your master branch. This is fine, but understand that we don't do Plain Jane merges, so when this gets merged (probably via a squash), your master branch will diverge from ours. Just fetch & do a hard reset after the merge is done, and you should be good to go.

emptyrivers:
Also, on the GCD spinner:

We have designs on this, but they won't be realized for some time. Use this as a drop-in to custom init action in the meantime:
```lua
local region = aura_env.region
if not region.GCD then
    local GCD = CreateFrame("COOLDOWN", nil, region, "CooldownFrameTemplate");
    region.GCD = GCD
end
region.GCD:SetDrawEdge(true)
region.GCD:SetFrameLevel(region.cooldown:GetFrameLevel())
region.GCD:SetHideCountdownNumbers(true)
region.GCD:SetSwipeColor(1,1,1,0)
region.GCD:SetDrawBling(false)
region.GCD:Hide()

local GCD = function()
    local dur, exp = WeakAuras.GetGCDInfo()
    if dur then
        local start = region.GCD:GetCooldownTimes()
        if start and start ~= exp-dur then
            region.GCD:SetCooldown(exp-dur, dur)
            region.GCD:Show()
        end
    else
        region.GCD:Hide()
    end
end
region.GCD:RegisterEvent("SPELL_UPDATE_COOLDOWN");
region.GCD:SetScript("OnEvent", GCD)
GCD()```

File: WeakAuras/RegionTypes/Icon.lua (hunk 1 above)
Comment (lotheac) [reply WeakAuras/RegionTypes/Icon.lua]: why do you want to set region.cooldownSwipe here? region:SetInverse is analogous to this function but that doesn't change its corresponding value either.

File: WeakAuras/RegionTypes/Icon.lua (hunk 5 above)
Comment (lotheac) [reply WeakAuras/RegionTypes/Icon.lua]: i could, but that would be inconsistent with the above code that doesn't use SetInverse either.

**Comment by lotheac** [comment]
On Sat, Nov 10 2018 14:19:54 -0800, Allen Leigh Faure wrote:
> Also, on the GCD spinner:
> 
> We have designs on this, but they won't be realized for some time. Use this as a drop-in to custom init action in the meantime:
> ```lua
> local region = aura_env.region
> if not region.GCD then
>     local GCD = CreateFrame("COOLDOWN", nil, region, "CooldownFrameTemplate");
>     region.GCD = GCD
> end
> region.GCD:SetDrawEdge(true)
> region.GCD:SetFrameLevel(region.cooldown:GetFrameLevel())
> region.GCD:SetHideCountdownNumbers(true)
> region.GCD:SetSwipeColor(1,1,1,0)
> region.GCD:SetDrawBling(false)
> region.GCD:Hide()

this should actually SetDrawEdge(false) and SetDrawSwipe(true) instead
of drawing the edge and setting the swipe alpha to 0, but I get the
idea. thanks for the snippet.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

File: WeakAuras/RegionTypes/Icon.lua (hunk 5 above)
Comment (emptyrivers) [reply WeakAuras/RegionTypes/Icon.lua]: Well, the region methods are inconsistent with themselves about this throughout the icon region. I’d like to move towards constructing the region methods, and then using them for initial setup. I think this will lead to greater consistency, and fewer opportunities for unexpected bugs to occur because some housekeeping was forgotten – just rely on the method that you just constructed to do the housekeeping.

File: WeakAuras/RegionTypes/Icon.lua (hunk 1 above)
Comment (emptyrivers) [reply WeakAuras/RegionTypes/Icon.lua]: For consistency's sake, I would like to move towards all of the region methods setting the values they get to the region (and also maybe validating them). I understand that SetInverse doesn't do this, but it's the direction I want to take this. Note the glow methods, color, coloranim, scale, width & height, etc. Not every method does this appropriately, but I believe that it should.

File: WeakAuras/RegionTypes/Icon.lua (hunk 5 above)
Comment (emptyrivers) [reply WeakAuras/RegionTypes/Icon.lua]: So, to be clear, the pattern I want is:

1. make region method
2. call region method with value from data
3. conditions call region methods with the appropriate values to update the region

**Comment by lotheac** [comment]
updated to address the rest of your requests.

File: WeakAuras/RegionTypes/Icon.lua [hunk 6]
```diff
@@ -223,7 +235,8 @@ local function create(parent, data)
   local cooldown = CreateFrame("COOLDOWN", "WeakAurasCooldown"..frameId, region, "CooldownFrameTemplate");
   region.cooldown = cooldown;
   cooldown:SetAllPoints(icon);
-  cooldown:SetDrawEdge(false);
+  region:SetCooldownEdge(data.cooldownEdge);
+  region:SetCooldownSwipe(data.cooldownSwipe);
```
lotheac: but actually this doesn't work -- the region methods are defined in modify() and not callable here.

lotheac: changed create() back to using SetDrawEdge/SetDrawSwipe for this reason.

if you want to move the region function definitions outside of modify(), that's probably best done separately from this pr...

emptyrivers: Oh, I'm an idiot. I didn't notice that was the create method.

emptyrivers: So, yeah, it kind of would be nice to change how exactly modify() works. But that's definitely beyond the scope of this PR. 😂

lotheac:
On Sat, Nov 10 2018 18:37:55 -0800, Allen Leigh Faure wrote:
> > @@ -223,7 +235,8 @@ local function create(parent, data)
>    local cooldown = CreateFrame("COOLDOWN", "WeakAurasCooldown"..frameId, region, "CooldownFrameTemplate");
>    region.cooldown = cooldown;
>    cooldown:SetAllPoints(icon);
> -  cooldown:SetDrawEdge(false);
> +  region:SetCooldownEdge(data.cooldownEdge);
> +  region:SetCooldownSwipe(data.cooldownSwipe);
> 
> Oh, I'm an idiot. I didn't notice that was the create method.

this code was actually in both create and modify. modify() uses the
methods it defines now.

-- 
Lauri Tirkkonen | lotheac @ IRCnet

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -223,7 +235,8 @@ local function create(parent, data)
   local cooldown = CreateFrame("COOLDOWN", "WeakAurasCooldown"..frameId, region, "CooldownFrameTemplate");
   region.cooldown = cooldown;
   cooldown:SetAllPoints(icon);
-  cooldown:SetDrawEdge(false);
+  cooldown:SetDrawEdge(data.cooldownEdge);
+  cooldown:SetDrawSwipe(data.cooldownSwipe);
```
InfusOnWoW: That doesn't belong in create, that should only be done in modify.

WeakAuras/RegionTypes/Icon.lua
InfusOnWoW: Those functions only do that, because they check if the values changed and do a early return based on that. If there's no early return I don't see a particular value in having the values in region.

Stanzilla:
Thanks for your contribution!

Issue #687, opened 2018-08-29
Dynamic Group - Arcs


vraolji:
An "Arc" option for a Dynamic group. Where one can define the angle of the arc and length.

Currently using clockwise/counterclockwise and radius, but these assume a full circle. In order to clump elements along an edge of the "circle", blank placeholders need to be made to fill the circle and force the visible objects in place.

vraolji:
Or another option would be to add a "Arc/Bend" option to a horizontal or vertical group, similar to "Stagger" but would evenly spread them in a arc by the defined number.

emptyrivers:
This is now available in the latest alpha.

PR #872, opened 2018-10-18
Dynamic group rewrite

emptyrivers:
This is not yet ready to merge.

### Things to add

- [x] Fix a bug where canceling a Delete confirmation on a dynamic group could leave it permanently suspended.
- [x] Fix a bug where children with state indices of disparate types would cause the sorting to error out.
- [x] Implement [WoWAce 705](https://www.wowace.com/projects/weakauras-2/issues/705)
- [x] Implement [WoWAce 857](https://www.wowace.com/projects/weakauras-2/issues/857)
- [x] Rewrite how children of dynamic groups are sorted, to use an incremental insertion sort that's maintained only on active (visible) children. This could also allow custom sort scripts very easily (though the precise details on that I haven't thought through yet).
- [x] Rewrite how children of dynamic groups are positioned. Positioning code is untangled so that the exact grow type doesn't have to be checked dozens of time, and future additions are quite simple.
- [ ] Implement #687, by adding a start & end angle option to circle & countercircle grow types.
- [ ] Add new feature indicators where necessary
- [ ] Test literally everything
### Regressions to fix:
- [ ] DoResize is positioning the MoverResizer in the wrong position.
- [ ] Editing the grow type can cause positioning to break
- [ ] Grid and custom grow types can cause modifyThumbnail to error out. Likely, that function needs to be untangled as well.

### DIFF
--- a/WeakAuras/RegionTypes/DynamicGroup.lua
+++ b/WeakAuras/RegionTypes/DynamicGroup.lua
@@ -48,56 +48,74 @@ function WeakAuras.GetPolarCoordinates(x, y, originX, originY)
   return r, theta;
 end
 
+local selfPoints = {
+  default = "CENTER",
+  RIGHT = function(data)
+    if data.align  == "LEFT" then
+      return "TOPLEFT"
+    elseif data.align == "RIGHT" then
+      return "BOTTOMLEFT"
+    else
+      return "LEFT"
+    end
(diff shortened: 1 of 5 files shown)


Stanzilla:
From a feature request text file I just found: custom angles for dynamic group growth (use-case half-circles)

PR #905, opened 2018-10-29
add minimap icon

mrbuds:
tooltip

note that it add a new dependency for LibDBIcon-1.0, i have not edited packaging script

### DIFF
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -18,6 +18,7 @@ externals:
  WeakAuras/Libs/LibSpellRange-1.0: git://github.com/ascott18/LibSpellRange-1.0.git
  WeakAuras/Libs/LibRangeCheck-2.0: https://repos.wowace.com/wow/librangecheck-2-0/trunk/LibRangeCheck-2.0
  WeakAuras/Libs/LibCustomGlow-1.0: https://repos.curseforge.com/wow/libcustomglow
+ WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk/LibDBIcon-1.0
 
 enable-nolib-creation: no
 
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -30,6 +30,7 @@ function WeakAurasTimers:ScheduleTimerFixed(func, delay, ...)
 end
 
 local LDB = LibStub:GetLibrary("LibDataBroker-1.1")
+local LDBIcon = LibStub("LibDBIcon-1.0")
 local LCG = LibStub("LibCustomGlow-1.0")
 local timer = WeakAurasTimers
 WeakAuras.timer = timer
@@ -83,9 +84,22 @@ function SlashCmdList.WEAKAURAS(msg)
     elseif(msg == "pprint") then
       WeakAuras.PrintProfile();
       return;
+    elseif(msg == "minimap") then
+      WeakAuras.ToggleMinimap();
+      return;
     end
   end
   WeakAuras.OpenOptions(msg);
+end
+
+function WeakAuras.ToggleMinimap()
+  WeakAurasSaved.minimap.hide = not WeakAurasSaved.minimap.hide
+  if WeakAurasSaved.minimap.hide then
+    LDBIcon:Hide("WeakAuras");
+    prettyPrint(L["Use /wa minimap to show the minimap icon again"])
+  else
+    LDBIcon:Show("WeakAuras");
+  end
 end
 
 BINDING_HEADER_WEAKAURAS = ADDON_NAME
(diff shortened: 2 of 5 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -30,6 +30,7 @@ function WeakAurasTimers:ScheduleTimerFixed(func, delay, ...)
 end
 
 local LDB = LibStub:GetLibrary("LibDataBroker-1.1")
+local LDBIcon = LibStub("LibDBIcon-1.0")
```
InfusOnWoW: Don't we need to bundle something to enable that library?

mrbuds: Yes we need to bundle the library to make it work, btw here i used LibStub("LibDBIcon-1.0") instead of  local LDB = LibStub:GetLibrary("LibDBIcon-1.0") i have no idea if there is any difference, but i will probably update this.

WeakAuras/WeakAuras.lua
```diff
@@ -1159,6 +1168,107 @@ function WeakAuras.CreatePvPTalentCache()
...
+      if(IsShiftKeyDown()) then
+        if not(WeakAuras.IsOptionsOpen()) then
+          WeakAuras.Toggle();
+        end
+      else
+        WeakAuras.OpenOptions();
+      end
+    elseif button == 'MiddleButton' then
+      WeakAurasSaved.minimap.hide = true
+      LDBIcon:Hide("WeakAuras");
```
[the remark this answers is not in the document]
mrbuds: This work with LibDBIcon only, i had not tested it with another libbroker addon and i see that the tooltip can be confusing.
I don't know what is better, completely remove the mid-click function, or just removed the text from the tooltip, or leave it like that.

mrbuds:
Previous code gave to LibDataBroker data to be use by any LibDataBroker addon like fubar, titanpanel or bazooka (i didn't know about this one), they manage the wa icon on their own.
If you don't have one of these addons there is no "wa" icon anywhere.

LibDBIcon is a LibDataBroker addon on it's own, that can be bundled with addons to show icon on minimap.

WeakAuras/WeakAuras.lua
InfusOnWoW: Ah take a look at commits to https://github.com/WeakAuras/WeakAuras2/commits/master/.pkgmeta on how to tell the packager to bundle libaries.

InfusOnWoW:
And there's no way to show the correct tooltip depending on where exactly the button is?

mrbuds:
Not that i know, as the documentation explain https://github.com/tekkub/libdatabroker-1-1/wiki/data-specifications the LDB object we define is like a "feed" available for display addons, they all get the same tooltip, OnClick function, icon, label etc..

The code i use for hiding the icon with middle click is specific to LibDBIcon.

All the addons i use don't have a OnClick function to hide the icon from minimap, and move this function either in their UI or in Interface > Addons panel (and with /slash command)

For this case it would be ideal to have a WeakAuras panel in Interface > Addons, that can later be extended with more options.

WeakAuras/WeakAuras.toc
```diff
@@ -32,6 +32,7 @@ BuffTrigger.lua
 BuffTrigger2.lua
 GenericTrigger.lua
 AuraEnvironment.lua
+Config.lua
```
InfusOnWoW: Add it to babelfish.lua too

Stanzilla:
Hrm this is going straight to a territory we always wanted to avoid back in the days, options for options. Which was also the reason I was hesitant to adding code editor settings when we did that. That was mostly because of discoverability though. I don't want people to go to a different place (like the Blizzard options) to change stuff in WA, so I guess the idea Infus had could be explored, the new fake category above or below "new". 

Otherwise I would also be fine with just enabling the minimap icon by default, same as LDB feed, people can turn it off via middle click, it's not too bad of a UX problem and only happens once and adding the /wa minimap slash command.

InfusOnWoW:
Yeah, if we add options we need that in the main window too.

And we had a few occasions where adding a option would have been nice. And yeah, I agree that we can merge it even if the middle click only works "half" of the time.

emptyrivers:
I think it’s fine to have meta-options, but it needs to be really clear that they’re exactly that. If an interface options category is added, I’d want a button to open WeakAurasOptions if possible. 

Also, plenty of addons have an interface options tab which doesn’t do anything except show/hide the minimap icon, so I don’t think it’ll be too confusing for users what’s going on there.

Stanzilla:
> Also, plenty of addons have an interface options tab which doesn’t do anything except show/hide the minimap icon, so I don’t think it’ll be too confusing for users what’s going on there.

it's mostly just the lack of ways to point people to them


Anyway, I think for the purpose of this PR, we can do without custom options but that is something we can explore in the future.

mrbuds:
Alright, so remove the option panel, and re-add mid-click hiding for minimap icon.

Do you want to show in the tooltip that mid-click hide icon, knowing that it work only for minimap? I don't think it is possible to make a different tooltip for it.

InfusOnWoW:
I'm fine with both.

Stanzilla:
> Do you want to show in the tooltip that mid-click hide icon, knowing that it work only for minimap? I don't think it is possible to make a different tooltip for it.

Yes please

.pkgmeta
```diff
@@ -18,6 +18,7 @@ externals:
  WeakAuras/Libs/LibSpellRange-1.0: git://github.com/ascott18/LibSpellRange-1.0.git
  WeakAuras/Libs/LibRangeCheck-2.0: https://repos.wowace.com/wow/librangecheck-2-0/trunk/LibRangeCheck-2.0
  WeakAuras/Libs/LibCustomGlow-1.0: https://repos.curseforge.com/wow/libcustomglow
+ WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk
```
Stanzilla:
```suggestion
 WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk/LibDBIcon-1.0
```

nebularg:
> Do you want to show in the tooltip that mid-click hide icon, knowing that it work only for minimap? I don't think it is possible to make a different tooltip for it.

`self:GetParent() == Minimap` will tell you if it's a minimap icon

mrbuds:
> 
> 
> > Do you want to show in the tooltip that mid-click hide icon, knowing that it work only for minimap? I don't think it is possible to make a different tooltip for it.
> 
> `self:GetParent() == Minimap` will tell you if it's a minimap icon

Thanks for the tips !

WeakAuras/WeakAuras.lua
```diff
@@ -1159,6 +1168,110 @@ function WeakAuras.CreatePvPTalentCache()
...
+    if(paused) then
+      tooltip:AddLine("|cFFFF0000"..L["Paused"].." - "..L["Shift-Click to resume"]);
+    else
+      tooltip:AddLine(L["Shift-Click to pause"]);
+      tooltip:AddLine(L["Right-Click to toggle profiling"]);
+      tooltip:AddLine(L["Shift-Right-Click to show profiling results"]);
+    end
+  end
+  if frame:GetParent() == Minimap then
+    tooltip:AddLine("Middle-Click to hide minimap icon");
```
Stanzilla:
```suggestion
    tooltip:AddLine(L["Middle-Click to hide minimap icon"]);
```

WeakAuras/WeakAuras.lua
```diff
@@ -83,6 +84,14 @@ function SlashCmdList.WEAKAURAS(msg)
     elseif(msg == "pprint") then
       WeakAuras.PrintProfile();
       return;
+    elseif(msg == "minimap") then
+      WeakAurasSaved.minimap.hide = not WeakAurasSaved.minimap.hide
+      if WeakAurasSaved.minimap.hide then
+        LDBIcon:Hide("WeakAuras");
```
Stanzilla:
```suggestion
        LDBIcon:Hide("WeakAuras");
        prettyPrint(L["Use /wa minimap to show the minimap icon again"])
```

WeakAuras/WeakAuras.lua
```diff
@@ -1159,6 +1168,110 @@ function WeakAuras.CreatePvPTalentCache()
...
+        if not(WeakAuras.IsOptionsOpen()) then
+          WeakAuras.Toggle();
+        end
+      else
+        WeakAuras.OpenOptions();
+      end
+    elseif button == 'MiddleButton' then
+      WeakAurasSaved.minimap.hide = true
+      LDBIcon:Hide("WeakAuras");
+      prettyPrint(L["Use /wa minimap to show again"])
```
Stanzilla:
```suggestion
      prettyPrint(L["Use /wa minimap to show the minimap icon again"])
```

Stanzilla:
Left a few comments, might be worth considering adding a ToggleMinimap() function that does the print and hiding/showing.

mrbuds:
Ok i added a WeakAuras.ToggleMinimap() and use it for middle-click on icon and /wa minimap
It works fine with Bazooka, so i have updated the tooltip to display it with "show" or "hide" accordingly.

PR #906, opened 2018-10-29
handle sharing of auras with people crossrealm in same group

mrbuds:
When sending addonmessage trough the "WHISPER" channel, first check if unit is on same realm
If not then check if on same group or raid and send the message trough the "PARTY" or "RAID" channel with a prefix "§§destinationName-destinationServer:message"

Then when receiving messages, when it is from channel "PARTY" or "RAID", check for the prefix, and check if your are the right recipient.

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -545,7 +545,7 @@ hooksecurefunc("ChatFrame_OnHyperlinkShow", function(self, link, text, button)
         ShowTooltip({
           {2, "WeakAuras", displayName, 0.5, 0, 1, 1, 1, 1},
           {1, L["Requesting display information from %s ..."]:format(characterName), 1, 0.82, 0},
-          {1, L["Note, that cross realm transmission is not possible"], 1, 0.82, 0}
+          {1, L["Note, that cross realm transmission is possible if you are on the same group"], 1, 0.82, 0}
         });
         tooltipLoading = true;
         receivedData = false;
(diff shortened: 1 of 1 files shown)


WeakAuras/Transmission.lua
```diff
@@ -1675,6 +1675,18 @@ function WeakAuras.ImportString(str)
   end
 end
 
+function crossRealmSendCommMessage(prefix, text, chattype, target, queueName, callbackFn, callbackArg)
+  if chattype == "WHISPER" and target and not UnitIsSameServer(target) then
```
InfusOnWoW: What does UnitIsSameServer do for merged realms?

WeakAuras/Transmission.lua
```diff
@@ -1675,6 +1675,18 @@ function WeakAuras.ImportString(str)
...
+    if UnitInRaid(target) then
+      chattype = "RAID"
+    elseif UnitInParty(target) then
+      chattype = "PARTY"
+    end
```
InfusOnWoW: Shouldn't that do:
else
 return

?

WeakAuras/Transmission.lua
```diff
@@ -1675,6 +1675,18 @@ function WeakAuras.ImportString(str)
   end
 end
 
+function crossRealmSendCommMessage(prefix, text, chattype, target, queueName, callbackFn, callbackArg)
```
InfusOnWoW: There's no point in having a chattype argument in here, as that is always "WHISPER".

WeakAuras/Transmission.lua
mrbuds: i trusted https://wow.gamepedia.com/API_UnitIsSameServer on this, and i just did my own tests, it returns true when unit is on a linked server

WeakAuras/Transmission.lua
mrbuds: yes

WeakAuras/Transmission.lua
mrbuds: indeed

WeakAuras/Transmission.lua
mrbuds: If that happen the whole PR can be reverted

InfusOnWoW: We might not immediately recognize that. If the code would do what I suggest, it would just magically start working.

PR #921, opened 2018-11-05
Smallerpatches

InfusOnWoW:


WeakAurasOptions/Locales/enUS.lua
```diff
@@ -159,7 +159,8 @@ There are several special codes available to make this text dynamic:
 |cFFFF0000%n|r - Name - The name of the display (usually an aura name), or the display's ID if there is no dynamic name
 |cFFFF0000%i|r - Icon - The icon associated with the display
 |cFFFF0000%s|r - Stacks - The number of stacks of an aura (usually)
-|cFFFF0000%c|r - Custom - Allows you to define a custom Lua function that returns a list of string values. %c1 will be replaced by the first value returned, %c2 by the second, etc.]=]
+|cFFFF0000%c|r - Custom - Allows you to define a custom Lua function that returns a list of string values. %c1 will be replaced by the first value returned, %c2 by the second, etc.
+|cFFFF0000%%|r - % - To show a percent sign]=]
```
psykzz: Is `]=]` a typo here?

emptyrivers: No. In lua, string literals are denoted by:

- 'single quotes'
- "double quotes"
- [[double square brackets]]

Or:

- [=[double square brackets with any number of equals signs like this]=]

So long as the number of equals signs matches up. So,

```lua
s1 = [=[this is a valid string]=]
s2 = [===[so is this]===]
s3 = [=[but this is not]==]
s3 = [===[and neither is this]]
```

PR #931, opened 2018-11-07
Add help text and game restart warning

### LINKED ISSUE #925: /wa help
Buds mentioned this in passing, and I don’t want this to be forgotten.

Basically, we should add a help option to the WeakAuras slash command, which prints out all the subcommands which are available to them.

e.g.
```
> /wa help
Usage:
/wa help - show this message
/wa pstart - begin a fresh profile
/wa pstop - finish profiling
/wa pprint - bring up a window with the results from the most recent profile
```

Stanzilla:
Decisions I made:
- lower the slash command message in case people want to do /wa hElP etc 
- Display Discord URL but not GitHub because Discord can pre-filter most problems
- Display the version mismatch warning 1s after load so all the other addons don't spam it away, maybe display on options open as well?

Closes #925

### DIFF
--- a/WeakAuras/Init.lua
+++ b/WeakAuras/Init.lua
@@ -1,18 +1,28 @@
 WeakAuras = {}
 WeakAuras.L = {}
 
-local versionString = GetAddOnMetadata("WeakAuras", "Version");
+local versionStringFromToc = GetAddOnMetadata("WeakAuras", "Version");
+local versionString = "@project-version@"
 --@debug@
-if versionString == "@project-version@" then
-  versionString = "Dev"
+if versionStringFromToc == "@project-version@" then
+  versionStringFromToc = "Dev"
 end
 --@end-debug@
-WeakAuras.versionString = versionString
+
+WeakAuras.versionString = versionStringFromToc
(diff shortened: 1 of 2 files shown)


mrbuds:
#925 for reference

WeakAuras/Init.lua
```diff
@@ -1,20 +1,30 @@
...
 
 WeakAuras.prettyPrint = function(msg)
   print(WeakAuras.printPrefix .. msg)
 end
 
+WeakAuras.versionMismatchPrint = function()
+  WeakAuras.prettyPrint("You need to restart your game client to complete the WeakAuras update!")
+end
+
+if versionString ~= versionStringFromToc and versionString ~= "Dev" then
```
emptyrivers: This will cause a lot of false positives whenever there is an update which doesn’t add any new files.

Stanzilla: Yeah well, doesn't hurt either?

emptyrivers: Well, it’s probably fine. People have awful hygiene when it comes to updating addons anyways.

Issue #929, opened 2018-11-07
[Use-Case] Healthstone


Stanzilla:
**What spell item is this about**
Warlock Healthstones primarily, general items in the bag secondarily.

**What are the expected features, what is missing/broken?**
I'd expect it to be an icon that shows the cooldown of the current charge, the text saying how many charges are left.

The current problem is the text, in a pre- https://github.com/WeakAuras/WeakAuras2/pull/913 world we have to use custom code to get the item count and do update it on trigger update, it looks like our item count trigger is missing an event when healthstone charges change.

**Aura in Question**

[...]

PR #932, opened 2018-11-07
add BAG_UPDATE_COOLDOWN event to "Item Count" trigger #929

mrbuds:
doesn't #929 :D

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -3683,6 +3683,7 @@ WeakAuras.event_prototypes = {
     type = "status",
     events = {
       "BAG_UPDATE",
+      "BAG_UPDATE_COOLDOWN",
       "PLAYER_ENTERING_WORLD"
     },
     internal_events = {


Stanzilla:
I had tested that, and it does not fix it for me.

Stanzilla:
oh also please remove the fix comment since it only fixes one aspect of it and we want to keep the ticket open for discussion

Stanzilla:
> When Dynamic Information is set to trigger2, to get the cooldown, it does not update text, guess I did not realize trigger update is linked to that.

Should the display text `trigger update` fire for *any* trigger's update?

asakawa-k:
That behaviour is, at least, what we've disseminated. That Trigger Update fires when the current Dynamic Info changes ins some way. (Side note, for a long time I thought that Trigger Update was identical to Every Frame because I kept testing it with timed Auras).

With a Healthstone tracker, whose main trigger is the item's CD, I wouldn't expect that the item count _could_ change without it also affecting the main trigger. Though perhaps the order of the events fired means that the count hasn't updated when the CD updates.

InfusOnWoW:
Currently it should only fire for the active trigger updates, but with the %#.var change, it does trigger for all triggers.

Issue #915, opened 2018-11-02
[beta2] Advanced glow options are hidden if glow is disabled


asakawa-k:
* ~~"CLEU" will mean nothing to most people and only raise questions I think. "Multi-target" is a known name to people who used WA before. I understand wanting to spell out the difference between it and "nameplate" but I'm not sure adding "(CLEU)" helps much in that regard.~~
* ~~"Group member count" and "Match Count" are poorly differentiated. I had to ask before it made sense, even though I had followed some of the discussion about these options being created. I don't have a great solution to suggest, it's a relatively deep concept. Might need a toolltip.~~
* ~~"Group Member Count" makes sense for the "group" unit but is also used in Arena, Boss, and so on. "Boss Count"?~~
* ~~With "Combined Matches Per Unit" selected I expected `%matchCount` to give a per-unit count. So I could, say, see how many DoTs I have on each nameplate unit.~~
* ~~Feels odd to have a "Use Custom Glow Color" Condition option that must be enabled before the "Glow Color" option will take effect. It doesn't stand out as a problem in the Display Tab context but in separate Conditions it's awkward. Could they be combined? Is an Enabling bool needed at all?~~
* ~~Pixel Glow uses the ActionButtonGlow start animation. I think there may be a ticket about this already, or is otherwise known.~~
* ~~War Mode Active (Load tab) still flagged as new. That's from the previous major version I think(?)~~
* Changing the glow type used by the Icon Aura requires you to enable it (permanently glowing). You _could_ then disable it and the alternate settings would remain but that's not intuitive behaviour. I think most Auras using glows will enable them through Conditions but any alternate glow there, without changing the base settings, would need to change up to three properties each time (enable, change, type, change colour).

~~On a side note, could a `unitID` value be added to the states too? With options like "Nameplate" existing, having the unit available in code blocks like On Show could make things like anchoring to nameplates quite easy and, I'm guessing, would be a relatively simple addition?~~

asakawa-k:
Oh. I had a look in the state table with VDT and didn't see it. 
I see why though, I was looking at the region's state, not the specific clone. 
Got the unit now. 👍

emptyrivers:
[...]

PR #943, opened 2018-11-10
make glow type option visible when glow is disabled

emptyrivers:
This is a bandaid fix to directly address the concerns in the issue.
There is a more fundamental issue present, which requires more time to resolve.
Specifically, there are several options which show/hide a 'subregion'
and also show/hide several sub options for that 'subregion'. So, it is
impossible to edit a subregion's sub options without showing it. Before now, this
was not an issue, since there was no subregion which you could hide or show in
a condition, and also which had sub options. But now the glow subregion does, and
I expect that future development will cause this same problem to appear with other
subregions.

GitHub Issue #915

### DIFF
--- a/WeakAurasOptions/RegionOptions/Icon.lua
+++ b/WeakAurasOptions/RegionOptions/Icon.lua
@@ -90,20 +90,17 @@ local function createOptions(id, data)
       name = L["Glow Type"],
       order = 21,
       values = WeakAuras.glow_types,
-      hidden = function() return not data.glow end,
     },
     useGlowColor = {
       type = "toggle",
       name = L["Glow Color"],
       desc = L["If unchecked, then a default color will be used (usually yellow)"],
       order = 23,
-      hidden = function() return not data.glow end,
     },
     glowColor = {
       type = "color",
       name = L["Glow Color"],
       order = 24,
-      hidden = function() return not data.glow end,
       disabled = function() return not data.useGlowColor end,
     },
     textHeader1 = {


WeakAurasOptions/RegionOptions/Icon.lua
```diff
@@ -90,7 +90,6 @@ local function createOptions(id, data)
       name = L["Glow Type"],
       order = 21,
       values = WeakAuras.glow_types,
-      hidden = function() return not data.glow end,
```
InfusOnWoW: What about glow color ?

InfusOnWoW:
Are there any other cases where we hide options, in that way? I don't think we have anything atm.

emptyrivers:
Pretty much every "subregion" (text 1, text 2, aurabar icon, aurabar spark, aurabar border) which has sub options has the same pattern of "you have to enable the subregion in order to modify its properties in the display tab". Just, up until now, I don't believe that any of them have the option to be *shown* via conditions, like the glow effect, so it was never an issue. #866 will introduce the same issue, just with the cooldown swipe on icons instead of the glow effect.

InfusOnWoW:
Right, so there's curerntly no other problem. Good.

WeakAurasOptions/RegionOptions/Icon.lua
emptyrivers: I didn't include it because it wasn't part of the original ticket. Do you want me to include that anyways?

InfusOnWoW: Well, I think the same problem applies to it too.

PR #961, opened 2018-11-18
Class Colour Toggles

Rehok:
I've added a Class Color toggle button to each region that supports texts, as some people don't have Colour picker addons or would use %c this is would be a easier way than doing custom code. I'm sure this could be implemented better and it only supports player class right now, Not sure if it could support clones with peoples names being sent?


asakawa-k:
With `state.unit` existing on so much stuff now that seems like a missed opportunity, and the most likely case, other than the player's own class, for class colouring to be used.

PR #962, opened 2018-11-18
Class Color toggles

Rehok:
I've added a Class Color toggle button to each region that supports texts, as some people don't have Colour picker addons or would use %c this is would be a easier way than doing custom code. I'm sure this could be implemented better and it only supports player class right now, not entirely sure how to access unit data for this toggle?

### DIFF
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -1218,9 +1218,9 @@ local function modify(parent, region, data)
     if data.timerClassColor then
       local col = RAID_CLASS_COLORS[select(2, UnitClass("player"))]
       timer:SetTextColor(col.r, col.g, col.b, col.a);
-      else
-        timer:SetTextColor(data.timerColor[1], data.timerColor[2], data.timerColor[3], data.timerColor[4]);
-      end
+    else
+      timer:SetTextColor(data.timerColor[1], data.timerColor[2], data.timerColor[3], data.timerColor[4]);
+    end
     animRotate(timer, textDegrees);
     timer:Show();
     timer.visible = true;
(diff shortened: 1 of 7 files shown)


InfusOnWoW:
I don't like how specific the solution is and solves just one use case at the cost of many additional options. That doesn't feel like the right trade-off to me.

Stanzilla:
I wondered if having something like LuaTexts https://www.wowace.com/projects/pitbull-unit-frames-4-0/pages/lua-texts would be worth it

emptyrivers:
Isn't that just an alternative to what we have with custom texts right now?

I've never used pitbull, so I'm rather unfamiliar with the system.

Stanzilla:
Yeah but we could some ideas from it to fix a few usecase problems with have, like the https://github.com/WeakAuras/WeakAuras2/pull/393 so there could be %progress and %progressShort and %name and %nameCColor etc

Rehok:
The idea would be to eventually get this to work with %name,%sourceName etc from all triggers and I thought if I had a pull request im more likely to not forget about this (I plan on doing more work tonight to see where i can use data.unit) and could get feedback on how the code looks if its all correct or better ways to do it. As this is the first time i've ever done UI Related settings I don't know if the Setter function is just passed a bool or does it get passed data aswell (judging from other ones I don't think it does)

emptyrivers:
I would rather have this as something you write into the text pattern itself, honestly.

So you could write:

```
%hunter|name
```
which would resolve to `"|cFFABD473"..state.name.."|r"`. Or something like that.

emptyrivers:
Dynamic colors are a feature that I think we'll look into in a more general sense. Will keep this open partially to remind myself of that.

Desik:
I feel like a simple "Use Class Color" checkbox would be really useful for everything that needs a color defined except for texts,  I still think that texts should be handle like @emptyrivers suggested simply because text is different in a way, like for example you probably only want to class color the name of the player in a sentence and not the whole sentence or something like that but for bars, borders and glows class colors would be a great addition to the already existing customization options available.

Personally I had to hard code class color on some of my auras, I play all 12 classes, so some of my auras are multiclass, It would be great if I could just select "Use Class Color" and be done with it. I can see other auras using this feature too like class bars, interrupt bars/icons and just in general I don't think it would take too much real estate in the options.

emptyrivers:
Forward duping to #2013

PR #978, opened 2018-11-25
re-inverse cast progression like it was done before

mrbuds:


WeakAuras/Prototypes.lua
```diff
@@ -4493,6 +4493,7 @@ WeakAuras.event_prototypes = {
                 destName = UnitExists(destUnit) and UnitName(destUnit) or "",
                 show = true,
                 changed = true,
+                inverse = true,
```
InfusOnWoW: I'm pretty sure inverse depended on whether the cast was channeling or not.

mrbuds: you are right

WeakAuras/Prototypes.lua
```diff
@@ -4493,6 +4493,7 @@ WeakAuras.event_prototypes = {
                 destName = UnitExists(destUnit) and UnitName(destUnit) or "",
                 show = true,
                 changed = true,
+                inverse = castType == "cast",
```
InfusOnWoW: Did you check that it now behaves like it did before?

mrbuds: yes i tested cast and channeling with older version and it behaved as you said

mrbuds:
Removed show = false on events UNIT_SPELLCAST_FAILED and UNIT_SPELLCAST_FAILED_QUIET

It fix an issue reported on discord with channeling, when "clicking" a channel spell that is already in progress it canceled the progress. (tested with penance)

Issue #971, opened 2018-11-21
[PTR] Opening / adding code to code blocks causes game to crash/code to duplicate


Causese:
**Describe the bug**
- Opening a codeblock w/o any code causes it to duplicate anything you type in there
- Opening a codeblock with code inside causes it to duplicate existing code? and causes a crash at some point unless you hit escape real fast

**Do you have an error log of what happened?**
no error

**To Reproduce**
Either open a code block with code inside to crash or open a codeblock without any code inside and type something for it to duplicate

**Screenshots**
https://i.gyazo.com/69ffbcbb293a6f98c655fde97a7cb721.gif

**Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?**

**Which version of WeakAuras are you using?**
2.9.1

PR #986, opened 2018-11-28
fix texteditor crashing on PTR - ticket #971

mrbuds:
fix ```60: attempt to index global 'L' (a nil value)[C]: in function `xpcall'``` when using cast trigger with option "inverse"

### DIFF
--- a/WeakAurasOptions/ForAllIndentsAndPurposes.lua
+++ b/WeakAurasOptions/ForAllIndentsAndPurposes.lua
@@ -1,5 +1,5 @@
 -- For All Indents And Purposes
-local revision = 19
+local revision = 20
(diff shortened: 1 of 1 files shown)


mrbuds:
fix #971, ForAllIndentsAndPurposes library get the position of the cursor by inserting "\1" in the textbox and then search the position of the character with string.find
Position always return nil
I replaced the \1 with a weird utf-8 icon ❦

Stanzilla:
Need to bump the revision in ForAllIndentsAndPurposes

Stanzilla:
infus already cherrypicked your other commit, can you remove it from this PR please?

mrbuds:
i removed it and rebased

PR #992, opened 2018-11-29
Wago update support

mrbuds:
From the field "url" extract version & slug, then search for the slug in a global Array "WeakAurasWagoUpdate".
Then compare versions and show update icon if a newer version of the aura is saved in WeakAurasWagoUpdate.
Clicking the icon call WeakAuras.ImportString()
Mouseover it show a tooltip with full aura name, author, old -> new version, and is ready to show some version note

It is possible to hide specific updates from the right click menu on auras, or group of aura, that information is (or will, i'm not sure if Ora has uploaded this modification on wago yet) remove from data by wago when importing the string.

Example data file for testing:
[WeakAurasWagoUpdate.zip](https://github.com/WeakAuras/WeakAuras2/files/2630390/WeakAurasWagoUpdate.zip)

preview

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -749,7 +749,7 @@ local methods = {
       if self.update.slug then
         local wago = WeakAurasWagoUpdate[self.update.slug]
         if wago then
-          WeakAuras.ImportString(wago.string)
+          WeakAuras.ImportString(wago.encoded)
         end
       end
     end
@@ -960,10 +960,10 @@ local methods = {
       local slug, version = self.data.url:match("wago.io/([^/]+)/([0-9]+)")
       if slug and version then
         local wago = WeakAurasWagoUpdate[slug]
-        if wago and wago.version > version then
+        if wago and wago.wagoVersion > version then
           self.update.slug = slug
           self.update.version = version
-          self.update.wagoVersion = wago.version
+          self.update.wagoVersion = wago.wagoVersion
           self.update.desc = L["From version "] .. self.update.version .. L[" To version "] .. self.update.wagoVersion;
           self.update:Show()
           self.update:Enable();
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -745,7 +745,6 @@ local methods = {
     end
 
     function self.callbacks.OnUpdateClick()
-      if not WeakAurasWagoUpdate then return end;
       if self.update.slug then
         local wago = WeakAurasWagoUpdate[self.update.slug]
         if wago then
@@ -954,22 +953,24 @@ local methods = {
     self.ungroup:SetScript("OnClick", self.callbacks.OnUngroupClick);
     self.upgroup:SetScript("OnClick", self.callbacks.OnUpGroupClick);
     self.downgroup:SetScript("OnClick", self.callbacks.OnDownGroupClick);
-    self.update:SetScript("OnClick", self.callbacks.OnUpdateClick);
-
-    if WeakAurasWagoUpdate and self.data.url then
-      local slug, version = self.data.url:match("wago.io/([^/]+)/([0-9]+)")
-      if slug and version then
-        local wago = WeakAurasWagoUpdate[slug]
-        if wago and wago.wagoVersion and wago.wagoVersion > version then
-          self.update.slug = slug
-          self.update.version = version
-          self.update.name = wago.name
-          self.update.author = wago.author
-          self.update.wagoVersion = wago.wagoVersion
-          self.update.title = L["Update "] .. self.update.name .. L[" by "] .. self.update.author
-          self.update.desc = L["From version "] .. self.update.version .. L[" To version "] .. self.update.wagoVersion;
-          self.update:Show()
-          self.update:Enable();
+    
+    if WeakAurasWagoUpdate then
+      self.update:SetScript("OnClick", self.callbacks.OnUpdateClick);
+      if self.data.url then
+        local slug, version = self.data.url:match("wago.io/([^/]+)/([0-9]+)")
+        if slug and version then
+          local wago = WeakAurasWagoUpdate[slug]
+          if wago and wago.wagoVersion and wago.wagoVersion > version then
+            self.update.slug = slug
+            self.update.version = version
+            self.update.name = wago.name
+            self.update.author = wago.author
+            self.update.wagoVersion = wago.wagoVersion
+            self.update.title = L["Update "] .. self.update.name .. L[" by "] .. self.update.author
+            self.update.desc = L["From version "] .. self.update.version .. L[" To version "] .. self.update.wagoVersion;
+            self.update:Show()
+            self.update:Enable();
+          end
         end
       end
     end
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -801,6 +801,69 @@ local methods = {
       end
     end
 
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -792,6 +799,73 @@ local methods = {
...
+      -- remove ignore flags
+      self.data.ignoreWagoUpdate = nil
+      -- show update frame
+      if self.data.url and self.data.url ~= "" then
+        local slug, version = self.data.url:match("wago.io/([^/]+)/([0-9]+)")
+        if slug and version then
+          local wago = WeakAurasWagoUpdate[slug]
+          if wago and wago.wagoVersion and wago.wagoVersion > version then
+            self.update.title = L["Update "] .. wago.name .. L[" by "] .. wago.author
+            self.update.desc = L["From version "] .. version .. L[" To version "] .. wago.wagoVersion
```
InfusOnWoW: *to* version

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -944,6 +1026,11 @@ local methods = {
     self.ungroup:SetScript("OnClick", self.callbacks.OnUngroupClick);
     self.upgroup:SetScript("OnClick", self.callbacks.OnUpGroupClick);
     self.downgroup:SetScript("OnClick", self.callbacks.OnDownGroupClick);
+
+    if WeakAurasWagoUpdate and not self.data.ignoreWagoUpdate then
+      self.callbacks.ShowWagoUpdate()
```
InfusOnWoW: This is the initialization function of a singular button. ShowWagoUpate() recursively updates all child buttons, thus any child button is initialized twice. Also the code above to initialze the menu is overwritten by ShowWagoUpdate() in some cases.

Also ShowWagoUpdate() does various things that aren't truely needed on each Show/Hide toggling. E.g. the lookup of the slug and the title/description setup.

I think you might be better to move some of the code from Show e.g. the lookup of of the slug to this initialize function and to not call ShowWagoUpdate here.

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -1705,6 +1792,25 @@ local function Constructor()
   expand:SetScript("OnEnter", function() Show_Tooltip(button, expand.title, expand.desc) end);
   expand:SetScript("OnLeave", Hide_Tooltip);
 
+  local update
+  if WeakAurasWagoUpdate then
+    update = CreateFrame("BUTTON", nil, button);
+    button.update = update
+    update.disabled = true;
+    update.func = function() end;
+    update:SetNormalTexture("interface\\minimap\\rotating-minimapcorpsearrow.blp");
```
InfusOnWoW: You can ask Exality if he wants to create a icon, I think a wago Icon with a download arrow might work pretty well.

WeakAuras/WeakAuras.lua
```diff
@@ -1187,10 +1187,41 @@ function WeakAuras.CreatePvPTalentCache()
...
+
 local function tooltip_draw()
   local tooltip = GameTooltip;
   tooltip:ClearLines();
   tooltip:AddDoubleLine("WeakAuras", versionString);
+  if WeakAurasWagoUpdate then
+    local updatedSlugsCount, updatedAuras = WeakAuras.CountWagoUpdates()
+    if updatedAuras > 0 then
+      tooltip:AddLine(" ");
+      tooltip:AddLine((L["%i updates from Wago for %i auras are ready to be install"]):format(updatedSlugsCount, updatedAuras));
```
InfusOnWoW: So if I understood that correctly, updatedSlugsCount and updatedAuras can differ if the user has multiple auras that have the same url in them? That sounds like a edge case.

And the message sounds a bit confussing in the normal case:
"5 update from Wago for 5 auras are ready to be installed."

I think just counting the number of auras that can be updated is good enough.

InfusOnWoW:
How does wago work with urls for groups of auras? Does wago remove the urls of childs, so that only the parent has a url?

InfusOnWoW:
Also I can't really review: https://github.com/mrbuds/wa-update/, as I have zero clues about the stuff in there. 

The PR looks good to me now.

mrbuds:
On import Wago add url field to the group and each aura in the group.

On the example screenshot i provide i only have ungrouped individual auras that needs update, that's why the distinction doesn't look clear on it.

If you have a group of let's say 20 auras the message will be "1 updates from Wago for 21 auras are ready to be install" (20+1 for the group)

InfusOnWoW:
Hmm, then displaying the update count is probably best? I'm not sure people will understand two numbers being there.

mrbuds:
I included Exality's animated wago icon.

I'm not sure i understand what you are asking @emptyrivers, how it's done now, you can mark auras for stop updating them, but you still have the ability to revert it.

What is missing is a visual on button to show that aura has the ignoreWagoUpdate field, is this what you want ?

emptyrivers:
No. I want the ability to tell WeakAuras to ignore a particular version, but if a *new* version comes along for the same uid, then notify me.

mrbuds:
I added a new menu entry for skipping current available update
It set self.data.skipWagoUpdate with version available from addon
The menu entry show only when aura is "synced" with wago

I also made the "sync with wago" menu entry show only if there is parse-able url defined.

Btw, i don't think "sync" is the right word to use here, as a author-user if i click on "sync with wago" i would expect it to take care of uploading my updates to wago, which it doesn't.

sragia:
That sounds like great feature for future. Get Ora to let you decide version format (major.minor.patch or w/e) and in game tag aura for release and companion app uploads it for you on game reload/logout.

> Btw, i don't think "sync" is the right word to use here, as a author-user if i click on "sync with wago" i would expect it to take care of uploading my updates to wago, which it doesn't.

Stanzilla:
`Update from Wago`?

mrbuds:
This was all too confusing i made a submenu with more clear labels, what do you think?
menu1
if i click on ignore this update it hide update icon and change menu like this:
menu2
if i click on ignore all updates it hide update icon and entry 2nd menu entry, and set ignore flag for updater
menu3
and i click on this it show icon back and show menu like on 1st screen

emptyrivers:
As a side note, I think it would be nice to not assume that the only source of updates will be wago. As far as WeakAuras is *actually* concerned, the source of these updates is an unknown addon calling an API. I can see people wanting to build addons which register updates directly, in which case naming the api as a wago one is incorrect.

That is beyond the scope of this initial pass, however.

Stanzilla:
Is that something we want to support though?

emptyrivers:
Why wouldn’t we want to do that? Sooner or later someone will do it anyways. I think it would be a better experience if addon authors didn’t have to pretend they were wago for this.

mrbuds:
I'm not against the idea but that require a complete rewrite of this PR to add a layer of abstraction, if you arer ok that's something i can do next year.

I have added "Update this aura" in the "Wago Update" menu, and updated the callbacks to show/hide it with the update icon

emptyrivers:
I think it’s fine to assume that only wago will source updates for now. As I said before, improving that out of scope.

Stanzilla:
Yeah well basically what buds said, would have to be rebuilt around a system where addons can register themselves as update provider etc. Can all be done at a later point if needed.

Stanzilla:
Also what about `Unignore`?

mrbuds:
I don't like this word

mrbuds:
UNIGNORE_QUEST = "Unignore"
IGNORE = "Ignore"
IGNORE_REMOVE = "Stop Ignoring"

Stanzilla:
Stop Ignoring Updates still weird?

emptyrivers:
Globalstrings don’t really let us skip translation, unless you can manage one which already says exactly what we want without concatenating something onto it.

Like, `IGNORE .. L[“ Updates”]` still requires a translation, so you may as well have the phrase be `L[“Ignore Updates”]`. Also, assuming the order of words in other languages is the same as English doesn’t always turn out well (I have no idea if that would be a problem here).

Stanzilla:
agree

PR #996, opened 2018-12-03
Make frameId unique

Dooez:
There is [code](https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/RegionTypes/Icon.lua#L225) for making frameId of icon regions unique, however it does not work currently.


InfusOnWoW:
Can you explain why it doesn't work?

Dooez:
There is no _G[frameId]. From what i can see only cooldown frame with "WeakAurasCooldown"..frameId is created in global

PR #999, opened 2018-12-06
Check for tables for default in AuthorOptions

### LINKED ISSUE #998: [Author Options] "Reset to Defaults" button comes back after reload
My WeakAura ([link](https://wago.io/IgnorePainTracker)) has an issue where the "Reset to Defaults" button will come back after reloading despite all of the author options already being at their default value.

I've made a few other auras with author options and the issue does not occur with those.

I compared the exported lua tables before/after hitting the Reset to Defaults button, and there is no difference, so nothing is actually changing. Leading me to believe that it's not an issue with the way I've made this aura (there is nothing in the code that directly modifies the aura_env.config table), i.e., it seems to be a bug with WeakAuras

xMarok:
Fixes #998

### DIFF
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -1167,7 +1167,7 @@ end
 local function allChoicesAreDefault(data)
   for _, option in ipairs(data.authorOptions) do
     if option.key ~= nil and option.default ~= nil and option.default ~= data.config[option.key] then
-		return tablesAreEqual(option.default, data.config[option.key])
+		  return tablesAreEqual(option.default, data.config[option.key])
     end
   end
   return true
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -1146,22 +1146,22 @@ local function mergeOptions(childIndex, merged, toMerge)
 end
 
 local function valuesAreEqual(t1, t2)
-	if t1 == t2 then return true end
-	local ty1 = type(t1)
-	local ty2 = type(t2)
-	if ty1 ~= ty2 then return false end
-	if ty1 ~= "table" then return false end
+  if t1 == t2 then return true end
+  local ty1 = type(t1)
+  local ty2 = type(t2)
+  if ty1 ~= ty2 then return false end
+  if ty1 ~= "table" then return false end
 
-	for k1, v1 in pairs(t1) do
-		local v2 = t2[k1]
-		if v2 == nil or not valuesAreEqual(v1, v2) then return false end
-	end
- 
-	for k2, v2 in pairs(t2) do
-		local v1 = t1[k2]
-		if v1 == nil or not valuesAreEqual(v1, v2) then return false end
-	end
-	return true
+  for k1, v1 in pairs(t1) do
+    local v2 = t2[k1]
+    if v2 == nil or not valuesAreEqual(v1, v2) then return false end
+  end
+
+  for k2, v2 in pairs(t2) do
+    local v1 = t1[k2]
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1147,13 +1147,26 @@ end
 
 local function allChoicesAreDefault(data)
   for _, option in ipairs(data.authorOptions) do
-    if option.key ~= nil and option.default ~= nil and option.default ~= data.config[option.key] then
+    if option.key ~= nil and option.default ~= nil and (option.default ~= data.config[option.key] or compareTables(option.default, data.config[option.key])) then
```
InfusOnWoW: This doesn't work because the logic is the wrong way around. You want to check if the tables are different.

Also this will not work if you are comparing two indentical values, as in that case compareValues is called with two values, and calling pairs on that will result in a error.

xMarok:
ef5be30

I believe that should do it. It's recursive and fixes the logic for checking if the values match.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1145,10 +1145,29 @@ local function mergeOptions(childIndex, merged, toMerge)
...
+		if v1 == nil or not tablesAreEqual(v1, v2) then return false end
+	end
+	return true
+end
+
 local function allChoicesAreDefault(data)
   for _, option in ipairs(data.authorOptions) do
     if option.key ~= nil and option.default ~= nil and option.default ~= data.config[option.key] then
-      return false
+		  return tablesAreEqual(option.default, data.config[option.key])
```
InfusOnWoW: The loop needs to continue if the values are equal. It should only return false if the values are not equal.
Also your function also checks for equality for non-table values, so the original check "option.default ~= data.config[option.key]" should not be necessary.

We might have a table comparing equal function already somewhere.

xMarok: Gotcha, sorry. Latest commit should (finally) be correct.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1145,11 +1145,33 @@ local function mergeOptions(childIndex, merged, toMerge)
...
+	end
+	return true
+end
+
 local function allChoicesAreDefault(data)
   for _, option in ipairs(data.authorOptions) do
-    if option.key ~= nil and option.default ~= nil and option.default ~= data.config[option.key] then
-      return false
-    end
+    if option.key ~= nil and option.default ~= nil then
```
emptyrivers:
```diff
 if option.key ~= nil and option.default ~= nil then
-  if tablesAreEqual(option.default, data.config[option.key]) then
-  elseif option.default ~= data.config[option.key] then
+ if not tablesAreEqual(option.default, data.config[option.key]) then
    return false
  end
end
```

xMarok: Oh, yeah that's more simple.

Why is that not correct? With that, it still returns true/false as it should, no (looking at the example inputs from my latest comment)?

emptyrivers: Well, for one thing, you can simply check that the option class is not `noninteractive` (see the top of the file), instead of checking that the key and default aren't nil.

InfusOnWoW: Right it's indeed correct. But that's not how that check should be written. A if, whose only purpose is to act as a additional check for the elseif branch is unnecessarily complex.

emptyrivers:
I pushed a fix to your whitespace (you should install editorconfig so that it's taken care of for you)

InfusOnWoW:
Thanks, merged now.

PR #1000, opened 2018-12-06
remove ignore & skip wagoUpdate flags when compressing data for export

mrbuds:


Stanzilla:
@emptyrivers can you make an issue for that?

PR #1002, opened 2018-12-08
Prepare for collapsible groups

emptyrivers:
This will cause a slight appearance change in the UI, when auras of disparate region types are Picked. Instead of a large font description header, an actual header type is chosen (this can of course be adjusted if necessary). Beyond that, there should be no user-facing changes from this pull request.

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -2604,7 +2604,7 @@ local function copyOptionTable(input, orderAdjustment)
   return resultOption;
 end
 
-local function flattenRegionOptions(allOptions, withoutHeader)
+local function flattenRegionOptions(allOptions)
   local result = {};
   local base = 1000;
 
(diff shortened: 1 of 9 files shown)


WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3510,7 +3474,7 @@ function WeakAuras.ReloadGroupRegionOptions(data)
     if(childData) then
       if (not regionTypes[childData.regionType]) then
         regionTypes[childData.regionType] = true;
-        regionTypeCount = regionTypeCount +1;
+        regionTypeCount = regionTypeCount + 1;
```
InfusOnWoW: Isn't regionTypeCount unused now?

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3533,32 +3497,26 @@ function WeakAuras.ReloadGroupRegionOptions(data)
...
+  replaceNameDescFuncs(allOptions, data);
+  replaceImageFuncs(allOptions, data);
+  replaceValuesFuncs(allOptions, data);
+  removeFuncs(allOptions);
+  fixMetaOrders(allOptions);
 
-  if (regionTypeCount > 1) then
-    addHeadersForRegionOptions(allOptions, regionOption);
-  end
+  local regionOption = flattenRegionOptions(allOptions, false);
```
InfusOnWoW: flattenRegionOptions only takes one value now.

Issue #944, opened 2018-11-11
Collapsible Option Groups


emptyrivers:
**Is your feature request related to a problem? Please describe.**

There's often dozens of controls in each tab of WeakAurasOptions. This can quickly overwhelm new users and slow down experienced users. It would be great if I could temporarily hide the options which I'm not interested in right now.

**Describe the solution you'd like**

There should be buttons on each "group" of options which, when clicked, toggle that group between being expanded (all options shown), and collapsed (only the header, and the button to expand the group is shown).

**Describe alternatives you've considered**

The only alternative I can think of is to start nesting option groups using AceConfig's own option group system. But this has its own problems, like being very unwieldy to write and use unless inline groups are used, and inline groups have the problem that they consume some of the available width. We already have issues with width playing havoc with the layout of our controls, so I don't think it's a good idea to invite yet more troublemakers, so to speak.

**Additional context**

#694 does this for custom options while in author-mode, and extending that to all of the options interface would be a solution to the root cause of #915 (and a better fix than #943). However, the scope of this is too large to be done in a bug-fix release, so I am writing this ticket instead and intend to work on this for 2.11, if not 2.10.

InfusOnWoW:
https://github.com/emptyrivers/WeakAuras2/tree/collapse

InfusOnWoW:
This is part of the next release.

emptyrivers:
Actually it’s already been released.

PR #1010, opened 2018-12-11
Collapsible Option Groups in Display tab

emptyrivers:
Implements #944 for the Display tab

## Type of change

## How Has This Been Tested?

- Create a new aura, and then click on the display tab
- You will see that the headers for each group include a button which, when clicked, will cause the group to collapse down to just the header.

## Work Remaining:

- Split option groups down into less monolithic groups:
	- [ ] Icon
	- [ ] Text
	- [ ] Aurabar
	- [ ] Texture
	- [ ] Progress Texture
	- [ ] Model
	- [ ] Dynamic Group
	- [ ] Group
- [ ] Verify that multi-select integrates nicely with collapsible option groups
- Self-review:
	- [x] Conform to Style Guidelines
	- [ ] Comments, commit messages, & documentation explain the changes properly
	- [x] No new warnings
	- [x] No known regressions


emptyrivers:
Rebased on top of #1034.

mrbuds:
Looks good to me.

I guess this is it out of the scope of this PR, but i wished Icon's texts were on there own option group

emptyrivers:
Yeah, that's the next step. All of the subregions will get their own header.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -3459,39 +3459,29 @@ end
...
-  local id = data.id;
-  WeakAuras.EnsureOptions(id);
-  local options = displayOptions[id];
-
-  local allOptions = {};
-  for regionType in pairs(regionTypes) do
-    if(regionOptions[regionType]) then
-      allOptions = union(allOptions, regionOptions[regionType].create(id, data));
-    else
-      regionType = {
```
InfusOnWoW: That code needs to stay.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -2687,7 +2695,30 @@ function WeakAuras.AddOption(id, data)
           end
           WeakAuras.ResetMoverSizer();
         end,
-        args = flattenRegionOptions(regionOption, id);
+        args = flattenRegionOptions(regionOption, id),
+        handler = {
```
InfusOnWoW: How does that work? I've never seen that.

emptyrivers: Every field in an AceConfig widget (except, irritatingly, for image) accepts a string value, in which case AceConfig will search down the options tree for a handler object, and call the relevant method. So in this case, on the collapse execute button, when clicked it calls `handler:CollapsedFunc(info, ...`.

I did it this way because I couldn’t see any other option to avoid RemoveFuncs from removing the function when multi-selecting.

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW: Indeed, I looked at each commit in isolation.

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW: So how does that work for the other places where we use a func, such as e.g. expand code editor? Those do work without requiring a handler?

emptyrivers:
Hmm, fair point. No idea how the Code Options manage this. I can take a second look.

Stanzilla:
Dynamic Group settings can not be collapsed, intended?

emptyrivers:
I blame infus for breaking group setting headers.

Stanzilla:
Imo one should be able to click the collapse header names in addition to the chevron, that possible?

emptyrivers:
It should be possible, and definitely worked as expected the last time I touched this pull request.

InfusOnWoW:
I fixed that collapsing on the group tab, and the position thing and the strange tooltip issue.

I couldn't reproduce the scrolling issue.

I don't know whether it's possible to make the whole header clickable. From a brief look, that's trivial.

PR #1034, opened 2018-12-16
Teach SetRegion to create regions initially hidden

emptyrivers:
# Description

If a user has enough auras that the login thread takes > 15 seconds to finish, then the game will begin properly before login is done. So ensure that the regions created by SetRegion do not start out shown.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] Install the SV file in #1015, and then log in.
- [x] Note that there are no regions visible when they should not be.

# Checklist:


emptyrivers:
This also appears to have a positive affect on performance, as well. Possibly due to the regions not needing to be re-rendered many times.

mrbuds:
I have not look at the code yet, just tested on trial account with big sv, no more temporary "?" icons everywhere on screen and load much faster

PR #1046, opened 2018-12-20
Stop showing update icons on every child of a group when possible

mrbuds:
# Description

If group has an update, don't show icon on childs with same source
refactor display of update icons and menu in button:RefreshUpdateIcon()
if button is part of a group, RefreshGroupUpdateIcon will do it
initializing update icons/menu after all buttons are created

preview

## Type of change

# How Has This Been Tested?
Test ignore next version, ignore all, stop ignoring on
- Top level aura
- Group
- Aura inside a group
- move an aura from source "A" into a group with source "B"

# Checklist:

It might be good to add sub co-routine for the initialization

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -803,160 +803,32 @@ local methods = {
       end
     end
 
-    function self.callbacks.wagoStopIgnoreAll()
-      -- remove ignore flags
+    function self.callbacks.wagoStopIgnoreAll(_, skipUpdateIcon)
       self.data.ignoreWagoUpdate = nil
       self.data.skipWagoUpdate = nil
-      -- update menu entries
-      local wagoMenu = self.menu[8].menuList
-      wagoMenu[1].func = self.callbacks.wagoIgnoreAll
-      wagoMenu[1].text = L["Ignore all Updates"]
-      if self:HasUpdate() then
-        tinsert(wagoMenu, {
-          text = L["Ignore this Update"],
-          notCheckable = 1,
-          func = self.callbacks.wagoIgnoreNext
-        })
-        tinsert(wagoMenu, {
-          text = " ",
-          notClickable = 1,
-          notCheckable = 1,
-        });
-        tinsert(wagoMenu, {
-          text = L["Update this Aura"],
-          notCheckable = 1,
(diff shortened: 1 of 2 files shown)


WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -1629,17 +1501,50 @@ local methods = {
...
+          for childIndex, childId in pairs(self.data.controlledChildren) do
+            local button = WeakAuras.GetDisplayButton(childId)
+            if button then
+              button.callbacks[actionFunc](nil, true)
+            end
+          end
+        end
+        WeakAuras.RefreshGroupUpdateIcon(self)
+      else
+        local hasUpdate, skipVersion = self:HasUpdate()
```
InfusOnWoW: This code is almots identically to what RefreshGroupUpdateIcon does.

If I see that correctly, making RefreshGroupUpdateIcon handle top level auras in addition to groups requires just adding a check for data.controlledChildren around the loop.

And then this code can be simplified a bit too.

Otherwise, it took me a bit to figure out how the recursion works. I don't have a suggestion on how to write that clearer though.

mrbuds: i removed WeakAuras.RefreshGroupUpdateIcon, it was indeed unnecessary and combining them should make the code a bit easier to understand

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -1629,17 +1501,68 @@ local methods = {
...
+            local childHasUpdate, childSkipVersion, _, childSlug = childButton:HasUpdate()
+            if childHasUpdate and slug ~= childSlug and not childSkipVersion then
+              showGroupUpdateIcon = true
+              childButton:ShowUpdateIcon()
+            else
+              childButton:HideUpdateIcon()
+            end
+          end
+        end
+        self:RefreshUpdateMenu()
```
InfusOnWoW: These lines here are identically between groups / top level.

For both you are calling RefreshUpdateMenu + Show/HideUpdateIcon, and for both your arecalling self:HasUpdate.

So what you can do is move the if self.data.controlledChildren down to just after the line "local showGroupUpdateIcon = false". And move the end of that if to just above self:RefreshMenu.

PR #1066, opened 2018-12-30
Misc

emptyrivers:
# Description

Fixes a semi-rare edge case with the import tooltip, and implements [multiselect](https://github.com/WeakAuras/WeakAuras2/projects/6#card-16090799) as a custom option

## Type of change

- [x] This change requires a documentation update.

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -304,7 +304,7 @@ local function install(data, oldData, patch, mode, isParent)
       for key in pairs(patch) do
         local checkButton = keyToButton[key]
         if not isParent and checkButton == checkButtons.anchor then
-          checkButton = checkButton.arrangement
+          checkButton = checkButtons.arrangement
         end
         if checkButton and not checkButton:GetChecked() then
           patch[key] = nil
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -883,7 +883,7 @@ local function up(data, option, index)
   if option.references then
     return function()
       for _, optionID in pairs(option.references) do
-        if index == 1 then return true end
+        if optionID <= 1 then return true end
       end
     end, function()
       for childID, optionID in pairs(option.references) do
@@ -914,7 +914,7 @@ local function down(data, option, index)
     return function()
       for childID, optionID in pairs(option.references) do
         local childData = data[childID]
-        if index == #childData.authorOptions then return true end
+        if optionID >= #childData.authorOptions then return true end
       end
     end, function()
       for childID, optionID in pairs(option.references) do
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -57,6 +57,7 @@ local optionClasses = {
   range = "simple",
   color = "simple",
   select = "simple",
+  multiselect = "simple",
   description = "noninteractive",
   space = "noninteractive",
 }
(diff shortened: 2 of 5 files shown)


WeakAuras/Transmission.lua
```diff
@@ -1677,6 +1676,7 @@ function WeakAuras.Import(inData, target)
     local received = StringToTable(inData, true)
     if type(received) == 'string' then
       -- this is probably an error message from LibDeflate. Display it.
+      wipe(pendingData)
```
InfusOnWoW: So, the clean up happens in a hook to OnHide, which is not called in that case because the tooltip is already visible.  (Note that your fix is also incomplete, there's more cleanup to be done.) 

We have a few options:

* We should at least move the cleanup into ShowTooltip, e.g. either directly calling the cleanup, or indirectly via HideUIPanel

This solves all problems where we show the ItemRefTooltip. But if i recall correctly ItemRefTooltip is a global shared tooltip?

So:
* We might need to look at what other methods we might need to hook with a cleanup,
* Or figure out a way to not use a global shared tooltip.

emptyrivers: ShowTooltip happens after ShowDisplayTooltip in the case of a good import. So we can’t move the cleanup there without also changing when data is populated, or otherwise restructuring (otherwise we’d cleanup the very data we want to use).

What code did I forget to include?

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -704,6 +705,178 @@ local typeControlAdders = {
...
+      values = defaultValues,
+      get = function(_, k)
+        return option.default and option.default[k]
+      end,
+      set = function(_, k, v)
+        if option.references then
+          for childID, optionID in pairs(option.references) do
+            local childData = data[childID]
+            local childOption = childData.authorOptions[optionID]
+            childOption.default[k] = v
```
InfusOnWoW: You might want to add a check here that the key is correct for that child option

I guess it's not needed since it'll be removed again by the validate.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -704,6 +705,178 @@ local typeControlAdders = {
...
+            WeakAuras.Add(data)
+            WeakAuras.ReloadTriggerOptions(data)
+          end
+        end,
+        image = "Interface\\AddOns\\WeakAuras\\Media\\Textures\\delete",
+        imageWidth = 24,
+        imageHeight = 24
+      }
+    end
+    args["option" .. i .. "newvaluespace"] = {
```
InfusOnWoW: Not really a space :)

emptyrivers: This follows the convention in other parts of the codebase, where a disabled checkbox on the left is called a space. Also it’s the same as the select code above.

WeakAuras/Transmission.lua
emptyrivers: I am pretty sure I did mention in the commit message that this isn't a great fix. But anyways, that extra code isn't necessary in this case because ShowTooltip actually handles the case, so long as pendingData is empty.

Since this is such a minor issue, I think I will retract the fix for now and look for a better solution, so that the multiselect can be merged.

Causese:
https://github.com/WeakAuras/WeakAuras2/issues/1075

and unticking a multiselect toggle outside of author mode also unticks the default inside author mode

emptyrivers:
I can't reproduce that.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -704,6 +705,178 @@ local typeControlAdders = {
...
+        defaultValues[i] = "|cFF4080FF" .. L["Value %i"]:format(i)
+      else
+        defaultValues[i] = v
+      end
+    end
+    args["option" .. i .. "default"] = {
+      type = "multiselect",
+      width = WeakAuras.normalWidth,
+      name = L["Default"],
+      -- name = name(data, option, "default", L["Default"]),
```
Stanzilla: remove?

emptyrivers: Ah, I forgot about that. I was lazy and didn't want to deal with figuring out how to blue the name, and forgot to fix that.

emptyrivers: But, on further thought, the sub-values do the trick i think. So I'll just remove the commented code.

Issue #708, opened 2018-09-05
"Dynamic information from first active" icon prioritizes Trigger 1 with Pet abilities


amiyuy:
**Describe the bug**
WA is using the first pet ability in the first trigger for the Dynamic information for Automatic Icon instead of the one that's available right now.  I have Nether Shock (Trigger 2) available on my current pet, it shows Nature's Grace's icon from Trigger 1.

If I move the current pet's ability to Trigger 1, it shows that icon.

If I start with the spell from Trigger 2, it uses that icon, but then if I switch to the spell from Trigger 1, it stays with that one, even if I switch back to the spell from Trigger 2.

Pulling out a pet that has the same type of ability (ie. Moth with Serenity Dust) will activate the aura, even though Serenity Dust is not part of the triggers.

I'm not sure if this is pet ability weirdness or because of Custom Activation function, or what.

**To Reproduce**
1. Import aura 
2. Start with Nether Ray pet
3. Icon should be Nether Shock
4. Dismiss and summon Stag
5. Icon will switch to Nature's Grace
6. Switch to Nether Ray
7. Icon will stay as Nature's Grace when it should switch to Nether Shock
8. Icon will stay as Nature's Grace until relog

**Which version of WeakAuras are you using?**
2.7.5

**Screenshots**
Video: https://i.imgur.com/QVYRxxK.gifv

**Additional context**
```
[...]

PR #1081, opened 2019-01-03
Smallerpatches

InfusOnWoW:
# Description

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -1454,8 +1454,8 @@ do
     end
   end
 
-  function WeakAuras.GetSpellCooldown(id, ignoreRuneCD, showgcd)
-    if (not spellKnown[id]) then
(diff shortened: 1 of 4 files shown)


InfusOnWoW:
I'm annoyed at adding such a minor option to the Cooldown Progress trigger, but I don't see a way to fix this without reopening #708.

Both are argueably valid use cases, which the trigger should support.

emptyrivers:
Perhaps it would be better to add a combat res prototype trigger, instead of adding an obscure option to cool down progress for one use case.

Like, I wouldn’t look for a combat res timer in the cooldown progress section if I wasn’t on a class that had a combat res.

InfusOnWoW:
I'm pretty sure this isn't the only spell that behaves in that way, even though I can't point to a immediate use case beyond combat res

PR #1085, opened 2019-01-05
fix wago menu

mrbuds:
fix wago menu at same position for auras and groups
rename "Ignore this Update" to "Skip this Version"
rename "Stop ignoring this Update" to "Don't skip this Version"
show skip version option, and update option, only if aura as update data

prev

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -871,12 +871,12 @@ local methods = {
     self.menu = {
       {
         text = L["Rename"],
-        notCheckable = 1,
+        notCheckable = true,
         func = self.callbacks.OnRenameClick
       },
       {
         text = L["Copy settings..."],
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -936,6 +936,16 @@ local methods = {
       notCheckable = 1,
       func = function() WeakAuras.ExportToTable(data.id) end
     });
+
+    if WeakAurasCompanion then
+      tinsert(self.menu, {
+        text = '|TInterface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon:0|t' .. L["Wago Update"],
+        notCheckable = 1,
+        hasArrow = true,
```
emptyrivers: That looks like it should be a boolean value. Does the API really require numbers for this?

mrbuds: fixed with  #44500d4

PR #1099, opened 2019-01-14
New glow conditions 

1ps3:
There was a request in the discord to be able to use the glow frame picker inside of conditions and it got my thinking about how glow is handled at the moment. So this PR I hope addresses it. At the moment there's 4 different options for glow in the coniditions dropdown, this PR merges all 4 into a single glow condition that allows the user to pick the glow style, colour and action as well as adding another condition for glowing something that isn't the current aura. It's not 100% done yet, but I wouldn't mind some thoughts. 
wow_e7aju4jo56

Also I fixed something that was making the sound/chat message conditions not show a name on the first dropdown.


WeakAurasOptions/ConditionOptions.lua
```diff
@@ -438,12 +438,138 @@ local function addControlsForChange(args, order, data, conditionVariable, condit
...
+    args["condition" .. i .. "value" .. j .. "glow_frame_picker"] = {
+      type = "execute",
+      width = WeakAuras.doubleWidth*.3,
+      name = blueIfNoValue2(data, conditions[i].changes[j], "value", "glow_frame_picker", L["Pick"], L["Choose Frame"]),
+      desc = descIfNoValue2(data, conditions[i].changes[j], "value", "glow_frame_picker", propertyType, L["hoose Frame"]),
+      order = order,
+      func = function()
+        if type(data.conditions[i].changes[j].value) == "table" and data.conditions[i].changes[j].value.frame_input then
+          --WeakAuras.StartFrameChooser(WeakAuras.GetData(data.conditions[i].changes[j].value.frame_input), {"conditions", i, "changes", j, "value", "frame_input"});
+          print(WeakAuras.GetData(data.conditions[i].changes[j].value.frame_input))
```
emptyrivers: Another debug print

WeakAuras/RegionTypes/RegionPrototype.lua
```diff
@@ -123,6 +134,63 @@ function WeakAuras.regionPrototype.AddProperties(properties, defaultsForRegion)
   end
 end
 
+local function SetFrameGlow(self, options)
```
emptyrivers: You’re duplicating code. The better option is to expose the original code somehow and reuse it if you can. That way if we ever need to fix a bug in this we only have to do it once.

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -438,12 +438,138 @@ local function addControlsForChange(args, order, data, conditionVariable, condit
...
+      elseif (propertyType == "animate") then 
+    local animation_types = {
+      pulse = L["Pulse"],
+      bounce = L["Bounce"],
+      flip = L["Filp"],
+      shake = L["Shake"],
+      spin = L["Spin"],
+      spiral = L["Spiral"],
+    }
+    args["condition" .. i .. "value" .. j .. "animation_type"] = {
```
emptyrivers: Animation??? Does this really belong in this pull request?

Conditional animations are a much bigger scope then this, I think.

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -472,7 +598,7 @@ local function addControlsForChange(args, order, data, conditionVariable, condit
       type = "select",
       width = WeakAuras.normalWidth,
       values = WeakAuras.sound_types,
-      name = blueIfNoValue2(data, conditions[i].changes[j], "value", "sound", L["Differences"]),
+      name = blueIfNoValue2(data, conditions[i].changes[j], "value", "sound", L["Sound"], L["Sound"]),
```
emptyrivers: This seems out of scope for the ticket you’re trying to address, no? 

If you want to make changes like this, don’t forget to mention that you are making them.

mrbuds:
About the changes to the Glow and Cooldown conditions condition, i think the right way to do it would be to show all options we have in display tab
preview
and same for cooldowns
Which might need some refactoring and is kind of a project on it's own to be future proof

PR #1124, opened 2019-01-24
reset glow on size change

### LINKED ISSUE #1123: Using Conditions to create an Action Button Glow and on an already existing Action Button Glow does not behave as expected when changing the size of an icon aura.
**Describe the bug**

Changing the size of an icon through conditions causes Action Button Glow to misbehave.
Expected Behavior: The Action Button Glow should match the size of the icon.
Actual Behavior:
Case 1 - Both parts of glow match height as expected. The animated portion of the glow matches the original width of the aura. The static portion appears to match 75% of the new width of the aura.
Case 2 - Incredibly inconsistent behavior. Sometimes behavior matches Case 1. Sometimes Case 3 (after Case 3 has been reloaded). Unable to consistently replicate differences in behavior.
Case 3 - Glow is unchanged, matches original size of aura. However, when reloading the animation portion remains the original size of the aura and the static portion appears to match 75% of the new size.

**To Reproduce**
 
The auras supplied below may be imported and the aberrant behavior will be immediately evident.

Alternatively:
1. Create an icon aura
2. Set trigger to Status - Condition - Always Active Trigger

Case 1:
3. Create a condition that does the following:
-a. If Trigger 1 - Active - True
-b. Then:
-c. Height (value signficantly higher than original, I used 200 throughout)
-d. Width
-e. Set Glow Visibility - Checked
4. Close WA Options and observe.

Case 2:
3. Create a condition that does the following:
-a. If Trigger 1 - Active - True
-b. Then:
-c. Height
-d. Width
4. Create an additional condition that does the following:
-a. If Trigger 1 - Active - True
-b. Then:
-c. Set Glow Visibility - Checked
5. Close WA Options and observe.

Case 3:
3. Enable Show Glow Effect Glow under Display (default settings).
4. Create a condition that does the following:
-a. If Trigger 1 - Active - True
-b. Then:
-c. Height
-d. Width
5. Close WA Options and observe.

**Screenshots**

Both screenshots show: Baseline - Case 1 - Case 2 - Case 3. Along with a 64px wide and 150px wide aura to showcase the sizes.

https://i.imgur.com/guVHGgO.png
https://i.imgur.com/1o0I0uu.png

Case ~2~ 1? behaving oddly:
https://i.imgur.com/eLqsf15.png
Normally that was Case 2. I guess it's more inconsistent than I thought.

Tested with only WeakAuras officially related addons enabled (WeakAuras 2, WeakAuras Options, WeakAuras Templates, WeakAuras Companion, and WeakAuras Model Paths)

2.11.1

Unknown.

**Additional context**

In the event that this was an "always active trigger" problem, I did recreate the issue with an always showing icon that only applied conditions upon applying an aura and the strange behavior still occurred, even in case 3.

https://wago.io/q4tBqpvAu

Contains a full suite of a baseline aura, all 3 cases in order, and 2 additional auras to compare size of misbehaving glows.

emptyrivers:
# Description

Fixes #1123

## Type of change

# How Has This Been Tested?

Follow the steps to reproduce in #1123. The expected behavior is now the actual behavior.

# Checklist:

# Additional Work

The starting animation will play again on size change, which is unfortunate. But that is a problem best solved upstream in LibCustomGlow.

### DIFF
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -509,7 +509,6 @@ local function modify(parent, region, data)
   function region:UpdateSize()
     local width = region.width * math.abs(region.scalex);
     local height = region.height * math.abs(region.scaley);
-
     region:SetWidth(width);
     region:SetHeight(height);
     if MSQ then
@@ -520,6 +519,9 @@ local function modify(parent, region, data)
     icon:SetAllPoints();
 
     region:UpdateTexCoords();
+    if region.glow then
+      region:SetGlow(true);
+    end
   end
 
   function region:UpdateTexCoords()
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -1617,7 +1617,7 @@ function WeakAuras.ShowDisplayTooltip(data, children, matchInfo, icon, icons, im
       else
         data.desc = nil;
       end
-      WeakAuras.ShowDisplayTooltip(data, children, nil, nil, import, nil, "desc");
+      WeakAuras.ShowDisplayTooltip(data, children, nil, nil, nil, import, nil, "desc");
       if(WeakAuras.GetDisplayButton) then
         local button = WeakAuras.GetDisplayButton(data.id);
         if(button) then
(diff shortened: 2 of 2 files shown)


InfusOnWoW:
I think some addons (that is omnicc) can be configure to play a animation on start/finish glow. This needs to be tested with that and whether that'll look good.

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -509,7 +509,11 @@ local function modify(parent, region, data)
   function region:UpdateSize()
     local width = region.width * math.abs(region.scalex);
     local height = region.height * math.abs(region.scaley);
-
+    local glowing;
+    if region.glow then
+      region:SetGlow(false);
```
InfusOnWoW: I briefly discusses this with @Dooez in discord today. He suggested that the best way is to fix the show method of the glow to take size changes into account.

So we don't need to hide the glow here.

Stanzilla:
@emptyrivers follow up PR for the docs Infus mentioned please

PR #1182, opened 2019-02-12
add checkbox for streamer mode

mrbuds:
# Description
Checking it hide urls from tooltips

## Type of change

# How Has This Been Tested?
- [x] tooltip on mouseover an aura button
- [x] import tooltip

# Checklist:

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -1377,7 +1377,7 @@ function WeakAuras.ShowDisplayTooltip(data, children, matchInfo, icon, icons, im
     tinsert(tooltip, {1, "\""..data.desc.."\"", 1, 0.82, 0, 1});
   end
 
-  if hasUrl then
+  if hasUrl and not WeakAuras.savedVars.db.streamerMode then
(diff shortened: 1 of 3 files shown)


mrbuds:
@emptyrivers
> i have no personal stake in this, but philisophically, I'd prefer not to encourage people hiding their creations

@Stanzilla 
> I'm indifferent about this
> Leaning to it being something wago should care about
> Because then we could also handle it properly in the client

@InfusOnWoW last opinion about it was
> I don't think we should add it, feels like a rather strange option...
> but I can be bought, just tell them that the team needs a few new MSI 2080TIs

I personally doesn't care if it's merged or closed, but i'd like a MSI 2080TI too

So who has spare 2080TIs to share ?

InfusOnWoW:
I would prefer if wago had a permission system to a checkbox, and I agree with rivers that I would prefer people to default to sharing their stuff. 

That's something that is very much at the base of the sucess of the community around WeakAuras. Freely sharing what you created and letting others take inspiration is how most people are introduced to WA, and thus imho hiding should be the exception.

PR #1194, opened 2019-02-21
fix padWithLinebreaks by switching to upstream FAIAP

mrbuds:
# Description
Fix adding trailing line break when editing custom code

### DIFF
--- a/WeakAurasOptions/ForAllIndentsAndPurposes.lua
+++ b/WeakAurasOptions/ForAllIndentsAndPurposes.lua
@@ -976,11 +976,11 @@ if not IndentationLib.revision or revision > IndentationLib.revision then
     local linebreak = stringbyte("\n")
     local space = stringbyte(" ")
     function lib.padWithLinebreaks(code)
-        local len = stringlen(code) + 1
+        local len = stringlen(code)
         local breakfound = 0
+        local byte
         while true do
-            len = len - 1
-            local byte = stringbyte(code, len)
+            byte = stringbyte(code, len)
             if byte == linebreak then
                 breakfound = breakfound + 1
                 if breakfound == 2 then return code, false end
@@ -988,6 +988,7 @@ if not IndentationLib.revision or revision > IndentationLib.revision then
                 if breakfound == 0 then return code .. "\n\n", true end
                 if breakfound == 1 then return code .. "\n", true end
             end
+            len = len - 1
         end
     end
 
--- a/WeakAurasOptions/ForAllIndentsAndPurposes.lua
+++ b/WeakAurasOptions/ForAllIndentsAndPurposes.lua
@@ -1,5 +1,5 @@
 -- For All Indents And Purposes
-local revision = 21
+local revision = 22
 -- Maintainer: kristofer.karlsson@gmail.com
 
 -- For All Indents And Purposes -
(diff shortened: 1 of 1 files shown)


Stanzilla:
Did people actually complain about this?

InfusOnWoW:
yes people notice extra ever increasing line breaks, but this looks strange, the code explicitly checks whether the code ends in \n\n, and only adds those if they are missing so this should only ever add 2

mrbuds:
I'll see if i can come with a fix to the function.
Current problem is :
- it make sure the editbox always finish with 2 \n as you type
- When you type a new block keyword all text after this keyword is indented, the 2 \n included, so last character is \n but the previous one is a space, so function add a \n at the end, which gets itself indented etc ...

mrbuds:
padWithLinebreaks restored and fixed to handle indented line breaks

WeakAurasOptions/ForAllIndentsAndPurposes.lua
```diff
@@ -974,16 +974,21 @@ if not IndentationLib.revision or revision > IndentationLib.revision then
 
     -- returns the padded code, and true if modified, false if unmodified
     local linebreak = stringbyte("\n")
+    local space = stringbyte(" ")
     function lib.padWithLinebreaks(code)
-        local len = stringlen(code)
-        if stringbyte(code, len) == linebreak then
-            if stringbyte(code, len - 1) == linebreak then
-                return code, false
+        local len = stringlen(code) + 1
```
emptyrivers: I am nitpicking, but this number doesn’t really represent the length of the string anymore.

WeakAurasOptions/ForAllIndentsAndPurposes.lua
```diff
@@ -974,16 +974,21 @@ if not IndentationLib.revision or revision > IndentationLib.revision then
...
+        local breakfound = 0
+        while true do
+            len = len - 1
+            local byte = stringbyte(code, len)
```
emptyrivers: This might cause issues if the code is < 2 bytes long. I’m not sure how stringbyte handles inputs < 1.

emptyrivers: A simple fix would be to check if the string is extremely short, and handle that as a special case.

mrbuds: it's working as intended on an empty custom code editbox

InfusOnWoW:
So this looks like it is indeed a bug in that library and the fix is good.

But, it's a library and we overwriting everyone elses version if we ship this.

@Stanzilla how do we proceed ?

Stanzilla:
So I don't actually know of any other addon that uses it and there is no repo for it either, it was just passed on as a "snippet" over the years. So I'd say we can roll with it and see if anyone complains, if they do, we can change the name of the lib.

InfusOnWoW:
Hmm, that's not ideal, but okay let's do that.

emptyrivers:
The simple answer would be to just make this fork local.

Stanzilla:
We just discovered that WoWLua, Plater and Details use it so yeah

Stanzilla:
Maybe we should just make an official repo for it and let the other authors know

emptyrivers:
That’s fine with me. Do we have permission to republish like that?

mrbuds:
Details & Plater use the same version as we do, with the fix on getCaretPos(editbox) we did
WoWLua seems to use an outdated version, their revision number is nil and i think no one saw that because WeakAuras load before WoWLua

Stanzilla:
Hey @grizzm0 what's the license on AllIndentsAndPurposes?

grizzm0:
@Stanzilla Tagged the wrong person? ;)

Stanzilla:
@grizzm0 certainly possible, I searched for the email address on GitHub and you came up, sorry!

grizzm0:
@Stanzilla No worries. Always wondered who managed to steal that email from me! :o I even got an odd spelling of Kristofer. ;)

InfusOnWoW:
@spkrka I believe you are the original author of ForAllIndentsAndPurposes, right?

spkrka:
Yes, that's me (but on my personal account @krka )

Here's the original repository: https://github.com/krka/ForAllIndentsAndPurposes
It appears my original repo is several revisions behind this version - perhaps I should try to reintegrate it.

Code change looks reasonable, and I can take a deeper look later. A simple thing to change could be to replace `while true do` with `while len >= 0`

(Sorry about the name steal @grizzm0 but I suspect I was in fact born before you :-) )

krka:
I've made the following PR in the original repository now: https://github.com/krka/ForAllIndentsAndPurposes/pull/4
It solves the same problem as this PR, but also adds some test cases for it. (And applies both for space and tab character)

Issue #1208, opened 2019-02-24
Improve custom code error handling


emptyrivers:
**Is your feature request related to a problem? Please describe.**

When custom code has an error, the error message given to the standard error handler isn’t very descriptive of *which* code block ran into an error. If you aren’t familiar with the kinds of stack traces you get, you might not know exactly where things went wrong.

**Describe the solution you'd like**
Since we wrap every custom script in xpcall, we can do whatever we like with the error message. We could add an “error cache” to WeakAuras (along the lines of what bugsack does, but individual to each script). We could direct users to contact the authors of the specific aura about the error they encountered, we could open Options to the specific line which had an error, even. There are lots of possibilities.

**Describe alternatives you've considered**

I haven’t really considered much yet, I just want to get this thought out of my brain and somewhere where others can contribute.

oratory:
It would be slick if the Companion could track and record errors, with the option to submit to Wago. Wago could then add an "Error reports" panel and alert the aura's author.

cont1nuity:
Having (semi) automatic bug reports to the authors is a dangerous thing in my opinion.
Imagine someone has had their way with the WA and change the code by just a bit and it fired errors constantly. As a developer, I would not accept that bug report. Or at least I would want to know in detail what was changed.
In this case it would be necessary to to track changes made to the WA and only hand in reports if the WA (code, usually) was not changed.
In genereal this is a nice idea, though. But I think it should be opt-in as dev/creator on wago to prevent spamming.

emptyrivers:
As it happens, we do have a feature request to track what has changed since the last import (or at least, to persist this original version), and I intend to implement that request eventually.

Heckate:
Sorry but I am not a coder - just a user and student of medicine- I repeatedly get undefined error in the companion and even in the addon- unable to install 'quaking assist' (WoW BFA 8.3) - am considering deleting & finding/installing manually- but is there is there a fix already or a better notion of how to ?

mrbuds:
@Heckate Hi, your problem is not related to this ticket, and the description you do doesn't allow us to help you, please come on discord at https://discord.gg/wa2 and explain with more details your issue(s).

InfusOnWoW:
This is now more or less done.

PR #1209, opened 2019-02-25
BT2upgrade icon update

mrbuds:
# Description
fix icon not hiding after converting to BT2 or on removing trigger
show trigger on icon click

## Type of change

## To Do
Need to click a second time to show the "trigger" tab when triggers options were not loaded (ace3 bug??)

### DIFF
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -633,6 +633,7 @@ function WeakAuras.CreateFrame()
   frame.FillOptions = function(self, optionTable, selected)
     AceConfig:RegisterOptionsTable("WeakAuras", optionTable);
     AceConfigDialog:Open("WeakAuras", container);
+    -- TODO: remove this once legacy aura trigger is removed
     if selected then
       container.content.obj.children[1]:SelectTab(selected)
     end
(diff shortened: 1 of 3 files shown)


WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -630,9 +630,12 @@ function WeakAuras.CreateFrame()
   unloadedButton:SetViewDescription(L["Toggle the visibility of all non-loaded displays"]);
   frame.unloadedButton = unloadedButton;
 
-  frame.FillOptions = function(self, optionTable)
+  frame.FillOptions = function(self, optionTable, selected)
     AceConfig:RegisterOptionsTable("WeakAuras", optionTable);
     AceConfigDialog:Open("WeakAuras", container);
+    if selected then
+      container.content.obj.children[1]:SelectTab(selected)
```
InfusOnWoW: I don't like that. 

I think it's brittle, although it's unlikely that anything will change that breaks it.
But also it's a code path which we won't ever test again, because it's a rather exceptional one.

So imho we should at least mark this with -- TODO: remove

emptyrivers: I just want to note that this might be handy to investigate for #1208.

mrbuds:
Added 3 "TODO: remove"

PR #1216, opened 2019-03-01
Smallerpatches

InfusOnWoW:
# Description

Fixes #(issue)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -1753,19 +1753,8 @@ do
       startTimeCooldown, durationCooldown = 0, 0
     end
 
-    local spellcount = GetSpellCount(id);
-    -- GetSpellCount returns 0 for all spells that have no spell counts, so we only use that information if
-    -- either the spell count is greater than 0
-    -- or we have a ability without a base cooldown
-    -- Checking the base cooldown is not enough though, since some abilities have no base cooldown, but can still be on cooldown
-    -- e.g. Raging Blow that gains a cooldown with a talent
-    if (charges == nil and spellcount > 0) then
-      charges = spellcount;
-    end
-
-    local basecd = GetSpellBaseCooldown(id);
     local onNonGCDCD = durationCooldown and startTimeCooldown and durationCooldown > 0 and (durationCooldown ~= gcdDuration or startTimeCooldown ~= gcdStart);
-    if ((basecd and basecd > 0) or onNonGCDCD) then
+    if (onNonGCDCD) then
       cooldownBecauseRune = runeDuration and durationCooldown and abs(durationCooldown - runeDuration) < 0.001;
       unifiedCooldownBecauseRune = cooldownBecauseRune
     end
@@ -1789,6 +1778,17 @@ do
         startTime, duration = startTimeCharges, durationCharges
         unifiedCooldownBecauseRune = false
       end
+    end
+
+    local spellcount = GetSpellCount(id);
+    local basecd = GetSpellBaseCooldown(id);
+    -- GetSpellCount returns 0 for all spells that have no spell counts, so we only use that information if
+    -- either the spell count is greater than 0
+    -- or we have a ability without a base cooldown
+    -- Checking the base cooldown is not enough though, since some abilities have no base cooldown, but can still be on cooldown
+    -- e.g. Raging Blow that gains a cooldown with a talent
+    if (charges == nil and not onNonGCDCD and (spellcount > 0 or not basecd or basecd == 0)) then
+      charges = spellcount;
     end
 
     return charges, maxCharges, startTime, duration, unifiedCooldownBecauseRune,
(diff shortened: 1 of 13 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -2772,6 +2772,14 @@ function WeakAuras.Modernize(data)
     end
   end
 
+  -- Version 4 was introduced Februar 2019 in BfA
```
Stanzilla: 4 should be 12?

InfusOnWoW: Indeed.

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasTwoColumnDropDown.lua
```diff
@@ -0,0 +1,210 @@
...
+
+    widget:AddChild(firstDropdown)
+    widget:AddChild(secondDropDown)
+
+    local OnFirstDropdownValueChanged = function(self, event, value)
+      local displayName = widget.userdata.firstList[value]
+      local treeValue = widget.userdata.tree[displayName]
+      if type(treeValue) == "table" then
+        local secondList = {}
+        for displayName, _ in pairs(treeValue) do
```
emptyrivers: You could simplify this to `for displayName in pairs(treeValue) do` if you like.

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasTwoColumnDropDown.lua
```diff
@@ -0,0 +1,210 @@
...
+    treeValue = treeValue[displayName2]
+    if not treeValue then
+      return nil
+    end
+    return treeValue
+  end,
+  ["SetList"] = function(self, list, order, itemType)
+    local tree = {}
+    for key, displayName in pairs(list) do
+      local pos = string.find(displayName, '/', 1, true)
```
emptyrivers: Hmm. This is a bit magical to me, and also exposes us to an (admittedly unlikely) issue with translations. If the name of e.g. `Glow` gets translated to contain a slash, then this could break the SetList method at least visually.

Perhaps instead of encoding the structure of a property change in the displayName, we could create a table in each property which contains the value for the first dropdown, and a value for the second (if it exists).

WeakAuras/GenericTrigger.lua
```diff
@@ -2130,6 +2130,37 @@ do
   end
 end
 
+local watchUnitChange
```
emptyrivers: You mentioned that there is a possible refactoring opportunity in your commit message. What did you have in mind?

mrbuds:
About the changes to the Glow and Cooldown conditions condition, i think the right way to do it would be to show a group of options taken from the display tab when you select "glow" or "cooldown"
preview
and same for cooldowns

This is a project on it's own to be future proof, and i think animations could be moved in display and condition could have an animation option that would show a group of options too

InfusOnWoW:
Integrating animations into conditions needs a lot of backend work, the ui is frankly the smallest part of that.

I still don't like having a condition where you have to reset all glow related stuff. The usual way it's used is to set up the specific glow look you want in the display, and then toggle it on/off via a condition.

But that's also out of scope of the PR.

mrbuds:
sometime the left widget disappears
![preview] (https://i.imgur.com/6TQpmmR.gif)

mrbuds:
couldn't find a bug in conditions anymore lgtm

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasTwoColumnDropDown.lua
InfusOnWoW: What I want to avoid is having the properties be a tree structure, as that makes the whole handling everywhere much more complex. I thik I might be able to cope with displayName be a array though.

WeakAuras/GenericTrigger.lua
InfusOnWoW: Currently, triggers subscribe to events. But for some triggers, the events fire a lot more often than the triggers actually care about. E.g. SPELL_COOLDOWN_CHANGED is fired whenever a single spell cd changes, but triggers for all spell watch the same signal, and thus evaluate.

That's clearly not needed.

This sidesteps that issue at the cost of creating lots of fake events, but the problem is much more general.

mrbuds:
found a tiny regression
make an aura with trigger: status > unit characteristics > unit=target, unitisunit=specific unit=boss1
now go like in eye of azshara dungeon, target boss, engage fight => aura doesn't show
same thing with master => aura show when encounter start

InfusOnWoW:
@mrbuds, @emptyrivers I pushed two fixes for the two issues you found, which I'll later squash.

WeakAuras/GenericTrigger.lua
emptyrivers: Ok, thanks

PR #1227, opened 2019-03-05
fix sharing of auras with a closing bracket in name

### LINKED ISSUE #1225: Weakauras not importing via chat caused by special symbols.
**Describe the bug**

Whenever I try to share a weakaura that has special symbols in it's name such as [ ] in case, it displays an error "Requested display does not exist"

There are no errors in bug-grabber I use it all the time to analyze this stuff in case shit hits the fan. 

**To Reproduce**
 

**Screenshots**

https://i.gyazo.com/dfeb36d5793c885ea6da74a794b8ff4e.png

Yes, base blizz + wa's

2.11.6

No

**Additional context**

mrbuds:
# Description
Fixes #1225

## Type of change

# How Has This Been Tested?
- [x] Share aura "[hello]"
- [x] Share group "normal name"

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -145,7 +145,7 @@ local function filterFunc(_, event, msg, player, l, cs, t, flag, channelId, ...)
   local remaining = msg;
   local done;
   repeat
-    local start, finish, characterName, displayName = remaining:find("%[WeakAuras: ([^%s]+) %- ([^%]]+)%]");
+    local start, finish, characterName, displayName = remaining:find("%[WeakAuras: ([^%s]+) %- (.*)%]");
     if(characterName and displayName) then
       characterName = characterName:gsub("|c[Ff][Ff]......", ""):gsub("|r", "");
       displayName = displayName:gsub("|c[Ff][Ff]......", ""):gsub("|r", "");
@@ -546,7 +546,7 @@ local receivedData;
 hooksecurefunc("ChatFrame_OnHyperlinkShow", function(self, link, text, button)
   buttonAnchor:Hide()
   if(link == "weakauras") then
-    local _, _, characterName, displayName = text:find("|Hweakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- ([^%]]+)%]|h");
+    local _, _, characterName, displayName = text:find("|Hweakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- (.*)%]|h");
     if(characterName and displayName) then
       characterName = characterName:gsub("|c[Ff][Ff]......", ""):gsub("|r", "");
       displayName = displayName:gsub("|c[Ff][Ff]......", ""):gsub("|r", "");


WeakAuras/Transmission.lua
```diff
@@ -145,7 +145,7 @@ local function filterFunc(_, event, msg, player, l, cs, t, flag, channelId, ...)
   local remaining = msg;
   local done;
   repeat
-    local start, finish, characterName, displayName = remaining:find("%[WeakAuras: ([^%s]+) %- ([^%]]+)%]");
+    local start, finish, characterName, displayName = remaining:find("%[WeakAuras: ([^%s]+) %- (.*)%]");
```
emptyrivers: This fails on the following chat message:

> Hey guys, look at these great auras: [WeakAuras: EmptyRivers - Aura 1] [WeakAuras: EmptyRivers - Aura 2]

The first link will request an aura named `Aura 1] [WeakAuras: EmptyRivers - Aura 2]`. I'm not entirely sure how the second link will behave, though.

WeakAuras/Transmission.lua
```diff
@@ -546,7 +546,7 @@ local receivedData;
 hooksecurefunc("ChatFrame_OnHyperlinkShow", function(self, link, text, button)
   buttonAnchor:Hide()
   if(link == "weakauras") then
-    local _, _, characterName, displayName = text:find("|Hweakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- ([^%]]+)%]|h");
+    local _, _, characterName, displayName = text:find("|Hweakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- (.*)%]|h");
```
emptyrivers: This also fails when somebody sends this chat message (for the same reason):
> Hey guys, look at these great auras: [WeakAuras: EmptyRivers - Aura 1] [WeakAuras: EmptyRivers - Aura 2]

emptyrivers:
Well, you might be able to get around the fact that there are pretty much always tricky strings which defeat your patterns, if we can use a hyperlink i.e. `|HWeakAuras:name:displayName|h`. I have no idea if this is something which we are allowed to do or not, though.

WeakAuras/Transmission.lua
mrbuds: i have always used weakauras links with nothing else than 1 link, i didn't know 2 links in same line was supported tbh

mrbuds:
TotalRp3 read the hyperlink by hooking ChatFrame_OnHyperlinkShow
https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/modules/ChatLinks/ChatLinks.lua#L228

This part format the chat message
https://github.com/Ellypse/Total-RP-3/blob/master/totalRP3/modules/ChatLinks/ChatLinks.lua#L129

WeakAuras/Transmission.lua
InfusOnWoW: It also fails if the chat message has a ] in it after the link. So "Guys, install this: [WeakAuras: EmptyRivers - Aura 1]  [].

The "proper" way to do this is to escape any ] in the name of the aura and to properly parse the msg.

Stanzilla:
From the code comments it reads like shift clicking is not supported with that method?

mrbuds:
I fear all fix i could try to make will add another bug for close to no gain at the end .. so i guess it's better if just close this and restore won't fix label on original issue

PR #1237, opened 2019-03-08
Little stuff

Nightwarden24:
# Description

PR contains a little changes aimed to improving the addon:
1. support the Fira Mono font in the Code Review frame for Ru locale (It has cyrillic characters);
2. use Blizzard strings instead of localized for some types
3. replacement of functions with new api to get race and class data

P.S. It would be desirable to review PR and merge it before 8.1.5 patch (allied race)

## Type of change

In my opinion, there's no desired option here (enhancement). So, I have chosen the most similar one.

# Checklist:


Stanzilla:
> As I recall, we use .luacheckrc for this kind of thing so this comment probably can be removed.

That's probably part of a larger cleanup where we move all inline rules to the rc file.

emptyrivers:
This pull request also contains a large number of changes to translated phrases. Have you confirmed that the result is for the most part the same or better than what we have currently?

Nightwarden24:
Official (blizzard) terminology is preferable to user/custom translation, isn't it? ​​You can see a content of strings in different languages [here](https://www.townlong-yak.com/framexml/live/GlobalStrings.lua). 
Actually, it's enough to look at where these strings are used in Blizzard UI code. But it's not always easy to find them (string operations are often used). 
And yea, I confirm that the strings contain what we need

InfusOnWoW:
Can you make a screenshot for the combatlog_raid_mark_check_type change? That is how it looks for you without your change, and how it looks with your change?

On my computer they pretty much look indentical.

emptyrivers:
I don’t think I would say that blizzard terminology is always preferable. It’s certainly easier to maintain, but there is not always an appropriate phrase. This is particularly true when you get into the subtleties of how things actually are translated. A 2 syllable word in English might get translated into one of several different complex phrases, depending on the context, and it’s not always obvious from the name of the global string how it is translated.

But thanks for confirming, anyways.

Nightwarden24:
@InfusOnWoW, you mean the icon size, right?

1920x1080, fullscreen, addon is off, UI scale is off

without change


with change

Nightwarden24:
> I don’t think I would say that blizzard terminology is always preferable. It’s certainly easier to maintain, but there is not always an appropriate phrase. 

Oh, I would argue with you. There are so many arguments, starting with the fact that almost nobody translates or people translate phrases without understanding the context (without considering the code where it is used), and ending with the fact that WA has a not quite correct translation model/system. But this is a topic for another discussion.

No one said about using Blizzard strings everywhere and at all times. But standard/basic things should have and use a common/unified terminology.

InfusOnWoW:
Yeah I meant the icon size, I think that change is fine though.

emptyrivers:
Hey @Nightwarden24, it looks like you were testing these changes on the PTR, and not on the live version of World of Warcraft. Your change to use C_ClassColor does not work on the live version, since that particular API package is new in the PTR. We have reverted the commit which introduced that change, now. If you wish to resubmit them (without the breaking change), open another pull request.

We should have caught this during review, but please double check that your changesets work on the live version of WoW.

Stanzilla:
(or mention it, still it's out fault for not testing properly)

PR #1243, opened 2019-03-10
custom grow & sort

emptyrivers:
This is a preview for custom layout and sort options. @mrbuds this is not a true pull request.

### DIFF
--- a/WeakAurasOptions/OptionsFrames/TextEditor.lua
+++ b/WeakAurasOptions/OptionsFrames/TextEditor.lua
@@ -243,11 +243,12 @@ local function ConstructTextEditor(frame)
     end
   end);
 
-  function group.Open(self, data, path, enclose, multipath, reloadOptions)
+  function group.Open(self, data, path, enclose, multipath, reloadOptions, setOnParent)
     self.data = data;
     self.path = path;
     self.multipath = multipath;
     self.reloadOptions = reloadOptions;
+    self.setOnParent = setOnParent;
     if(frame.window == "texture") then
       frame.texturePicker:CancelClose();
     elseif(frame.window == "icon") then
@@ -284,7 +285,7 @@ local function ConstructTextEditor(frame)
       end
       self.oldOnTextChanged(...);
     end);
-    if(data.controlledChildren) then
+    if(data.controlledChildren and not setOnParent) then
       local singleText;
       local sameTexts = true;
       local combinedText = "";
@@ -360,7 +361,7 @@ local function ConstructTextEditor(frame)
   end
 
   function group.Close(self)
-    if(self.data.controlledChildren) then
+    if(self.data.controlledChildren and not self.setOnParent) then
       local textById = editor.combinedText and extractTexts(editor:GetText(), self.data.controlledChildren);
       for index, childId in pairs(self.data.controlledChildren) do
         local text = editor.combinedText and (textById[childId] or "") or editor:GetText();
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -2341,12 +2341,13 @@ local function GetCustomCode(data, path)
   return data;
 end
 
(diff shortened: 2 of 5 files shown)


emptyrivers:
As an example, you can do something like:
image
(though why you'd want a parabola shaped dynamic group, I won't try to guess)

mrbuds:
Can i ask if you have an example of working custom code ?

InfusOnWoW:
Basically this:
https://github.com/WeakAuras/WeakAuras2/blob/b47c0de0a842c6eae721e2172003b0d670620645/WeakAuras/RegionTypes/DynamicGroup.lua#L219

That is the inner function, not the outer!

So you get: 
function(newPositions, activeRegions)

with newPositions a array of coordinates, which your function should fill
with activeRegions being a array of "regionData", which contain pointers to the actual data/region etc.

emptyrivers:
The exact details on the interface are being worked out, yes. But that’s the basic idea.

This iteration doesn’t give custom code the regionData objects directly, mostly out of fear of what people could do to bust it up. The object you get is a clone of the regionData, but with the region/data removed, and the region’s state added as an attribute.

On the sort function, you get 2 of these objects a and b, and you are expected to return true if a must be sorted to an earlier spot than b (in that respect it is equivalent to how table.sort works).

The layout/grow function is given two tables as infus describes. The coordinates are expected to be in the form of: {x, y, show}, with x,y being the absolute offset from the anchor point of the dynamic group, and show being an optional boolean which can be used to force the child to hide (but without triggering the normal on hide stuff like actions and animations, mind).

Custom sort is pretty much set in stone atm, mostly because it’s so simple. Custom layout is not yet fully developed, so the interface may change by the time I submit this for real.

InfusOnWoW:
aura_env of the individual auras would probably be somewhat interesting, since that gives access to the config too

WeakAurasOptions/OptionsFrames/TextEditor.lua
```diff
@@ -243,11 +243,12 @@ local function ConstructTextEditor(frame)
     end
   end);
 
-  function group.Open(self, data, path, enclose, multipath, reloadOptions)
+  function group.Open(self, data, path, enclose, multipath, reloadOptions, setOnParent)
```
InfusOnWoW: setOnGroup?

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -2341,12 +2341,13 @@ local function GetCustomCode(data, path)
   return data;
 end
 
-function WeakAuras.AddCodeOption(args, data, name, prefix, order, hiddenFunc, path, encloseInFunction, multipath, extraSetFunction, extraFunctions, reloadOptions)
+-- ? Possibly we should refactor this function to need a less extreme number of args?
```
InfusOnWoW: Well yes... But all those options are somehow needed....

InfusOnWoW:
So I looked through the PR, and it appears to be fine and I actually think the interface would work.

I wanted to comment on: "This iteration doesn’t give custom code the regionData objects directly, mostly out of fear of what people could do to bust it up."

Now, that matches what I would think too, but it's worth pointing out something:
In general, in WeakAuras I try to give custom code the same power as internal code. So if the internal interface to the sort/layout function is the regionData, then the custom code could simply use the same interface.

There's obviously the danger of people abusing that, but there's also a strong argument that because the WA code has traditionally been very hackable from custom code, that has lead to lots of innovation. That is, because we tend to give people too much power, we learn what people will do with that, and then design the next feature to make whatever hacks people did, a proper feature.

emptyrivers:
The primary thing I want to limit is access to the aura data. Of course authors can access it easily enough via GetData, but I think it is not necessary to increase the opportunity for a user to do something dangerous which they have no knowledge of, and which could set them back hours of work with no recourse.

Perhaps we can instead limit our own use of the data, and provide parity that way.

emptyrivers:
and by “we” I mean “I”

WeakAurasOptions/OptionsFrames/TextEditor.lua
emptyrivers: No, setOnGroup implies to me that all of the group would be affected. Which is actually the default behavior.

PR #1252, opened 2019-03-16
change position of movers when they get out of screen

mrbuds:
# Description
preview


WeakAurasOptions/OptionsFrames/MoverSizer.lua
```diff
@@ -101,6 +101,34 @@ local function ConstructMover(frame)
   return top, bottom, left, right
 end
 
+local function PositionMover(frame)
+  if frame.moveTop:GetLeft() then
```
emptyrivers: I’m not asking you to change this, but can you clarify what moveTop and moveBottom refer to? Are those the buttons which nudge the aura up and down?

PR #1254, opened 2019-03-18
Add alignment snapping for auras and groups

mrbuds:
# Description
Add alignment snapping for auras and groups
Pressing shift ignore the snapping

https://i.imgur.com/Yl0rKDs.gif

## Type of change

# How Has This Been Tested?
- [x] snapping aura
- [x] snapping group

# Checklist:

### DIFF
--- a/WeakAurasOptions/OptionsFrames/MoverSizer.lua
+++ b/WeakAurasOptions/OptionsFrames/MoverSizer.lua
@@ -517,8 +517,10 @@ local function ConstructMoverSizer(parent)
     end
 
     mover.doneMoving = function(self, event, key)
-      if event == "MODIFIER_STATE_CHANGED" and (key == "LCTRL" or key == "RCTRL") then
-        mover.align = BuildAlignLines(mover)
+      if event == "MODIFIER_STATE_CHANGED" then
+        if key == "LCTRL" or key == "RCTRL" then
+          mover.align = BuildAlignLines(mover)
+        end
         return
       end
 
--- a/WeakAurasOptions/OptionsFrames/MoverSizer.lua
+++ b/WeakAurasOptions/OptionsFrames/MoverSizer.lua
@@ -98,7 +98,21 @@ local function ConstructMover(frame)
   right:GetPushedTexture():SetRotation(-math.pi/2)
   right:SetScript("OnClick", function() moveOnePxl("right") end)
 
-  return top, bottom, left, right
+  local linex = frame:CreateLine(nil, "OVERLAY", 7)
+  linex:SetThickness(2)
+  linex:SetColorTexture(1,1,0)
+  linex:SetStartPoint("BOTTOMLEFT", UIParent)
+  linex:SetEndPoint("BOTTOMRIGHT", UIParent)
+  linex:Hide()
+
+  local liney = frame:CreateLine(nil, "OVERLAY", 7)
+  liney:SetThickness(2)
+  liney:SetColorTexture(1,1,0)
+  liney:SetStartPoint("TOPLEFT", UIParent)
+  liney:SetEndPoint("BOTTOMLEFT", UIParent)
+  liney:Hide()
+
+  return top, bottom, left, right, linex, liney
 end
 
 local function PositionMover(frame)
@@ -345,7 +359,7 @@ local function ConstructMoverSizer(parent)
   frame.top, frame.topright, frame.right, frame.bottomright, frame.bottom, frame.bottomleft, frame.left, frame.topleft
   = ConstructSizer(frame)
 
-  frame.moveTop, frame.moveBottom, frame.moveLeft, frame.moveRight = ConstructMover(frame)
+  frame.moveTop, frame.moveBottom, frame.moveLeft, frame.moveRight, frame.linex, frame.liney = ConstructMover(frame)
   PositionMover(frame)
 
   frame.top.Clear()
@@ -398,7 +412,7 @@ local function ConstructMoverSizer(parent)
   end
 
   frame.ReAnchor = function(self)
-    if(mover.moving.region) then
+    if mover.moving.region then
       self:AnchorPoints(mover.moving.region, mover.moving.data)
     end
   end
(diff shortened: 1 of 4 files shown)


mrbuds:
Fixed for groups, and made group ignore snapping with it's children

Do you think the shift key for ignoring snapping need "something" to show that you can use it ? or is it common enough on editing software to let users find out themselves ? Or should it be the other way around, you need to press a key to activate snapping ?

emptyrivers:
Enable it by default, imo. If we had a preferences panel, then I’d be comfortable dropping a setting for that in there (we don’t need a row of checkboxes across the top of the options panel for every little one-off setting).

Stanzilla:
+1 for default enable. You can leave shit holding to disable it, probably does not need that much exposure and we can mention it in the changelog.

mrbuds:
ok that's current behavior

emptyrivers:
@Stanzilla why is this an epic?

Stanzilla:
> @Stanzilla why is this an epic?

because I did not read what the label text says and thought it's epic!

emptyrivers:
😂

WeakAurasOptions/OptionsFrames/MoverSizer.lua
```diff
@@ -450,6 +464,32 @@ local function ConstructMoverSizer(parent)
...
+        if k ~= data.id and v.view.visibility ~= 0 then
+          tinsert(x, v.view.region:GetLeft())
+          tinsert(x, v.view.region:GetRight())
+          tinsert(y, v.view.region:GetTop())
+          tinsert(y, v.view.region:GetBottom())
+        end
+      end
+      table.sort(x)
+      table.sort(y)
+      for k, v in ipairs(x) do
```
InfusOnWoW: the convention is to use k,v for "key", "value" that is "pairs".

WeakAurasOptions/OptionsFrames/MoverSizer.lua
```diff
@@ -450,6 +464,32 @@ local function ConstructMoverSizer(parent)
...
+        if v ~= x[k+1] then
```
InfusOnWoW: So here, we probably want to do something like, if abs(v - x[k+1]) > 2 then

Basically throwing away anything that is almost the same as the next coordinate.

InfusOnWoW: On thinking again, nah what you have is actually fine.

WeakAurasOptions/OptionsFrames/MoverSizer.lua
```diff
@@ -98,7 +98,21 @@ local function ConstructMover(frame)
   right:GetPushedTexture():SetRotation(-math.pi/2)
   right:SetScript("OnClick", function() moveOnePxl("right") end)
 
-  return top, bottom, left, right
+  local linex = frame:CreateLine(nil, "OVERLAY", 7)
```
Stanzilla: lineX maybe?

mrbuds:
These last minute commits adds:
- Snapping will search for matching frames in reverse order if the mouse is moved in left or bottom direction
It should make positioning feel more natural when there are 2 close matching frames
- Added horizontal and vertical lines for middle of UIParent https://i.imgur.com/K7SOQ5y.jpg
- Holding ctrl will align with middle of frame instead of sides
But there is a little trick with the Ctrl key, the list of snapping lines is build when you start moving frames (mousedown)
If you don't hold Ctrl on mousedown the snapping will center your aura with sides of visible auras https://i.imgur.com/o4BSiFa.png
If you hold Ctrl on mousedown, the snapping will center your aura with middle of visible auras https://i.imgur.com/zgtXjIQ.png

mrbuds:
Added selection of auras by clicking directly on their displayed region, support ctrl for multi selection.

TODO: (for 2.13 ?)
- make resizing use snapping lines
- handle multi-selection for moving & resizing

InfusOnWoW:
I've found two bugs so far:

1) 
* Create a new icon
* Create a second icon
* Make sure both are visible by e.g. clicking on the loaded eye
* Select them both by clicking on the actual auras
=> In the list view both are shown as selected, but changing any settings only affects one aura

2)
* Create a new icon
* Click the eye icon next to it and drag it to a new place
* Create a second new icon
* Drag it away
=> Clicking on the first created aura doesn't select it

InfusOnWoW:
Another small bug:
https://youtu.be/svnewt6nAgE

As I drop the middle icon towards the left or right line, it disappears and the other line appears. And as I drag towards that line, it disappears.

InfusOnWoW:
The OnMouseDown handled remains after closing the options.
Steps to reproduce:
* Add a "print" to the handler for OnMouseDown
* Open the options, Close the Options
* Click on a aura

InfusOnWoW:
We also break some auras if they for some reason set a script, but I'm entirely okay with that.

mrbuds:
> 
> 
> I've found two bugs so far:
> 
> 
>     * Create a new icon
> 
>     * Create a second icon
> 
>     * Make sure both are visible by e.g. clicking on the loaded eye
> 
>     * Select them both by clicking on the actual auras
>       => In the list view both are shown as selected, but changing any settings only affects one aura
> 
> 
> 
>     * Create a new icon
> 
>     * Click the eye icon next to it and drag it to a new place
> 
>     * Create a second new icon
> 
>     * Drag it away
>       => Clicking on the first created aura doesn't select it

fixed

InfusOnWoW:
Pressing Shift while dragging seems to cancel the dragging for me.

InfusOnWoW:
> if event == "MODIFIER_STATE_CHANGED" and key == "LCTRL" or key == "RCTRL" then

In lua, and binds stronger than or so this is:

if (event == "MODIFIER_STATE_CHANGED" and key == "LCTRL") or key == "RCTRL" then

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -661,21 +661,21 @@ function WeakAuras.CreateFrame()
     self:FillOptions(displayOptions[tempGroup.id]);
   end
 
-  frame.ClearPicks = function(self, except)
+  frame.ClearPicks = function(self, noHide)
```
InfusOnWoW: There's a caller that calls ClearPicks with a aura name: 
WeakAuras.ClearPicks(data.id);

mrbuds:
> 
> 
> The OnMouseDown handled remains after closing the options.
> Steps to reproduce:
> 
>     * Add a "print" to the handler for OnMouseDown
> 
>     * Open the options, Close the Options
> 
>     * Click on a aura

https://github.com/WeakAuras/WeakAuras2/pull/1254/commits/175e96fcfa8f0cded73713cb6e6841ca2a483196 fix it but it feels dirty, do you see a better way to do it?

InfusOnWoW:
I think that's okay.

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
mrbuds: "except" is not used by the function

mrbuds: or i'm blind

InfusOnWoW: That doesn't mean you should fix the call that passes a parameter!

mrbuds:
Everything fixed, i also removed the ability to use ctrl for multi-selection directly on regions until mover/sizer handle it

Stanzilla:
it's time for the obligatory "bump the minor" comment

PR #1266, opened 2019-03-25
BuffTrigger2: multiselect for buffclass

mrbuds:
# Description
Allow to choose multiple debuff class instead of just one

img

# Migration !
**It add a data migration that is backward incompatible with previous versions of weakauras**

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -1884,14 +1884,14 @@ local function createScanFunc(trigger)
       end
     ]]
   end
-
-  if use_debuffClass and trigger.debuffClass then
+  if use_debuffClass then
     local ret2 = [[
-      if matchData.debuffClass ~= %q then
+      local tDebuffClass = %s;
+      if not tDebuffClass[matchData.debuffClass] then
         return false
       end
     ]]
-    ret = ret .. ret2:format(trigger.debuffClass)
+    ret = ret .. ret2:format(trigger.debuffClass and type(trigger.debuffClass) == "table" and WeakAuras.SerializeTable(trigger.debuffClass) or "{}")
   end
 
   if trigger.ownOnly then
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -7,8 +7,11 @@ This file adds the following API to the WeakAuras object:
 DisplayToString(id, forChat)
 Converts the display id to a plain text string
 
-DisplayToTableString(id)
+DisplayAuraToString(id)
 Converts the display id to a formatted table
+
+SerializeTable(data)
+Converts the table data to a formatted table
 
 ShowDisplayTooltip(data, children, icon, icons, import, compressed, alterdesc)
 Shows a tooltip frame for an aura, which allows for importing if import is true
@@ -726,14 +729,18 @@ local function recurseStringify(data, level, lines)
   end
 end
 
-function WeakAuras.DisplayToTableString(id)
+function WeakAuras.DisplayAuraToString(id)
   local data = WeakAuras.GetData(id)
-  if(data) then
-    local lines = {"{"};
-    recurseStringify(data, 1, lines);
-    tinsert(lines, "}")
-    return table.concat(lines, "\n"):gsub("|", "||");
+  if data then
+    return WeakAuras.SerializeTable(data)
   end
+end
+
+function WeakAuras.SerializeTable(data)
+  local lines = {"{"}
+  recurseStringify(data, 1, lines)
+  tinsert(lines, "}")
+  return table.concat(lines, "\n"):gsub("|", "||")
 end
 
 function WeakAuras.RefreshTooltipButtons()
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -732,7 +732,7 @@ end
 function WeakAuras.DataToString(id)
   local data = WeakAuras.GetData(id)
   if data then
-    return WeakAuras.SerializeTable(data)
+    return WeakAuras.SerializeTable(data):gsub("|", "||")
   end
 end
 
@@ -740,7 +740,7 @@ function WeakAuras.SerializeTable(data)
   local lines = {"{"}
   recurseStringify(data, 1, lines)
   tinsert(lines, "}")
-  return table.concat(lines, "\n"):gsub("|", "||")
+  return table.concat(lines, "\n")
 end
 
 function WeakAuras.RefreshTooltipButtons()
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -7,7 +7,7 @@ This file adds the following API to the WeakAuras object:
 DisplayToString(id, forChat)
 Converts the display id to a plain text string
 
-DisplayAuraToString(id)
+DataToString(id)
 Converts the display id to a formatted table
 
 SerializeTable(data)
@@ -729,7 +729,7 @@ local function recurseStringify(data, level, lines)
   end
 end
 
-function WeakAuras.DisplayAuraToString(id)
+function WeakAuras.DataToString(id)
   local data = WeakAuras.GetData(id)
   if data then
     return WeakAuras.SerializeTable(data)
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -5571,23 +5571,3 @@ function WeakAuras.FindUnusedId(prefix)
   end
   return id
 end
-
-function WeakAuras.SerializeTable(t)
-  local s = {"{"}
-  for k, v in pairs(t) do
-    s[#s+1] = k
-    s[#s+1] = "="
-    if type(v) == "boolean" then
-      s[#s+1] = v and "true" or "false"
-    elseif type(v) == "string" then
-      s[#s+1] = ("%q"):format(v)
-    elseif type(v) == "number" then
-      s[#s+1] = v
-    elseif type(v) == "table" then
-      s[#s+1] = WeakAuras.SerializeTable(v)
-    end
-    s[#s+1] = ","
-  end
-  s[#s+1] = "}"
-  return table.concat(s)
-end
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -2860,9 +2860,8 @@ function WeakAuras.Modernize(data)
         and type(triggerData.trigger.debuffClass) == "string"
         and triggerData.trigger.debuffClass ~= ""
         then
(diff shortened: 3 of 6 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -2850,6 +2850,24 @@ function WeakAuras.Modernize(data)
     end
   end
 
+  -- Version 14 was introduced March 2019 in BFA
+  if data.internalVersion < 14 then
+    if data.triggers then
+      for triggerId, triggerData in pairs(data.triggers) do
+        if type(triggerData) == "table"
```
emptyrivers: If the triggerData isn’t a table, then we have bigger problems :)

WeakAuras/WeakAuras.lua
```diff
@@ -2850,6 +2850,24 @@ function WeakAuras.Modernize(data)
...
+        and triggerData.trigger
+        and triggerData.trigger.debuffClass
+        and type(triggerData.trigger.debuffClass) == "string"
+        and triggerData.trigger.debuffClass ~= ""
+        then
+          local idx = trigger.debuffClass..""
+          data.triggers[triggerId].trigger.debuffClass = {}
+          data.triggers[triggerId].trigger.debuffClass[idx] = true
```
emptyrivers: You should be able to simplify these two statements to `triggerData.trigger.debuffClass = { [idx] = true }`

WeakAuras/WeakAuras.lua
```diff
@@ -5554,3 +5572,23 @@ function WeakAuras.FindUnusedId(prefix)
   end
   return id
 end
+
+function WeakAuras.SerializeTable(t)
```
emptyrivers: This would be the 4th serialization of tables that we have in WeakAuras, by my count, not including AceSerialize. Do we really need another one?

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -144,7 +144,21 @@ local function GetBuffTriggerOptions(data, optionTriggerChoices)
           and not IsSingleMissing(trigger)
           and trigger.use_debuffClass)
       end,
-      values = WeakAuras.debuff_class_types
+      values = WeakAuras.debuff_class_types,
+      get = function(info, index)
```
emptyrivers: Did you test this with multi-selected auras?

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -144,7 +144,21 @@ local function GetBuffTriggerOptions(data, optionTriggerChoices)
...
+        return trigger.debuffClass and trigger.debuffClass[index] or false
+      end,
+      set = function(info, index)
```
emptyrivers: Also make sure the set works.

WeakAuras/WeakAuras.lua
```diff
@@ -5554,3 +5572,23 @@ function WeakAuras.FindUnusedId(prefix)
...
+  local s = {"{"}
+  for k, v in pairs(t) do
+    s[#s+1] = k
+    s[#s+1] = "="
+    if type(v) == "boolean" then
+      s[#s+1] = v and "true" or "false"
+    elseif type(v) == "string" then
+      s[#s+1] = ('"%s"'):format(v)
```
emptyrivers: Nope. This will create an invalid string with input that contains a “.

WeakAuras/WeakAuras.lua
emptyrivers: Also what’s with the concatenation with an empty string?

WeakAuras/WeakAuras.lua
mrbuds: it's for filtering "disjunctive" and "activeTriggerMode"

WeakAuras/WeakAuras.lua
mrbuds:
```
local Serializer = LibStub:GetLibrary("AceSerializer-3.0")
print(Serializer:Serialize({a = true, b = true}))
```
> ^1^T^Sa^B^Sb^t^^

```
print(WeakAuras.SerializeTable({a = true, b = true}))
```
> {a=true,b=true,}

Do you which function i could use for this case ?

WeakAuras/WeakAuras.lua
mrbuds: will fix

WeakAuras/WeakAuras.lua
mrbuds: will fix

WeakAurasOptions/BuffTrigger2.lua
mrbuds: i didn't, and it doesn't work indeed

WeakAuras/WeakAuras.lua
emptyrivers: Ah, fair. ipairs would work then.

WeakAuras/WeakAuras.lua
emptyrivers: Ah, I suppose AceSerialize is intended for creating strings that can be deserialized with AceSerialize.

I was actually thinking about the DisplayToTableString method in transmission specifically. It might make sense to go ahead and merge that function with this one, since they do functionally the same thing.

WeakAurasOptions/BuffTrigger2.lua
emptyrivers: I was afraid of that. I wish we didn’t have to handle multi select, it’s so annoying :(

WeakAuras/WeakAuras.lua
mrbuds: fixed

WeakAurasOptions/BuffTrigger2.lua
mrbuds: correction, it was working, i miss tested because it's very rare i do multi-selection changes with different state on each aura and didn't know how it was showing
https://i.imgur.com/uOT6Cg4.gif

emptyrivers: I still don’t really like that. Every other control in WeakAuras (aside from custom code) obeys the requirement that after one user interaction, any discrepancy is eliminated. Also, the name is always blue if there is a discrepancy. The behavior of this control does not conform to this requirement, which makes it unique amongst non-code controls.

Think about what the user would assume if they saw that. They’d assume the same thing you did, and ask why this doesn’t work.

mrbuds:
The set & get functions should be moved somewhere in WeakAurasOptions.lua to be usable by any trigger, with fixes for showing correctly in blue each checkbox with discrepancy on multi-selection.
Tbh i'm still a bit lost with the complexity of this part of the code.
I'd like ideally to also have the possibility to make each sub-checkbox a tri-state, both for triggers and load conditions.

WeakAurasOptions/BuffTrigger2.lua
mrbuds: fixed

WeakAuras/BuffTrigger2.lua
```diff
@@ -1884,14 +1884,14 @@ local function createScanFunc(trigger)
...
+  if use_debuffClass then
     local ret2 = [[
-      if matchData.debuffClass ~= %q then
+      local tDebuffClass = %s;
+      if not tDebuffClass[matchData.debuffClass] then
         return false
       end
     ]]
-    ret = ret .. ret2:format(trigger.debuffClass)
+    ret = ret .. ret2:format(trigger.debuffClass and type(trigger.debuffClass) == "table" and WeakAuras.SerializeTable(trigger.debuffClass) or "{}")
```
emptyrivers: One small note: this will cause a small memory leak, which someone is bound to complain about. I don't personally care that much unless it causes a serious issue for somebody, but somebody else might care.

WeakAuras/Transmission.lua
```diff
@@ -726,16 +729,20 @@ local function recurseStringify(data, level, lines)
   end
 end
 
-function WeakAuras.DisplayToTableString(id)
+function WeakAuras.DisplayAuraToString(id)
```
emptyrivers:
```suggestion
function WeakAuras.DataToString(id)
```
either that, or aura data, imo. If you don't care that much you can skip this comment.

WeakAuras/Transmission.lua
```diff
@@ -726,16 +729,20 @@ local function recurseStringify(data, level, lines)
...
+  if data then
+    return WeakAuras.SerializeTable(data)
   end
 end
 
+function WeakAuras.SerializeTable(data)
+  local lines = {"{"}
+  recurseStringify(data, 1, lines)
+  tinsert(lines, "}")
+  return table.concat(lines, "\n"):gsub("|", "||")
```
emptyrivers: the gsub here is intended specifically so that pipes don't get treated as magic characters in a blizzard fontstring. It's probably best to move this to DisplayAuraToString.

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -135,7 +135,7 @@ local function GetBuffTriggerOptions(data, optionTriggerChoices)
       hidden = function() return not (trigger.type == "aura2" and trigger.unit ~= "multi" and not IsSingleMissing(trigger)) end
     },
     debuffClass = {
-      type = "select",
+      type = "multiselect",
       width = WeakAuras.normalWidth,
       name = L["Debuff Type"],
```
emptyrivers: Can we go ahead and add a new feature indicator?

mrbuds: on first or second title ?
img

emptyrivers: Ah, the first one.

WeakAuras/BuffTrigger2.lua
mrbuds: I'm not good for spotting this, do you have suggestions ?

Maybe instead of using WeakAuras.SerializeTable(trigger.debuffClass) to make a new table it could use string ? That doesn't sound very efficient to me but maybe it is better to handle a string like with the following example  ?

```Lua
local t = {
  truc = true,
  chose = true,
  bidule = true
}
local matchData_debuffClass = "bidule"

function table2str(t)
  local keyset={}
  local n=0
  for k,v in pairs(t) do
    n=n+1
    keyset[n]=k
  end
  return table.concat(keyset,"#")
end
local tDebuffClass = table2str(t)
print(string.match(tDebuffClass , matchData_debuffClass))
```

InfusOnWoW: That doesn't matter imho, that code only runs once per trigger construction. So typically once.

Stanzilla:
This looks sane to me but since it's @InfusOnWoW baby i'd like him to sign off on it

mrbuds:
I checked code and tested, looks good to me

PR #1286, opened 2019-04-14
collect errors from login thread

emptyrivers:
# Description

This way, people can see errors which appear during the login process and report them to us.

## Type of change

Enhancement, i guess

# How Has This Been Tested?

I checked to see that it works

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -1414,13 +1414,17 @@ loadedFrame:SetScript("OnEvent", function(self, event, addon)
   elseif(event == "PLAYER_LOGIN") then
     local startTime = debugprofilestop()
     local finishTime = debugprofilestop()
+    local ok, msg
     -- hard limit seems to be 19 seconds. We'll do 15 for now.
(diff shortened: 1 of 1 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -1414,14 +1414,18 @@ loadedFrame:SetScript("OnEvent", function(self, event, addon)
...
     -- hard limit seems to be 19 seconds. We'll do 15 for now.
     while coroutine.status(loginThread) ~= 'dead' and finishTime - startTime < 15000 do
-      coroutine.resume(loginThread)
+      ok, msg = coroutine.resume(loginThread)
       finishTime = debugprofilestop()
     end
     if coroutine.status(loginThread) ~= 'dead' then
       WeakAuras.dynFrame:AddAction('login', loginThread)
     end
+    if not ok then
```
mrbuds: Is it not preferable to move this in the previous while loop ?

emptyrivers: A coroutine which encounters an uncaught error will terminate and have a status of dead.

So, maybe? I don't think it makes any difference either way.

PR #1295, opened 2019-04-22
Option groups

emptyrivers:
See documentation for this feature here: https://github.com/WeakAuras/WeakAuras2/wiki/Custom-Options#option-groups.

This preview introduces a new custom option type, option groups. Option groups will create a table in data.config (at the specified key), inside which the values of the sub options for that group can be found.

3 option group types are planned, though currently only one group type is implemented in this preview:

- [x] Simple
	- This is essentially just a collapsible header:
image
image

- [ ] ~~Branch~~ **This type will not be implemented, see the comments below**
	- ~~This group type will include a dropdown which the author populates, and the user is expected to select. Which options are then rendered after that depend on the choice in the dropdown. This will allow simplistic branching of options, for a more reactive interface.~~
- [x] Entry (does anybody have a better name?)
	- This group type will allow users to enter tabular data. The common use case for this is, for example, add a list of spells to track, or a list of names which should be shown in order (for e.g. an interrupt rotation).

Feedback on UI design or code approach is welcome at this stage. Please note that this is in active development, so any bugs that are found are likely to be fixed by the time that you report them.


WeakAuras/WeakAuras.lua
```diff
@@ -2868,6 +2870,27 @@ function WeakAuras.Modernize(data)
     end
   end
 
+  -- Version 15 was introduced in April 2019
+  -- This adds a path to everything in authorOptions
+  -- so that we can locate it.
+  -- The other option is that we can build it locally each time
+  -- as transient data
+  if data.internalVersion < 15 then
```
emptyrivers: This data migration is, in principle, unnecessary. It is possible to construct transient path data from the database and use that instead, and save on the SV disk size. This would require a moderate restructuring of the codebase to handle it, so for the moment I think it is not worth it.

mrbuds:
Some suggestions for group types's name :
"Collapsible Group" instead of "Simple" ?
"Conditional Option(s)" instead of "Branch" ?
"Entry List" instead of "Entry" ?

I'm not sure i fully understand how far you want to go with Entry type.
Can you give an example of what the data save will be like ?
Something like this?
```
aura_env.config.entryList = {
  [1] = { name = "entry1name", option1 = true } ,
  [2] = { name = "entry2name", option1 = false } ,
}
```

emptyrivers:
Simple groups are now fully implemented. This includes most (all) of the framework needed for the other group types.

emptyrivers:
After some time deliberating, I have decided that I will not be supporting branch type groups. The use case that they solve would be better suited by implementing an interface for the author to add hidden/disabled methods to the options.

emptyrivers:
https://cdn.discordapp.com/attachments/273144328447197186/582836896707772428/2019-05-28_02-43-21.mp4 here is a small preview of what is currently possible

emptyrivers:
There are likely more than a few bugs, and there is certainly some very dark and evil magic that we might be able to banish in this pull request, but it is (finally) at the point where testing can begin.

WeakAuras/WeakAuras.lua
emptyrivers: Ah, this is a rebase error. I did not actually end up needing this migration.

WeakAuras/WeakAuras.lua
```diff
@@ -3049,49 +3072,74 @@ function WeakAuras.AddMany(table)
...
         end
       end
+    elseif optionClass == "group" then
+      authorOptionKeys[option.key] = "group"
+      local subOptions = option.subOptions
+      if type(config[option.key]) ~= "table" then
+        config[option.key] = {}
+      end
+      local subConfig = config[option.key]
+      if option.groupType == "array" then
```
emptyrivers: There is quite a bit of black magic going on to make this validation work. To avoid that I would need to do another fairly large refactoring of custom options, which I may yet do.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -47,98 +48,87 @@
...
-  header = "noninteractive",
-}
+local conflict = {} -- magic value
 
-local function atLeastOneSet(data, references, key)
-  for childID, optionID in pairs(references) do
-    if data[childID].authorOptions[optionID][key] ~= nil then
+local optionClasses = WeakAuras.author_option_classes
+
+local function atLeastOneSet(references, key)
```
emptyrivers: A very large percentage of the changes in this file are due to a simplification which I only thought of recently. Essentially, a single selected aura is now considered to be a group with just one member. This simplified the reasoning process greatly, and should reduce the number of stupid bugs due to neglecting to change the multi selected code path or whatever.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1636,32 +1490,239 @@ local function addUserModeOption(options, config, args, data, order, i)
...
   elseif optionClass == "noninteractive" then
     userOption = {
       type = "description",
-      order = order,
+      order = order(),
       name = "",
-      width = (option.width or 1) * WeakAuras.normalWidth,
+      width = (option.width or 1) * WeakAuras.normalWidth
     }
+  elseif optionClass == "group" then
```
emptyrivers: This block (and its corresponding user mode block) are the only real “additions” in this pull request.

Causese:
reset to default isn't greyed out and clickable while everything is at default. (nothing major)
simply create option group and exit author mode to see this behaviour.

Causese:
1. enter author mode & add option and set it to "option group"  
2. add sub option & set it to description and then to color
error:
```
3x WeakAuras\WeakAuras-@project-version@.lua:3080: table index is nil
WeakAuras\WeakAuras-@project-version@.lua:3080: in function <WeakAuras\WeakAuras.lua:3075>
WeakAuras\WeakAuras-@project-version@.lua:3110: in function <WeakAuras\WeakAuras.lua:3075>
WeakAuras\WeakAuras-@project-version@.lua:3301: in function `PreAdd'
WeakAuras\WeakAuras-@project-version@.lua:3389: in function `Add'
WeakAurasOptions\AuthorOptions.lua:1377: in function <WeakAurasOptions\AuthorOptions.lua:1329>
```

Causese:
"Start Collapsed" isn't taken into account when clicking "Reset to Defaults"

Causese:
Add option -> change to "Space"
notice how "Add Option" is not at the end of the option

Causese:
1. Add an option followed by another option (do not use duplicate to add the second option)
2. duplicate the second option
notice how it freezes your game and eventually causes this error:
```
1x [string "safecall Dispatcher[3]"]:9: script ran too long
[string "safecall Dispatcher[3]"]:9: in function <[string "safecall Dispatcher[3]"]:5>
```

Causese:
1. Add Slider/Number option type and try to change any number except the one from slider.
notice how it always resets to default values

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1096,538 +855,640 @@ local typeControlAdders = {
...
+      desc = desc(option, "useCollapse", L["If checked, then this option group can be temporarily collapsed by the user."]),
+      order = order(),
+      width = WeakAuras.normalWidth,
+      get = get(option, "useCollapse"),
+      set = set(data, option, "useCollapse"),
+    }
+    args[prefix .. i .. "collapseDefault"] = {
+      type = "toggle",
+      name = name(option, "collapse", L["Start Collapsed"]),
+      desc = desc(option, "collapse", L["If checked, then this option group will be initially rendered in a collapsed state."]),
```
InfusOnWoW: I think, you can replace, "be initially rendered in a collapsed state" with "start collapsed", which is much shorter ?

WeakAuras/WeakAuras.lua
```diff
@@ -3049,49 +3072,74 @@ function WeakAuras.AddMany(table)
...
+        for k, v in pairs(subConfig) do
```
InfusOnWoW: Wouldn't it be a good idea to validate the limit here (if it is set)?

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1636,32 +1497,238 @@ local function addUserModeOption(options, config, args, data, order, i)
...
+                local childConfigList = optionData.config[childOption.key]
+                local childData = optionData.data
+                if #childConfigList ~= 0 then
+                  tremove(childConfigList)
+                  setPage(id, optionData.path, #childConfigList)
+                  -- the author's init action need a chance to react to this change
+                  -- but the values themselves don't need to be revalidated
+                  -- So we can skip the full Add, but do reset aura_env
+                  -- !!! this will need some testing to ensure that there aren't any surprises!
+                  WeakAuras.ClearAuraEnvironment(childData)
```
InfusOnWoW: I would always do a full Add even if technically it's not needed.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1636,32 +1497,238 @@ local function addUserModeOption(options, config, args, data, order, i)
...
+            name = L["Move Entry Up"],
+            order = order(),
+            func = function()
+              for id, optionData in pairs(option.references) do
+                local childOption = optionData.options[optionData.index]
+                local childConfigList = optionData.config[childOption.key]
+                local childData = optionData.data
+                if childConfigList[page] then
+                  childConfigList[page], childConfigList[page - 1] = childConfigList[page - 1], childConfigList[page]
+                  WeakAuras.ClearAuraEnvironment(childData.id)
```
InfusOnWoW: Same e.g. here.

WeakAuras/WeakAuras.lua
InfusOnWoW: So the code can be removed?

InfusOnWoW:
I didn't actually test it yet, and I can't say tha I looked at all the Option code in detail. I'll give it a try tomorrow.

WeakAuras/WeakAuras.lua
emptyrivers: it would, if i had gotten around to implementing the limit on arrays ;)

InfusOnWoW:
This aura is set to a fixed array of 5 entries, but in user mode no entries show up nor is there any way to add any entries:

!nw1xZTTnm8pn7XTBwlnxFn2wU135kNzQ0R9fMqjbjZR0KAuuoX5H8zFaGsokEDB5oFsYG4V)WpakNjZLIdsrf9d)SX4Euwifb4PWSvoBqOFgKzZsKcvFyVZVTnOD2ou3m22U(Ilffo1c4txtJbF33blHUsYP)aonXGzyiHAvVjqNzvhqLf9fV8q84xEav4rDvyp52GmlH89zbdrrBB7dx65Kx9S4njGbSnS5)(pnGjSYBIkvGXmmgMgVRV9nU61ebfUWzmQ2oqoxkkh)ESGhS1OpOd5S3Q1pbvdheLO8E1PXC6trd6yK)ddPb1psTQcdA584)tU1PT4zlsZYt3jfN2wx3bisHvNYwInRluatnNp2MOFj8Z)GFEveKrvCMk3JwS(A5YP2J5Ki4vbfxTL9Db3HCm831wPcudE49OPPvnCXRlDwov96MgWpHFefqFwbf911rm4ZPBUD1DBgHCvVxHnKUwWywxrwh4wmCeqwjAL(jP4(f3iYVxKFZo0z9wnQsRrDcDoJLdwXMGbaugSND2j36b2jIBt3SH0R3(AQXOUQmOpc5rHFXvH9JFf5o4jp7ChyGMBe0GIuSYRF(Lh(ZEvfHwV8q(W803EtFrFqrSnE2jO8bggWALeEFS07ae6WsEajSolqX8GsBF)AxRT6U9VB9dmrMs3fmlPiw9X56Zz6u)gOPpDyeRgB)e1yPUlYulOwCNkqzatkEna)N0qMUpWENVnpF7x2T(tFoFcCVYOAWAy7D5BwNLkf7bDZEeJV(kASdAVbjoLHDuHZt(ofVFJKYR8WLd6ZPUr119pKYZFxiKedM6lMRIBgz(X8yjUzYIgoLxGPm2(oWeX1zI1ltJAg5(f9HGZU9i4n0Ea(G96YFybmXqur0RXfgZCZk(4IV2EZ3eNxar17qe(Frvp0Gn0ye5PtSVIb50YbcYyd67tiSNb84fbjy4uM29XLbCPQXKuM9XHMMu8lDtU(4FFQyLh(REWwEcdZVL8HHiHM3YaQBekfLkdWrJGGm4XbV)ozrjJlnlgxkUI2OfbbXIDPPzts3lzviivRB4lg5LYdudTf7uXv8ira8wL5ROaccZMD95e8N0YPrp94yfk)PjqnvTZD(kWZx(iWezjCukoIo9QeQIK)9d

InfusOnWoW:
Same for being limited to 5 entires, there's no way to add any entries:
!nw12YTnom0VM9XDNkVjD6Rj2YTEgx5SMkD6(ctOKGK5uksvkkN48q(2xaqjxvV3YynY0G46bhaltK5srRuurp4XgJ7jzHueGNdjRD2Gq)ciZswifQHWbNFxxq7S9OUzST9dfxkkCQdW3UMgd(9qpSc6ljN(n40mdsWqc1QbtGUZQArLfdfV(y86xFev4jDv4a52a(z0XnE3q3p53ZArcx6mgvxpiVvkkNopf8rBn6wDiN9wR65rXXFR8E1PP05Jr17zm46XKGqMuRQWavumOFV4oN2I3TmnlpDVuCAxDDpeKzVdDOTeHTluatmNpcy0Zc(9VZVVkwUOkotL7jlwDDCXu7XCse8QGIR1YH(GRnhd)9DvQab1JFpzAAvdx66sNLtvVUPb8Z6urb0XkOyOUoIbFkD7DRVF7eGRg8kS)33bgZMkY6a3gHJaYpqR0ie(WYBe5piYVzp6SbRgvPZOoHoNXYrRytWaakd2Xo7K78a7eXDPB3s6ny)rQXOUQmOpc5rHF2vH9JFn5D0nV4CTmqZncIYkfR96xE9X)yqvrO1RpMpYS)6p1x0TkIOXS4GYhyyaRvs4dXsVhqOdl5rKW6SafZwL2(21UwB19hEZ6hyAmLUlzwsrS6JtyNZ05(fpRjeFeRMA)e1yLUpYulOwCVkqzatk(ra(pPHmDFK9E7U88DFE)Mp(P8zW9AJQbRHD3NVDtwQuCa0nhqm(9xrdDq3niXPmSNkCkUgNI30qs5Lp4ca95u3O67)Bs55VlesIbt9fZvXDum)42yjUfSn8YHjgYsmLX2xlte3Kj2SknQzK7xmeco7UJG3q7b4loOl)MfWedrfXGg3vM4sk(WYV0DZxfNx)q17ye(Frvp0Gn0ye5PtSVIb50QrcYud6pNrypd4XvYlWWPmDhIld4svJjPm7dJnnP4x6NTi)FFQyTh((aylpHH53wC9yKqZ7ya1nbLIsLb4Orqqg80O3FJSOftlnlMwkUM2OfbbXY9PPzZs3lzviivRB4)IIxkpsn0wStfxWJebWBvMVGciiml59NtW)Hwon6PNgRq5ppdQPQ9wNVcOPpIPjfRGJsXr0PxTGQi5F9

InfusOnWoW:
In User Mode moving a entry does not change which entry is shown. That is, I expect on moving entry 2 up to entry 1, to be shown the "same" entry so the first one.

InfusOnWoW:
Changing a aura to limited/fixed size does not remove excessive entries.

InfusOnWoW:
There's no way to setup defaults. I think that's not needed for the first version, and it's not clear to me how to even do that in the ui.

InfusOnWoW:
In this group, the two auras have a different amount of entries. Depending in which order they are selected either 1 or 3 entries are show in User Mode

!vJ12YTTnq0VMmtFPDcLTt9RwYujAgfkxd6u3(aTHiHOqnfadjOKLEWF7DxaEbIImMj9w6mEKPwSayVC29Subob(bK8aY1STbKnbKi8pq0sA4trzY0jYezgiWZr)3i9NNP)88aV3(txeOaLLzrSSXLBjGmoHF4anl6Lh9LYefhe98IvRYzkylvQhSmGS3sAglMlf(7tzbK4mzrALIe(bgCHVlGqfHRLz3i5c4sN4657EBLsdYoPHk4kY1USIMPGhaPR4cE(AZZWJgPWZQmECmll3CQ(1cWhJyllwTYySFWD(ntVBoSU(R0Im6i4cszjjZIYnNBEXs2wMqrGDXFoG8WKRi(pq8V6w4WkWlLKMq3dhorq3Wk3LElWfWOjQ1nhYnzm9HqUXD(CuVcrJPPmgFbhYKBh9R)Wvp)5Kr)HeDnfltqt(e4squOmIwOGi6I0Q4cgKe8nu872Hjse4vOWhmozolukaNR0Nfsbd37gkxmCTRd8dtFm9aoLhB3lpEgQuYQsOW4f((l(4C3PWrTkdIFqQIQOyAReZmfLAsxKj3666b7pKMW0QyqqZeAKOdiaU9v8yt4WSyfo9c9IrCR4vIKQRykYzpeMqZZrGnK9drHBka4FfGYSABP5i8ULW671nkgmzVYawynuSxVsXEwnQvfckZzQeqFMkjaDgNi3nNjIbyLNZBpUqSZQm9z8LlYC0MiuWhj3ji74GzngRuYPkm9YW4syrUsUXhoS7sJqzKIY)xTrThVSVA166t7Q2)xxRQD0TmFJWpkJGm0pcPeCLK01gqSoRIzWaY0m(HxE8xkOrieh6Wcg1tmw6viKtDlwgPbGnvhvPq7wW1i8YwTvHnWSnndGOI13CGJuHj8xvLZARcLEIsA3EhpcbFV7Cm)YdFA)1L9aqRhHNdaSPbLh1d42zV)d(wXRPj0yOpYI78NpZZnGSMXJxRmxB7ofvfY)1RD12LRGUmHfH1aO7mnJ95cMiCpYgn6cJW7TsjABEcyZqp0nAS0mpYSRDnAAGVllukPyXwwgaaTkS7gwO33Ai2kyGTdrTE6TB1LTVoXw92hM2n92hK(QA))vt62ZjWH9306aBRCnp3e2bu0bPCttOTPb4OZTcDTriLn66inOxjG8M8E4sdhgGfQjEtQM5vwHni1urLeCgXOpO3rfuQSY43Sqn6g5Cibh4DzjRVJ0z5Lt(u6v3tqlfal56oQ6G1ynrvt91Ytj0AptqRodw9y15aLmooHPp2Rz5H4j(eB)rf96(KqpWILV8Or2lp6GnMxrHchChLDdCQ7YaNwj)0YQlIlsluTp8rDE4JoYCsSi6QUtYr3PQ6okh80AZ1AHcHywcnnxtSfA9S2Mk3BcFdx5Bb61YncOzzyDRXIFVrFDpeVlmwbMygxnECvSORCEf2OfVE)O3xLHV9ibNWK)kKqdNAVFGy31XDwTnmoV)JMlOB2)2Zi0m9Wx3Cb6Q8PxMgU4ap)N)97T60D80w)RqJ2mLYFVeOdHl4FuI1Hg56VK475xLQxcWt5xpDoI2SsDqW2)7DDY0EweRFJCpNq4y1a3ILPBAOt4EAA5)LjES4s6HTPxQPAUTVZ4EAgLyOJQO9OUERZ2JxBnWF75J)MEBeZpiqZ4n2VEZx3p0t7bGchwJhlEAqYwyLZhHlg8N)

emptyrivers:
Fixed all the bugs I could find. Please let me know if you find anymore.

emptyrivers:
@InfusOnWoW it is intended that users cannot create or destroy entries, when the array is a fixed size. However, users do have the ability to reorder the entries as they wish.

emptyrivers:
Documentation for options groups may be found [here](https://github.com/WeakAuras/WeakAuras2/wiki/Custom-Options#option-groups).

Causese:
if you add an option group and and another option group as sub option, it causes this error:
`Message: Interface\AddOns\WeakAurasOptions\AuthorOptions.lua:1981: attempt to get length of local 'path' (a nil value)`

Causese:
1. add option, change it to seperator
2. add another option and duplicate it
`Message: Interface\AddOns\WeakAurasOptions\AuthorOptions.lua:1214: table index is nil`

Causese:
this one really doesn't want to go away 😄
upon entering a number in number option's "max" box
`Message: Interface\AddOns\WeakAuras\WeakAuras.lua:3121: attempt to compare string with number`

Causese:
one of the recent comits caused this I think:
import: https://wago.io/cotank4
what it looks like: https://i.gyazo.com/5eff3dbe14f24d55fc7ac6c2f00d4dc6.gif
what it should look like: https://i.gyazo.com/1c1021d6a6adb9f7714da97e79c63ed0.png

mrbuds:
- Import <https://wago.io/BFADungeonTargetedSpells>
- Notice "particules = 14", "Frquency = 0.3", "AutoCast's Scale = 1"
- Notice Reset to Default is grey
- Switch to Author mode
- Default values for these 3 fields is 0
- Export to Lua to check default values in data : they are at 14 / 0.3 / 1

emptyrivers:
rebased on top of master.

emptyrivers:
I had a bit of extra time, so i went ahead and implemented move in/out of group controls. I didn't spend much time on the icons, so if somebody has any better designs that they can pull out of the hat, please feel free to submit them.
in/out of group preview

InfusOnWoW:
You forgot to add the new icons for move out of group/insert into group

emptyrivers:
@Stanzilla me neither, but it was the best i could come up with that wasn't fragile as hell. I figured that somebody would suggest some magic solution which makes a lot of sense, once option groups are released.

PR #1301, opened 2019-04-25
collapsible headers for conditions

emptyrivers:
# Description

This solves https://github.com/WeakAuras/WeakAuras2/projects/8#card-18039065.

## Type of change

# How Has This Been Tested?

I checked to see if it worked and it did.


Stanzilla:
Small afterthought: As long as we don't have automated or custom names for the headers, I think we should auto expand them here. @emptyrivers

PR #1305, opened 2019-04-25
remove stale feature indicators

emptyrivers:
I think in 2.13, these features will have aged enough to not count as "new" anymore.

Templates from 2.7.3
Custom Options, Glow, Cooldown Settings from 2.10.0


InfusOnWoW:
I don't have any opinion on that, and I guess stan will have one.

PR #1323, opened 2019-05-04
reworked DBM and BigWigs triggers

### LINKED ISSUE #1322: DBM_TimerUpdate
Support the `DBM_TimerUpdate` event which will pass 
`fireEvent("DBM_TimerUpdate ", id, elapsed, totalTime)` 

This would find existing bar by ID, then overwrite elapsed and total with new values

mrbuds:
DBM &  BigWigs:
- made both more similar and readable
- renamed "Adjust Timer" to "Offset Timer"
- added a "count" filter & condition, working only if bossmod show a (%d+) in their bar

DBM:
- fix handling of DBM_TimerUpdate event
- new filter & condition for bar type
- removed the "ID" field from trigger
- renamed Spell Id field
- renamed colorId state property to dbmType
- new property added to state available for custom code
  dbmColor

BigWigs:
- new "Emphasized" tristate checkbox & condition
- new "Is Cast" tristate checkbox & condition
- removed "addon" field from trigger
- new properties added to state available for custom code
  bwBarColor, bwTextColor, bwBackgroundColor

Fixes #1322 

## Type of change

=> removed "addon" field for BW trigger, and "ID" for DBM trigger, as i think both are not used and confuse users a lot

# How Has This Been Tested?
- [x] Many tests with bw & dbm test bars
- [x] Many tests with bw & dbm on eye of azshara 1st boss
- [x] `DBM:GetModByName("2004"):TestFunction(20)` for dbm bugfix


mrbuds:
description updated with last changes

MysticalOS:
Just want to add a note that

https://github.com/WeakAuras/WeakAuras2/pull/1323/commits/500f800db91f145a76d110d6efc34b57847646e0

You have 3 as "debuff" i actually more broadly use it as "target" which doesn't just apply to debuffs but also area targeted spells (think shadow crash for example, it's not a debuff, but it's targeted). so I grouped them under same "targeted" category

Just adding that clarification so you don't call it "debuff"

mrbuds:
Alright thanks for your input!
I have renamed "Debuff" and kept "Important"

This how looks updated trigger:
- DBM
img
- BigWigs
img

SpatenLa:
Is it possible to keep the order the same for both? mainly about the Spell Id field, the "Type" field not so much as the type are checkboxes in BigWigs
grafik

MysticalOS:
It looks like you removed timerId field. that'll break support for timers that have no spell ID (pull timers, break timers, custom localized timers within mods that may not have a spell Id such as "next phase" or something).

While I realize it was a confusing thing, and certainly shouldn't have been at top where it was, it's definitely needed in the event a spellID or dungeon journal ID doesn't exist. in other words, the spellId arg of events does have a chance to be nil. Hmm actually i suppose in that regard you can go "if not spellId, spellId = timerid" and then still leave one box in weak auras. but then that box would have to except both string and number

mrbuds:
timerId is still used internally as index, just the same as before, i just removed the ability to filter timers with it because i don't think it's used but i can be wrong about it

MysticalOS:
I have seen users come into dbm discord and say they use it, but in most cases they should have been using spellId instead, so I definitely see how it had more mis use then proper use. so yeah you're probably right it should be gone, I forgot they can just parse the string timers that have no spellID by message still

WeakAuras/GenericTrigger.lua
```diff
@@ -2196,344 +2196,407 @@ end
...
-      bars = {};
-      WeakAuras.ScanEvents("DBM_TimerStopAll", id);
+      local bar = bars[id]
+      bar.message = msg
+      bar.expirationTime = expirationTime
+      bar.duration = duration
+      bar.icon = icon
+      bar.timerType = timerType
+      bar.spellId = tostring(spellId)
+      bar.count = tostring(msg:match("(%d+)") or 0)
```
InfusOnWoW: msg.match already returns a string, so you do something like msg:match("(%d+)") or "0"

WeakAuras/Prototypes.lua
```diff
@@ -2998,159 +2998,129 @@ WeakAuras.event_prototypes = {
...
+        trigger.remaining_operator or "<"
+      )
     end,
     statesParameter = "full",
     args = {
-      {
-        name = "id",
-        display = L["Id"],
-        type = "string"
-      },
```
InfusOnWoW: I don't think we can remove the id, we even have a explanation on how the id works in #pins. 

We can probably remove it for 9.0, but not in the middle of a expansion set.

mrbuds:
I have re-added the Id for DBM but renamed it as "Timer Id" (GUI only) so there is less confusion with spell or encounter id.

`ViragDevTool_AddData(aura_env.state.addon, "addon")` in action > on show 
img

WeakAuras/GenericTrigger.lua
```diff
@@ -2334,10 +2337,10 @@ do
     return bars
   end
 
-  function WeakAuras.GetDBMTimer(message, operator, spellId, extendTimer, dbmType, count)
+  function WeakAuras.GetDBMTimer(id, message, operator, spellId, extendTimer, dbmType, count)
     local bestMatch
-    for id, bar in pairs(bars) do
-      if WeakAuras.DBMTimerMatches(id, message, operator, spellId, dbmType, count)
+    for timerId, bar in pairs(bars) do
+      if WeakAuras.DBMTimerMatches(timerId, id, message, operator, spellId, dbmType, count)
```
InfusOnWoW: id here.

InfusOnWoW:
You did test that, right? It looks good to me now.

mrbuds:
I'd like to add some tooltip, but checkbox & tristate doesn't seem to handle the "desc" field

InfusOnWoW:
Try adding a desc = arg.desc, in ConstructOptions at the "right" place.

PR #1333, opened 2019-05-10
Fix Custom Sort For Dynamic Groups

Paojy:
custom sort bug fix


emptyrivers:
This shows how much people actually test my submissions >.<

PR #1340, opened 2019-05-13
Smallerpatches

InfusOnWoW:


mrbuds:
- New Event>Chat Message
- Check "Source Name", notice "Source Name" turn from grey to yellow
- Uncheck it, color doesn't change back to grey
https://i.imgur.com/72Ovsos.gif

Same behavior with any trigger with "string" options
On master branch it's working

InfusOnWoW:
Fixed

PR #1345, opened 2019-05-14
Fix cast trigger event + force enable WeakAurasCompanion addon

### LINKED ISSUE #1343: Check if Companion is installed but disabled
We do get a few but regular reports of users where they don't see pending updates in game, because a "helpful" addon manager disabled the new addon.

We could detect that the WeakAuras companion addon is installed, but disabled and show a warning in the WeakAuras window.

** Other solutions
Document this better (?)

mrbuds:
# Description
Fixes #1343

## Type of change

# How Has This Been Tested?
Disable companion -> reload -> icons show in waoptions
New icon -> cast trigger -> unit="player", caster's target="player" -> select pnj, cast HS, select self (not working on master)

### DIFF
--- a/WeakAuras/Init.lua
+++ b/WeakAuras/Init.lua
@@ -36,6 +36,9 @@ end
 WeakAuras.PowerAurasPath = "Interface\\Addons\\WeakAuras\\PowerAurasMedia\\Auras\\"
 WeakAuras.PowerAurasSoundPath = "Interface\\Addons\\WeakAuras\\PowerAurasMedia\\Sounds\\"
 
+-- force enable WeakAurasCompanion because some addon manager interfere with it
+EnableAddOn("WeakAurasCompanion")
+
 --These function stubs are defined here to reduce the number of errors that occur if WeakAuras.lua fails to compile
 function WeakAuras.RegisterRegionType()
 end
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -4715,7 +4715,7 @@ WeakAuras.event_prototypes = {
         "UNIT_SPELLCAST_INTERRUPTED"
       };
       AddUnitChangeEvents(trigger.unit, result)
-      if trigger.target and trigger.target ~= "" then
+      if trigger.use_destUnit and trigger.destUnit and trigger.destUnit ~= "" then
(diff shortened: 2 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -4715,8 +4715,9 @@ WeakAuras.event_prototypes = {
         "UNIT_SPELLCAST_INTERRUPTED"
       };
       AddUnitChangeEvents(trigger.unit, result)
-      if trigger.target and trigger.target ~= "" then
+      if trigger.use_destUnit and trigger.destUnit and trigger.destUnit ~= "" then
```
emptyrivers: What does this fix, exactly? I’m not familiar with this but of code.

WeakAuras/Init.lua
```diff
@@ -36,6 +36,9 @@ end
 WeakAuras.PowerAurasPath = "Interface\\Addons\\WeakAuras\\PowerAurasMedia\\Auras\\"
 WeakAuras.PowerAurasSoundPath = "Interface\\Addons\\WeakAuras\\PowerAurasMedia\\Sounds\\"
 
+-- force enable WeakAurasCompanion because some addon manager interfer with it
+EnableAddOn("WeakAurasCompanion")
```
emptyrivers: Did you test this without the companion addon installed? I want to be sure that we don’t error out during init for 80%+ of our users :D

emptyrivers: Also, “interfere”

Stanzilla:
In theory fine with force enabling, don't think many users would complain because they intentionally turned it off?

WeakAuras/Prototypes.lua
mrbuds: trigger.target doesn't exists, "Caster's Target" option is in trigger.destUnit
I guess at one point i renamed it and forget to edit it here

If trigger needs to know who is the target of trigger.unit then 
  register event "UNIT_TARGET" to know when target/focus/boss1/... change target

ps: i forced-push removal of AddUnitChangeEvents on trigger.destUnit

WeakAuras/Init.lua
mrbuds: yes `/run EnableAddOn("hellorivers")` doesn't raise any error

PR #1353, opened 2019-05-20
Smallerpatches

InfusOnWoW:


Stanzilla:
Group and display tab headers need the collapsable treatment as well, not quite happy with "expand" for the pencil button, any other ideas? something like "show more options" or "show advanced options"?

PR #1388, opened 2019-06-08
Smallerpatches

InfusOnWoW:


Stanzilla:
I feel like `Player Effective Level` is not very clear to users

InfusOnWoW:
You or Ironii need to suggest something better than...

emptyrivers:
Just "Effective Level" with maybe a tooltip should do the job, I think.

InfusOnWoW:
I've added a tooltip. I did keep the "Player" prefix, because most of the preceding options have it, and removing it from one probably is a bit strange.

PR #1404, opened 2019-06-16
AuthorOptions Group: use description or spellid for entry name

mrbuds:
This will look for a key named desc, description, spell or spellid to use for entry name instead of "Entry %i"
If the value is a number it will try to use GetSpellInfo on it


emptyrivers:
What is this supposed to achieve?

Stanzilla:
better names by default, I think

emptyrivers:
I see.

PR #1407, opened 2019-06-18
expose dynamic group helper functions

emptyrivers:
This way, users who want to write custom growth/sort functions can use them as building blocks


emptyrivers:
I wrote this without too much thought, and on further consideration, the grow/sort functions aren't immediately useful as they are, since they generate functions using the aura data. But compareExpirationTimes is a handy shortcut, certainly.

InfusOnWoW:
I don't have much a opinion, though the sort and grow tables are odd to use, though that doesn't bother me.

PR #1417, opened 2019-06-23
Templates update

mrbuds:
# Description
- When creating multiple auras at once, pre-select the last template option for each aura instead of the first : https://i.imgur.com/eue62Gn.jpg

- Added a few spells for all specs of warlock and rogue

Fixes https://www.wowace.com/projects/weakauras-2/issues/1348

PS: maybe don't merge it yet, Infus expressed concerned about adding resource builder/spender spell without cooldown in templates, also this doesn't add these spells for all classes, so it can be considered has a work in progress for the new spells part of this PR


mrbuds:
To backup my wish to add builders & spenders without cooldown to templates, i have an idea of a new feature that is interstitial with this and the default option feature. 
I'm thinking about adding the ability to drag & drop spells directly from spell book & talents UI to create new auras.

Issue #1139, opened 2019-01-30
Show that an aura (group) has been changed from imported version


Ghoschdi:
**Is your feature request related to a problem? Please describe.**
It is sometimes really problematic to identify (groups of) Auras that have been modified after the latest version has been imported from Wago or specific options have not been ticked during the update (e.g. update position is deselected per default). So it is hard to figure out if you really have the exact same setting as the version from Wago, so each group has to be checked manually via (re-)importing the string from Wago.

**Describe the solution you'd like**
It would be good to have a "modified" or "changed" tag or symbol next to the version information when hovering over your groups/auras. Maybe whenever an AURA is changed manually instead of an update via import?

**Describe alternatives you've considered**
An alternative could be to have an option to reset the aura (group) back to the defaults from the import or undo any changes. But i assume that would mean we need a shadow copy of the string or a versioning system either ingame or a re-import option in the companion

**Additional context**
The re-import option has been discussed in the companion project, see https://github.com/WeakAuras/WeakAuras-Companion/issues/93

Issue #1062, opened 2018-12-28
Quirks in the update code


Stanzilla:
So in this case, the size of the bars (bars are children) is changed when the aura is upgraded even though "Display" and "Size & Position" is unchecked. @emptyrivers said that the size of children is saved in the "Group Arrangement" group. Which I think is a bit irritating. What do you guys think?

[WeakAurasGroupTest.zip](https://github.com/WeakAuras/WeakAuras2/files/2716895/WeakAurasGroupTest.zip)

Stanzilla:
Nope, can still repro https://streamable.com/berk6

emptyrivers:
This is not a bug :)

I think the current situation is correct. In part due to consistency (group size & scale are categorized together, so child size & scale should be to), and in part because it's the only place that makes sense if we want to distinguish placement from general Display Tab options (which everybody seems to agree on). If the author makes a change which involves adding a new child, then the size and positioning (in fact the layout in general, which might be a better name for the category than Group Arrangement) does not necessarily allow for the new child to be added without everything else needing adjustment. And this includes the width and height of children.

Stanzilla:
Well we certainly need to better explain it then, the size changing when I deselect "size" is not great

emptyrivers:
@InfusOnWoW @mrbuds any thoughts on this?

mrbuds:
I missed the discussion about "Group Arrangement", i didn't know this option exists and tbh I would not be against merging it with "Size & Position" (noooo don't beat me rivers)

emptyrivers:
Here is where the request was made:
https://cdn.discordapp.com/attachments/198544915536412683/531842140846686238/image1.png https://cdn.discordapp.com/attachments/198544915536412683/531842140293169172/image0.png

InfusOnWoW:
I haven't followed the discussion too closely, but when I discussed this with emptyrivers on the intial PR.

It's not easy, and certainly not something that can be done 100% correctly. The basic issue is that it isn't easy to merge user and author changes and how they interact, and that there's really not a good way to do this 100% correctly.

I thought Group Arrangement was just the order of the auras in its group. That is inside "controlledChildren". (Which matters for dynamic groups.)

Whereas Size/Positions is basically the subset of display that affects how children are layouted.

Stanzilla:
yeah that was my thought as well

InfusOnWoW:
[...]

Issue #1039, opened 2018-12-17
Migrate db.displays to index by uid


emptyrivers:
**Is your feature request related to a problem? Please describe.**

With internalVersion 10, all auras are guaranteed to have a uid. So, we could index the database via the uid, and potentially rid ourselves of huge swaths of code related to Renaming.

**Describe the solution you'd like**
On ADDON_LOADED, check if the user’s db has the `data` field. If not, then create a table and populate it with aura data, indexed by uid. Then, optionally delete the displays field, or reduce it to just an id<=>uid map or some such.

**Describe alternatives you've considered**
We could also simply do nothing.

**Additional context**
This will require all sorts of research to make work, without losing readability. It will also potentially break custom code if we are not careful, so some care in deciding exactly what to do is needed.

emptyrivers:
This could also potentially allow duplicate names. Whether or not that is a good idea is something I do not know.

emptyrivers:
WeakAuras.GetDataByUID exists now. So it's "just" a matter of using that instead of GetData in new code, and eventually refactoring existing stuff to use it.

InfusOnWoW:
I think there's little point in having a ticket open for this. As far as I'm concerned new code should generally use uids instead of ids, and when refactoring code we should strive to generally use uids instead of ids.

But that's a rather long term (aka heat death of the universe) effort.

PR #1428, opened 2019-06-28
Data History

emptyrivers:
# Description

This feature introduces a new collection in the WeakAuras database, which contains snapshots of data at the moment when that data was imported. This gives us the capability to do a lot of neat things, such as:

- [x] Determine if a user has modified an aura in significant ways since importing (see #1139)
- [x] Restore data to its original incarnation
- [ ] Determine which differences in an update are external (i.e. changes by the author), and which are internal (i.e. changes by the end user), for more intelligent preservation of user settings (see #1109, #1062)
- [ ] Undelete auras which were mistakenly deleted
- [ ] Enhance the templates feature to be able to update auras created from templates (semi)-automatically.

... and more!

Additionally, this begins the process of migrating the database to be indexed by uid, see #1039


emptyrivers:
Currently, the only visible feature is the ability to restore an aura from the history collection, which is included as a preview of what is to come.

WeakAuras/WeakAuras.lua
```diff
@@ -3322,6 +3365,17 @@ local function pAdd(data)
...
 
   db.displays[id] = data;
+
+  data.uid = data.uid or WeakAuras.GenerateUniqueID()
+  local otherID = UIDtoID[data.uid]
+  if not otherID then
+    UIDtoID[data.uid] = id
+  elseif otherID ~= id then
+    -- duplicate uid
+    data.uid = WeakAuras.GenerateUniqueID()
```
emptyrivers: This will cause a collision with approximately a 1 in 73 quintillion chance. If someone legitimately does get a collision like this, then we should award the person who makes the bug report a prize ;)

WeakAuras/WeakAuras.lua
```diff
@@ -5733,3 +5787,55 @@ function WeakAuras.SafeToNumber(input)
...
+  return uid and db.history[uid]
+end
+
+function WeakAuras.RemoveHistory(uid)
+  if uid then
+    db.history[uid] = nil
+  end
+end
+
+function WeakAuras.SkipVersion(uid, version, skip)
```
emptyrivers: This will eventually replace the current wago version skipped status.

WeakAuras/WeakAuras.lua
```diff
@@ -5733,3 +5787,55 @@ function WeakAuras.SafeToNumber(input)
...
+end
+
+function WeakAuras.IsVersionSkipped(uid, version)
+  local history = WeakAuras.GetHistory(uid)
+  if history then
+    return history.skippedVersions[version]
+  end
+end
+
+function WeakAuras.RestoreFromHistory(uid)
```
emptyrivers: This can be used to undelete, and revert changes. I believe that it will be useful to make this restoration more intelligent about parent/child relationships eventually, but I have not yet decided how that should happen.

WeakAuras/WeakAuras.lua
```diff
@@ -5733,3 +5787,55 @@ function WeakAuras.SafeToNumber(input)
...
+  local history = WeakAuras.GetHistory(uid)
+  if history and history.data then
+    WeakAuras.Add(history.data)
+  end
+end
+
+function WeakAuras.ClearOldHistory(daysBack, includeNonDeleted)
```
emptyrivers: When an aura is deleted, the history is not immediately removed, so that they may undelete if they decide that it shouldn't have been deleted.

A sensible UI to "undelete" auras will need to be developed. I imagine that it will be similar in look and feel to a UI to "import" things which external addons have registered, but which haven't yet been added to the database proper (that feature already exists but it is entirely unused, and is not particularly usable in its current state, since the UI is not very intuitive at all).

emptyrivers:
Note that this contains a "slight" change to how modernizations are done. Since we now retain two "copies" of data, it makes sense to keep both copies modernized. So, WeakAuras.Modernize now may receive a second parameter, isHistory, which is true when it is a historical bit of data being modernized. This is useful for when your modernization involves making some change to the history directly, e.g. in this pull request, some history is automatically generated in order to extract the skip/ignore update flags.

emptyrivers:
I'm closing this because I want to employ a more iterative approach, so that we can get history data in sooner. Otherwise we'll have all these nice features but no data to actually use them.

PR #1437, opened 2019-07-07
fix embeds.xml & WeakAuras.toc packaging conditions for classic

mrbuds:
Fix libclassiccast & libclassicdurations not loading
Fix .toc files trying to load Model & ModelPicker files

Following error needs to be fixed upstream (or fork?)

```
2x ...akAuras\Libs\LibRangeCheck-2.0\LibRangeCheck-2.0-100192.lua:1049: Attempt to register unknown event "PLAYER_TALENT_UPDATE"
[C]: in function `RegisterEvent'
...akAuras\Libs\LibRangeCheck-2.0\LibRangeCheck-2.0-100192.lua:1049: in function `activate'
...akAuras\Libs\LibRangeCheck-2.0\LibRangeCheck-2.0-100192.lua:1093: in main chunk

Locals:
(*temporary) = <unnamed> {
 0 = <userdata>
}
(*temporary) = "PLAYER_TALENT_UPDATE"
```

### DIFF
--- a/.pkgmeta-classic
+++ b/.pkgmeta-classic
@@ -15,7 +15,6 @@ externals:
  WeakAuras/Libs/LibDataBroker-1.1: git://github.com/tekkub/libdatabroker-1-1.git
  WeakAuras/Libs/LibCompress: https://repos.wowace.com/wow/libcompress/trunk
  WeakAuras/Libs/LibSpellRange-1.0: git://github.com/ascott18/LibSpellRange-1.0.git
- WeakAuras/Libs/LibRangeCheck-2.0: https://repos.wowace.com/wow/librangecheck-2-0/trunk/LibRangeCheck-2.0
  WeakAuras/Libs/LibCustomGlow-1.0: https://repos.curseforge.com/wow/libcustomglow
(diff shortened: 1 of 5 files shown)


emptyrivers:
Are you sure this is the right fix? That looks backwards to me.

mrbuds:
Yes, right now embeds.xml condition is commented on classic
```xml
  <!--@retail
  <Include file="Libs\LibClassicDurations\LibClassicDurations.xml"/>
  <Include file="Libs\LibClassicCast\lib.xml"/>
  @end-retail@-->
```
And WeakAuras.toc try to load Model.lua

emptyrivers:
The thing is, that I would expect the retail package to also be broken if we’ve been using that tag incorrectly. If only the classic package is broken. Then that sounds like an issue with our packaging script not interpreting the tags correctly.

InfusOnWoW:
The fix looks correct to me, I would expect
<--@retail [...] @retail --> to mean that whatever is between those tags should be included for retail.

mrbuds:
"I would expect the retail package to also be broken" > retail package try to load the 2 extra libs in embeds.xml but doesn't ship them
.toc files changes were an omission

PR #1440, opened 2019-07-07
fix LibClassicDurations embedding in .pkgmeta

mrbuds:


mrbuds:
followed recommandations from https://github.com/rgd87/LibClassicDurations/blob/master/README.md#embedding-in-pkgmeta

Stanzilla:
Urm, no? Why does he not fix his lib layout?

Stanzilla:
It's fine the way it is now, closing

PR #1442, opened 2019-07-07
Smallpatches

InfusOnWoW:


WeakAuras/WeakAuras.lua
```diff
@@ -3244,8 +3244,8 @@ function WeakAuras.Modernize(data)
     end
   end
 
-  -- Version 17 was introduced May 2019 in BFA
-  if data.internalVersion < 1 then
+  -- Version 20 was introduced May 2019 in BFA
+  if data.internalVersion < 20 then
```
emptyrivers: ha, that's an unfortunate typo.

InfusOnWoW: well, the story is: I rebased my patch and wanted to run buds conversions, but not mine. So I disabled them with that check. But, forgot to reenable them afterwards.

PR #1429, opened 2019-06-29
Custom Anchor - Dynamic Group "Group by Frame" - Border Settings - Groups thumbnails

mrbuds:
# Description
- Killed group & dynamicgroup thumbnail, use the animated icons instead, with ability to set your own
img
img
- Renamed and reorganized a few options in Group and Display tab
https://i.imgur.com/hJIWwj3.png
https://i.imgur.com/sfXrmLt.png
- Use same border settings for both groups and dynamic groups and changed default settings for something closer to pixel perfect style
- New dependency LibGetFrame, publicly accessible with WeakAuras.GetUnitFrame documentation at <https://github.com/mrbuds/LibGetFrame/blob/master/README.md>
- All regions can use a custom anchor function, custom function needs to return a frame
```Lua
function()
  return SUFUnitplayer
end
```
img
- Dynamic group's growers have a new option "Group by Frame", it can be "unit frame", "nameplates" or a custom function. Enabling this option disable border options.
img
Custom anchor per unit example :
```Lua
function(frames, activeRegions)
    for _, regionData in ipairs(activeRegions) do
        local unit = regionData.region.state and regionData.region.state.destUnit
        if unit then
            local frame = C_NamePlate.GetNamePlateForUnit(unit)
            if frame then
                frames[frame] = frames[frame] or {}
                tinsert(frames[frame], regionData)
            end
        end
    end
end
```
- Custom grower interface is incompatible with previously written custom grow, but easy to adapt
It now needs to return an array keyed by anchor frame, or "" if anchor is dynamic group region, followed by an array keyed by regionData `[anchor][regionData] = { x , y }`

Custom grower example:
```Lua
function(newPositions, activeRegions)
    local frames = {}
    for _, regionData in ipairs(activeRegions) do
        local unit = regionData.region.state and regionData.region.state.unit
        if unit then
            local frame = C_NamePlate.GetNamePlateForUnit(unit)
            if frame then
                frames[frame] = frames[frame] or {}
                tinsert(frames[frame], regionData)
            end
        end
    end
    for frame, regionsData in pairs(frames) do
        local totalWidth = #regionsData - 1
        for _, regionData in ipairs(regionsData) do
            totalWidth = totalWidth + (regionData.data.width or regionData.region.width)
        end
        local x, y = - totalWidth/2, - (#regionsData - 1)/2
        newPositions[frame] = {}
        for i, regionData in ipairs(regionsData) do
            x = x + (regionData.data.width or regionData.region.width) / 2
[...]

PR #1452, opened 2019-07-09
Use same border options for dynamic group as normal groups

mrbuds:
Use same border options for dynamic group as normal groups

Changed border positioning, anchor with most top/bottom/left/right children instead of group calculated size,
this work better when children are not anchored to the center of the dynamic group


mrbuds:
example of group (top) and dynamic group with grid layout (bottom) with changed default settings
img

ps: my icons use "Clean Icons: Thin"

mrbuds:
i use another solution to fix borders on dynamic group included in https://github.com/WeakAuras/WeakAuras2/pull/1429

PR #1490, opened 2019-07-22
resort any region which is changed

emptyrivers:
This fixes a problem where trying to use a custom sort on e.g. a buff trigger that sorts by stacks doesn't properly keep the list sorted.
Reported on discord


emptyrivers:
Damn, the vscode pull request extension doesn't give you any chance to actually draft the damn thing before submitting. This is largely in response to the fact that the current raid has a mechanic which heavily encourages using a custom sorted dynamic group on certain debuffs, but the debuff in question has no duration nor expiration time, so is not flagged to be resorted by the trigger state system. The simple fix is to just interpret any change to the state (i.e. `state.changed == true`) as needing a resort, since we have no idea how a custom sort is actually intending to sort things.

PR #1509, opened 2019-07-31
misc

emptyrivers:


WeakAuras/WeakAuras.lua
```diff
@@ -3100,6 +3101,13 @@ function WeakAuras.Modernize(data)
...
+    -- second conversion: migrate name/realm conditions to tristate
+    if data.load.use_name == false then
+      data.load.use_name = nil
+    end
+    if data.load.use_realm == false then
+      data.load.use_realm = nil
+    end
   end
 
   if data.regionType == "model" and WeakAuras.BuildInfo <= 80100 then -- prepare for migration at 8.2
```
InfusOnWoW: This can be removed too

Stanzilla: @emptyrivers wanna do this now or later?

PR #1510, opened 2019-07-31
Add history collection

emptyrivers:
# Description

History is a uid-indexed table containing historical information about displays:
  source: string indicating where this data came from. Possible values:
    "user": aura was created by the user via the WeakAuras interface
    "import": aura was imported via WeakAuras.Import (i.e. pasted string or addon channel transmission)
    "addon": aura was imported via WeakAuras.RegisterDisplay (i.e. an addon requested the addition)
  data: snapshot of data as it was at last import, or nil if source == "user", and the user has not chosen to save their changes to history
  addon: if source == "addon", then this string indicates which addon supplied the last import
  skippedVersions: record of versions that the user has indicated they don't wish to see update notifications for
  allowUpdates: if false, then WeakAuras will ignore any addon-sourced update requests
  lastUpdate: UNIX timestamp of the last time the user completed an import for this display

Additionally, enforce UID uniqueness for all auras in display. In the future this will be naturally enforced, once all systems have been migrated.
But for now, check on every addon load to ensure that the database is sane

## Type of change

### DIFF
--- a/WeakAuras/History.lua
+++ b/WeakAuras/History.lua
@@ -1,0 +1,93 @@
+--[[
+This file contains all of the interfaces to the history collection. Indexed by uid of the active display.
+
+History:
+  Table containing historical information about display:
+  source: string indicating where this data came from. Possible values:
+    "user": aura was created by the user via the WeakAuras interface
+    "import": aura was imported via WeakAuras.Import (i.e. pasted string or addon channel transmission)
+    "addon": aura was imported via WeakAuras.RegisterDisplay (i.e. an addon requested the addition)
+  data: snapshot of data as it was at last import, or nil if source == "user", and the user has not chosen to save their changes to history
+  addon: if source == "addon", then this string indicates which addon supplied the last import
+  skippedVersions: record of versions that the user has indicated they don't wish to see update notifications for
+  allowUpdates: if false, then WeakAuras will ignore any addon-sourced update requests
+  lastUpdate: UNIX timestamp of the last time the user completed an import for this display
+--]]
+
+local WeakAuras = WeakAuras
+local history -- history db upvalue
+
+function WeakAuras.LoadHistory(historydb, ageCutoff)
+  history = historydb
+  if ageCutoff then
+    WeakAuras.ClearOldHistory(ageCutoff)
+  end
+end
+
+function WeakAuras.SetHistory(uid, data, source, addon)
+  if uid and data then
+    history[uid] = history[uid] or {}
+    local hist = history[uid]
+    hist.data = data
+    hist.source = source
+    hist.addon = source == "addon" and (addon or "unknown") or nil
+    hist.skippedVersions = hist.skippedVersions or {}
+    hist.lastUpdate = time()
+    if hist.allowUpdates == nil then
+      hist.allowUpdates = true
+    end
+  end
+end
+
+function WeakAuras.GetHistory(uid)
+  return uid and history[uid]
+end
+
+function WeakAuras.RemoveHistory(uid)
+  if uid then
+    history[uid] = nil
+  end
+end
+
+function WeakAuras.AllowUpdates(uid, allowed)
+  local hist = WeakAuras.Gethistory(uid)
+  if hist then
+    hist.allowUpdates = allowed ~= false
+  end
+end
+
+function WeakAuras.IsUpdateAllowed(uid)
+  local hist = WeakAuras.GetHistory(uid)
+  return hist and hist.allowUpdates
+end
+
+function WeakAuras.SkipVersion(uid, version, skip)
+  local hist = WeakAuras.GetHistory(uid)
+  if hist then
+    hist.skippedVersions[version] = skip ~= false
+  end
+end
+
+function WeakAuras.IsVersionSkipped(uid, version)
+  local hist = WeakAuras.GetHistory(uid)
+  if hist then
+    return hist.skippedVersions[version]
+  end
+end
+
+function WeakAuras.RestoreFromHistory(uid)
+  local hist = WeakAuras.GetHistory(uid)
+  if hist and hist.data then
+    WeakAuras.Add(CopyTable(hist.data))
+  end
+end
+
+function WeakAuras.ClearOldHistory(daysBack, includeNonDeleted)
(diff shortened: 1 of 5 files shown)


emptyrivers:
Currently, this does nothing but add snapshots on every import. Gathering the data is the easiest and least risky part of this project, but takes a long time, since we need users to do stuff. So, I'd like to get this bit released ASAP, and the other pieces of this project can be released when they're done.

WeakAuras/History.lua
```diff
@@ -0,0 +1,88 @@
...
+
+local WeakAuras = WeakAuras
+local history -- history db upvalue
+
+function WeakAuras.LoadHistory(historydb)
+  history = historydb
+  if history.clearOldHistory == nil then
+    history.clearOldHistory = 30
+  end
+  if history.clearOldHistory then
```
mrbuds: this will always be true

WeakAuras/History.lua
```diff
@@ -0,0 +1,88 @@
...
+  if uid and data then
+    history[uid] = history[uid] or {}
+    local hist = history[uid]
+    hist.data = data
+    hist.source = source
+    hist.addon = source == "addon" and (addon or "unknown") or nil
+    hist.skippedVersions = hist.skippedVersions or {}
+    hist.lastUpdate = time()
+  end
+end
```
mrbuds: You are overwriting previous data, don't you want to save it with something like `hist.versions[(hist.version or 0)+1] = hist.data` or using timestamp ?

emptyrivers: No, I don’t. That would bloat the database far more than I’m comfortable doing.

WeakAuras/History.lua
emptyrivers: Hardly. We already have precedent for settings in WeakAuras which aren’t normally editable (see the action squelch timer, or debug flag). I figure that whenever we get around to actually providing a settings panel, this will be a find setting to include.

WeakAuras/History.lua
mrbuds: Alright, so the idea is to have hist.data always have imported data through transmission in the state it was before any user modification, so you can 3-way compare
1- hist.data (last clean import)
2- data (actual aura with whatever user modification was made on it)
3- new clean update data
Am i right?

emptyrivers: In a nutshell, yes.

Stanzilla:
Anything you need tested?

emptyrivers:
@Stanzilla verify that importing will set the history data to the newly updated history and the lastUpdated value of that history entry is current, and also verify that there are no duplicate uids in the database.

WeakAuras/Transmission.lua
```diff
@@ -321,9 +332,12 @@ local function install(data, oldData, patch, mode, isParent)
     if data.uid and data.uid ~= oldData.uid then
       oldData.uid = data.uid
```
InfusOnWoW: Doesn't this need code to adjust the UID to ID mapping?

WeakAuras/History.lua
```diff
@@ -0,0 +1,88 @@
...
+  lastUpdate: UNIX timestamp of the last time the user completed an import for this display
+--]]
+
+local WeakAuras = WeakAuras
+local history -- history db upvalue
+
+function WeakAuras.LoadHistory(historydb)
+  history = historydb
+  if history.clearOldHistory == nil then
+    history.clearOldHistory = 30
```
InfusOnWoW: To me it looks like you store both the history tables, and this history value in the same table, which makes the later iteration "   for uid, hist in pairs(history) do" also pick up this value. That seems unintended.

WeakAuras/WeakAuras.toc
```diff
@@ -22,23 +22,28 @@
 ## SavedVariables: WeakAurasSaved
 ## OptionalDeps: Ace3, LibCompress, LibSharedMedia-3.0, AceGUI-3.0-SharedMediaWidgets, Masque, GTFO, LibButtonGlow-1.0, LibSpellRange-1.0, LibRangeCheck-2.0, LibDBIcon-1.0, LibClassicDurations, LibClassicCast-1.0
 
+# If you add files here, you have to add them to babelfish.lua as well
```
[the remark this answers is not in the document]
emptyrivers: Shhhhh

WeakAuras/History.lua
emptyrivers: Ah, no that is an artifact of when I moved this stuff to its own file. Thanks for checking.

WeakAuras/WeakAuras.lua
```diff
@@ -3371,6 +3385,33 @@ function WeakAuras.Modernize(data)
...
+          prettyPrint("duplicate uid \""..data.uid.."\" detected in saved variables between \""..data.id.."\" and \""..seenUIDs[data.uid].id.."\".")
+        end
+        data.uid = WeakAuras.GenerateUniqueID()
+        seenUIDs[data.uid] = data
+      else
+        seenUIDs[data.uid] = data
+      end
+    elseif data.uid ~= nil then
+      if not silent then
+        prettyPrint("invalid uid detected in saved variables for \""..data.id.."\"")
```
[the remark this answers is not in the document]
emptyrivers: Perhaps we can detect that this is an initial conversion, and if so then silence the check?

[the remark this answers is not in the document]
InfusOnWoW: yeah that might work

WeakAuras/Transmission.lua
emptyrivers: done

WeakAuras/History.lua
```diff
@@ -0,0 +1,88 @@
...
+  return uid and history[uid]
+end
+
+function WeakAuras.RemoveHistory(uid)
+  if uid then
+    history[uid] = nil
+  end
+end
+
+function WeakAuras.AllowUpdates(uid, allowed)
```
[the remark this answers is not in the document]
emptyrivers: No, I just forgot.

InfusOnWoW:
The code looks fine now, and from a quick test doesn't break anything,

Stanzilla:
@emptyrivers ready to merge?

emptyrivers:
History is automatically removed after 30 days, if the aura attached to it has been deleted.

Yes, I do intend to use that orphaned history when a user wishes to restore their stuff.

PR #1524, opened 2019-08-06
Default values for 2 columns widget

### LINKED ISSUE #1517: Tweaks to conditions tab styling
Our problems: 

- https://github.com/WeakAuras/WeakAuras2/issues/1516

As a first measure, rename all Glow Conditions to have the same name as their `Display` tab counterpart

- `Visibility` => `Show Glow`

mrbuds:
# Description
Allow set a default value for the 2 columns widget by setting "true" as 3rd element of "display" table

Fixes #1517

### DIFF
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -69,7 +69,7 @@ local properties = {
     default = 32
   },
   glow = {
-    display = { WeakAuras.newFeatureString .. L["Glow"], L["Show Glow"] },
+    display = { WeakAuras.newFeatureString .. L["Glow"], L["Show Glow"], true },
     setter = "SetGlow",
     type = "bool"
   },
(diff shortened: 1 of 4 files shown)


Stanzilla:
hello darkness my old friend, says the minor version number

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -69,7 +69,7 @@ local properties = {
     default = 32
   },
   glow = {
-    display = { WeakAuras.newFeatureString .. L["Glow"], L["Show Glow"] },
+    display = { WeakAuras.newFeatureString .. L["Glow"], L["Show Glow"], true },
```
emptyrivers: I dislike this for religious reasons, since display names don’t have anything to do with defaults. Is there any way that we can set the default flag on something other than the display name?

emptyrivers: Also some day in the distant future we may have a use for a dropdown tree that is three or more levels deep, and this interferes with that.

mrbuds: i didn't find how to pass along more properties to the ace widget

InfusOnWoW: That isn't ace though, that's the property definitions which get converted to ace tables in ConditionOptions.lua. 

The code is pretty hard to follow though, and it's called display there too....

PR #1535, opened 2019-08-08
fix import error with escape from < and >

Hoizame:
# Description

The signs `<` and `>` break the json import string (error 500).
Escape them fix the problem.
https://travis-ci.org/WeakAuras/WeakAuras2/builds/568530798#L267

## Type of change

# How Has This Been Tested?

- Tested the babelfish.lua itself
- Import tested with an test string ">><>Test<><<" ( on another project )
- Export with bigwigs packager ( on another project )

### DIFF
--- a/babelfish.lua
+++ b/babelfish.lua
@@ -95,6 +95,13 @@ local function parseFile(filename)
     return strings
 end
 
+-- escape < and > for import
+local function getJsonFormattedString(s)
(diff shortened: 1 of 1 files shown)


Stanzilla:
Thank you! Will revert and tell you about it when CurseForge deploys a server side fix!

Issue #1556, opened 2019-08-12
Issue Trying To Delete A Weakaura (Classic Version)


HolteUnderscore:
**Describe the bug**

Unable to delete existing weakauras

**Do you have an error log of what happened?**
 

Message: Interface\AddOns\WeakAuras\WeakAuras.lua:2231: attempt to index field '?' (a nil value)
Time: Mon Aug 12 05:31:51 2019
Count: 20
Stack: Interface\AddOns\WeakAuras\WeakAuras.lua:2231: attempt to index field '?' (a nil value)
[C]: ?
Interface\AddOns\WeakAuras\WeakAuras.lua:2231: in function `Delete'
...terface\AddOns\WeakAurasOptions\WeakAurasOptions.lua:1289: in function `DeleteOption'
...terface\AddOns\WeakAurasOptions\WeakAurasOptions.lua:1315: in function `OnAccept'
Interface\FrameXML\StaticPopup.lua:5003: in function `StaticPopup_OnClick'
[string "*:OnClick"]:1: in function <[string "*:OnClick"]:1>

**To Reproduce**
 

**Which version of WeakAuras are you using?**

Have tried with both 2.14.0-beta4-classic and beta5-classic

**Was it working in a previous version? If yes, which was the last good one?**

Get the same error on both beta 4 and beta 5 for classic

HolteUnderscore:
Disregard this if you like, it seemed to be one particular weakaura that was causing the problem.  I can delete others fine.  Had to delete each child individually in this particular one to get rid of it.

InfusOnWoW:
Without your settings, it's hard to tell what is exactly broken, but we think we know what the problem was and that it's fixed.

PR #1557, opened 2019-08-12
enhance parent-child validation

emptyrivers:
# Description

Once upon a time, there was an addon called SpeedrunSplits.
It was a good addon, used by many, in oldschool WoW.
Then, Blizzard decided to make oldschool WoW new again, and a lot of people were very happy.
The author of SpeedrunSplits decided to make his addon available for Clasasic WoW, and again, a lot of people were very happy.
But then, it was  covered that SpeedrunSplits broke the WeakAuras database accidentally, and a lot of people were very sad.
So, we the developers of WeakAuras will make the database more robust against what broke,
And hopefully people will be happy again.

...More seriously, the bug here was that SpeedrunSplits was replacing the tContains function with another
which had incompatible behavior, thus confusing SyncParentChildRelationships into thinking that groups
were corrupted. This issue has been fixed in SpeedrunSplits beta 2.4 and later, but the damage to the database remains.
So, improve SyncParentChildRelationships to detect the damage and repair it.
Also incidentally remove the vulnerability in the first place.

## Type of change

# How Has This Been Tested?

basic smoke testing

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -3464,10 +3464,11 @@ function WeakAuras.SyncParentChildRelationships(silent)
     if data.parent then
       if data.controlledChildren then
         if not silent then
-          prettyPrint("detected corruption in saved variables: "..id.." has both child and parent.")
+          prettyPrint("detected corruption in saved variables: "..id.." is a group that thinks it's a parent.")
         end
         -- A display cannot have both children and a parent
         data.parent = nil
+        parents[id] = data
       elseif not db.displays[data.parent] then
         if not(silent) then
           prettyPrint("detected corruption in saved variables: "..id.." has a nonexistent parent.")
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -3447,50 +3447,87 @@ function WeakAuras.ValidateUniqueDataIds(silent)
 end
 
 function WeakAuras.SyncParentChildRelationships(silent)
-  local childToParent = {};
-  local parentToChild = {};
+  -- 1. Find all auras where data.parent ~= nil or data.controlledChildren ~= nil
+  --    If an aura has both, then remove data.parent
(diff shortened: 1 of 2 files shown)


emptyrivers:
Also this closes #1556, probably.

WeakAuras/WeakAuras.lua
```diff
@@ -3447,52 +3447,84 @@ function WeakAuras.ValidateUniqueDataIds(silent)
...
+    local groupChildren = {}
+    local childrenToRemove = {}
+    for index, childID in ipairs(data.controlledChildren) do
+      local child = children[childID]
+      if not child then
+        if not silent then
+          prettyPrint("detected corruption in saved variables: "..id.." thinks it controls "..childID.." which doesn't exist.")
         end
+        childrenToRemove[index] = true
+      elseif child.parent ~= id then
```
mrbuds: Shouldn't you do `childrenToRemove[index] = true` for this case ?

emptyrivers: Yes, that is correct. Thanks for noticing :D

WeakAuras/WeakAuras.lua
```diff
@@ -3447,52 +3447,89 @@ function WeakAuras.ValidateUniqueDataIds(silent)
...
+          prettyPrint("detected corruption in saved variables: "..id.." has both child and parent.")
         end
         -- A display cannot have both children and a parent
-        data.parent = nil;
-      elseif(db.displays[data.parent] and db.displays[data.parent].controlledChildren) then
-        childToParent[id] = data.parent;
-        parentToChild[data.parent] = parentToChild[data.parent] or {};
-        parentToChild[data.parent][id] = true;
-      else
+        data.parent = nil
```
InfusOnWoW: In this code path you probably want to add the aura to parents, so that the later consistency checks on data.controlledChildren are done.

emptyrivers: Yup, that sounds right.

Issue #1558, opened 2019-08-12
Raise Text frame level above aurabar border


mrbuds:
**Describe the bug**
![img](https://i.imgur.com/FvTw94q.png)

```
[...]

PR #1563, opened 2019-08-13
Add border subregion

### LINKED ISSUE #1138: Icon Region: border
Icons can't have a border like bars have, it could be use as an indicator for ... anything

Same implementation as bar region

mrbuds:
# Description

Add border subregion for: icon, aurabar, texture, progresstexture

TODO:
- ~~migrate aurabar border~~
- add backdrop options ? (currently it's a feature loss for aurabar) => i'd prefer not, and if we really wanted it i think it would be better as a separate subregion
- ~~use it for dynamic group ?~~ no, maybe later
- ~~frameLevel~~ => let user decide with subregion order seems good to me
- ~~Make "Border X" number independent from other subregion type~~

Fixes #1138, #1558
## Type of change


Stanzilla:
Do we want people adding more than one border?

InfusOnWoW:
I think we probably should limit that to one. As to framelevels, which is on your todo list:
* We either just make use of the automatic frame level determination that is currently implemented, meaning users can choose how to stack their sub elements (which doesn't sound too useful for texts, but might become a bit more interesting with more types)
* We add a flag to the sub region type, that this sub region doesn't take part in the automatic stacking, and set a fixed frame level offset based on region type

mrbuds:
In current state, group & dynamicgroup are not using the border subregion. For them, having backdrop can make sense contrary to aurabar/icon/texture/... I like how border settings are displayed for groups & dynamicgroup.

I'm not happy with the add border subregion button, i wish it could be shown like it's done with groups (minus backdrop) and still be a subregion for the code structure improvement it gives.

WeakAuras/SubRegionTypes/SubBorder.lua
```diff
@@ -0,0 +1,88 @@
+if not WeakAuras.IsCorrectVersion() then return end
```
emptyrivers: Oh, no no no no. We aren’t doing this, please. I relented when @InfusOnWoW named the text subregion filed as SubText, but I am *not* pleased with the idea of naming every damn file in this directory SubSomething. That’s really dumb.

It’s a border, so just call it Border.lua

emptyrivers:
I don’t understand why we wouldn’t want multiple borders. In not too long, I anticipate that we’ll have the ability to add multiple textures to the same aura. It is not unreasonable that a user would want a border on each texture.

WeakAuras/SubRegionTypes/SubBorder.lua
Stanzilla: I agree :D

mrbuds:
I renamed the files.

So big question:
- Allow only 1 border per aura, show it as a subsection of "Icon Settings" / "Progress Bars Settings" / etc..
- Allow only 1 border per aura, add it with the "Add Border" button in the "Extra Elements" section (how it's currently done)
- Unlimited number of borders per aura, added with the "Add Border" button in the "Extra Elements" section

Stanzilla:
I'm not sure, as a first step I would leave it like is currently done.

mrbuds:
After discussion on discord, i have re-added possibility to add infinite number of borders.
And for AuraBar regiontype i have added more anchor option https://i.imgur.com/eoHnbmq.png

WeakAuras/WeakAuras.lua
```diff
@@ -3412,6 +3414,47 @@ function WeakAuras.Modernize(data)
...
+        border_size = data.borderSize,
+      }
+
+      data.border = nil
+      data.borderColor = nil
+      data.borderEdge = nil
+      data.borderOffset = nil
+      data.borderInset = nil
+      data.borderSize = nil
+      data.backdropColor = nil
```
InfusOnWoW: I wonder whether we should leave the backdrop related data in aurabar. If there are users that make use of backdrops, we can readd backdrops and have the right data

WeakAurasOptions/SubRegionOptions/Border.lua
```diff
@@ -0,0 +1,118 @@
+if not WeakAuras.IsCorrectVersion() then return end
+
+local SharedMedia = LibStub("LibSharedMedia-3.0");
+local L = WeakAuras.L;
+
+local screenWidth, screenHeight = math.ceil(GetScreenWidth() / 20) * 20, math.ceil(GetScreenHeight() / 20) * 20;
+
+local function DeleteSubRegion(data, index, regionType)
```
InfusOnWoW: These functions shouldn't be duplicated, we need to find a common place for them.

Also I'm adding duplicate in my PR.

InfusOnWoW:
Overall looks pretty good . One suggestion, and one thing that should be cleaned up. 

I'll have to consider the frame level thing a bit more and figure out to which levels the aurabar assigns, and what we did before.

mrbuds:
I copied what you did in https://github.com/WeakAuras/WeakAuras2/pull/1586/commits/dfc007dfd0dc104f7a375eff0b45ff50cc2afa2e so they will conflict

Stanzilla:
Superseded by https://github.com/WeakAuras/WeakAuras2/pull/1588

Issue #1185, opened 2019-02-13
Optimizing WeakAuras.NewDisplayButton


emptyrivers:
It calls ScanForloads. Why does it do that?
ScanForLoads is hilariously expensive.

The update thread used in Transmission.lua calls this for each installed aura. This is possibly the reason that the whole thing is so damn slow sometimes.

emptyrivers:
So, yeah, I hijacked the Profiling systems to also record the amount of time each `Weakauras.SomeFunction` executes. Then I ran an update on [this](https://wago.io/r1lqyX32M) aura. The results aren't perfect (times over 100%? I probably screwed something up), but check it out:

```

Functions:
Total time attributed to functions: 
NewDisplayButton 13947.53ms, 4833.21%
SortDisplayButtons 11638.35ms, 4033.01%
ScanForLoads 6502.43ms, 2253.27%
afterScanForLoads 5712.21ms, 1979.44%
Update 4093.32ms, 1418.45%
DeleteOption 3569.85ms, 1237.05%
ReloadTriggerOptions 1526.28ms, 528.90%
AddOption 1361.15ms, 471.68%
Add 1284.30ms, 445.05%
SetRegion 1129.75ms, 391.49%
PickDisplay 860.35ms, 298.14%
DoConfigUpdate 815.76ms, 282.68%
GetData 624.17ms, 216.29%
EnsureOptions 512.96ms, 177.76%
ReloadGroupRegionOptions 504.49ms, 174.82%
UpdateDisplayButton 470.31ms, 162.98%
CanHaveDuration 294.04ms, 101.89%
SetThumbnail 257.51ms, 89.23%
GetDisplayButton 232.82ms, 80.68%
IsOptionsOpen 225.61ms, 78.18%
TimerTick 218.09ms, 75.58%
EnsureDisplayButton 204.43ms, 70.84%
ConstructOptions 182.30ms, 63.17%
CheckNumericIds 181.52ms, 62.90%
DeepCopy 180.50ms, 62.55%
GetConditionOptions 157.71ms, 54.65%
IsAnimating 145.77ms, 50.51%
CollapseAllClones 133.47ms, 46.25%
GetOverlayInfo 123.13ms, 42.67%
ReplacePlaceHolders 109.88ms, 38.08%
ShowOptions 106.47ms, 36.90%
Delete 101.54ms, 35.19%
FixGroupChildrenOrderForGroup 96.13ms, 33.31%
[...]

PR #1586, opened 2019-08-18
Smallpatches

InfusOnWoW:


emptyrivers:
All of it looks good to me. The performance improvement for #1185 will introduce some churn, but it's probably not worth worrying about.

WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -4209,15 +4586,12 @@ end
...
 -------------------------------
 
---[[
+
 for _, class in pairs(templates.class) do
   for _, spec in pairs(class) do
-  -- TODO 8.0
-  -- tinsert(spec[5].args, { spell = 195710, type = "ability"}) -- Honorable Medallion
-  -- tinsert(spec[5].args, { spell = 208683, type = "ability", pvptalent = 1}) -- Gladiator's Medallion
+    tinsert(spec[7].args, { spell = 208683, type = "ability", pvptalent = 1}) -- Gladiator's Medallion
```
mrbuds: This doesn't test if `spec[7]` exists which make an error with current master for templates.class.ROGUE.classic

WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -1518,6 +1643,11 @@ templates.class.PRIEST = {
     [7] = {
       title = L["PvP Talents"],
       args = {
+        { spell = 197871, type="ability", pvptalent = 5, titleSuffix = L["cooldown"]},-- Dark Archangel
```
mrbuds: you can add `buff = true` to lines with `type="ability"` to give more choices

InfusOnWoW: I know there are a bunch of things I could have done more for pvp talents, but I already spent far too much on this feature. And there are lots of small issues in there, as I didn't test them all.

InfusOnWoW:
Can't we disable the templates for classic? I don't think we should release them anyway.

Stanzilla:
If some commits are problematic for classic, we can just split the PR?

mrbuds:
I'm not against disabling templates for classic until they are more refined

PR #1588, opened 2019-08-19
Rewrite Border as a sub element

InfusOnWoW:
And add it to icons, textures, progress textures
This drops the backdrop support from progress bars


WeakAuras/WeakAuras.lua
```diff
@@ -420,6 +420,7 @@ function WeakAuras.RegisterSubRegionType(name, displayName, supportFunction, cre
       default = default,
       addDefaultsForNewAura = addDefaultsForNewAura,
       properties = properties,
+      supportsAdd = supportsAdd == nil and true or supportsAdd,
```
emptyrivers:
```suggestion
      supportsAdd = supportsAdd == nil or supportsAdd,
```

mrbuds:
i fixed my SV with 
```Lua
for id, data in pairs(WeakAurasSaved.displays) do
   if data.regionType == "aurabar" then
      data.subRegions = data.subRegions or {}
      local t = data.subRegions
      if not(t[1] and t[1].type == "aurabar_bar") then
         tinsert(data.subRegions, 1, { ["type"] = "aurabar_bar" })
      end
   end
end
```

WeakAuras/WeakAuras.lua
```diff
@@ -3431,9 +3434,12 @@ function WeakAuras.Modernize(data)
       data.borderOffset = nil
       data.borderInset = nil
       data.borderSize = nil
-
-      data.subRegions = data.subRegions or {}
-      tinsert(data.subRegions, border)
+      -- TODO test this conversion code
```
mrbuds: need to add here
```Lua
tinsert(data.subRegions, 1, { ["type"] = "aurabar_bar" })
```

WeakAuras/WeakAuras.lua
```diff
@@ -3905,6 +3906,11 @@ function WeakAuras.PreAdd(data)
     WeakAuras.validate(data, default)
   end
 
+  local regionValidate = data.regionType and WeakAuras.regionTypes[data.regionType] and WeakAuras.regionTypes[data.regionType].validate
+  if regionValidate then
+    regionValidate(data)
```
mrbuds: I'm surprised you don't do it only once with WeakAuras.Modernize, what is the reason for doing it here?

PR #1597, opened 2019-08-25
fix migration

emptyrivers:
Presumably this was done to eliminate nils here. But `false or true` evaluates to true, so anybody who disabled text1 on icons will wake up to a nasty surprise.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -3217,7 +3217,7 @@ function WeakAuras.Modernize(data)
 
       local text1 = {
         ["type"] = "subtext",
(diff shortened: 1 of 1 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -3217,7 +3217,7 @@ function WeakAuras.Modernize(data)
 
       local text1 = {
         ["type"] = "subtext",
-        text_visible = data.text1Enabled or true,
+        text_visible = data.text1Enabled ~= false,
```
emptyrivers: oh, this is probably too clever actually.

PR #1627, opened 2019-08-27
bossmod triggers: fix arithmetic error if state.extend is nil

### LINKED ISSUE #1625: WeakAura Trigger Error
Greetings,
I have a WeakAura that I'm unable to detect where the bug is coming from.
The WeakAura and problem screenshot attacthed. 
https://wago.io/ja4IRIEUD
unknown (1)

mrbuds:
fix #1625

I don't understand how it can happen but it seems it's possible that state.extend can be nil which result in an arithmetic error, use 0 instead of nil if that happen

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -3356,7 +3356,7 @@ WeakAuras.event_prototypes = {
                 end
               end
             elseif event == "DBM_TimerStop" and state then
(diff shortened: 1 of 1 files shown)


GoldennDragonn:
So I just shoud make the changes to WeakAuras/Prototypes.lua as you suggested or it will be fixed in future update of WeakAura addon ?

PR #1629, opened 2019-08-27
fix "fallback" regionType

mrbuds:
Importing a model or a group including one break weakauras

This prevent importing an aura of an unsupported region, but doesn't fix the issue if there is already one imported. I have not found a proper solution for this


mrbuds:
ignore this PR for now, i didn't know about the fallback region

PR #1632, opened 2019-08-28
.luacheckrc wrong local var

lamboley:
Bad copy paste ?


Stanzilla:
Hah, never saw the comment was wrong, thank you!

PR #1642, opened 2019-08-30
classic: WA_GetUnitAura return duration & expirationTime from LibClas…

mrbuds:
…sicDurations

### DIFF
--- a/WeakAuras/AuraEnvironment.lua
+++ b/WeakAuras/AuraEnvironment.lua
@@ -4,6 +4,12 @@ local WeakAuras = WeakAuras
 local L = WeakAuras.L
 local prettyPrint = WeakAuras.prettyPrint
 
+local LCD
+if WeakAuras.IsClassic() then
+  LCD = LibStub("LibClassicDurations")
+  LCD:RegisterFrame("WeakAuras")
+end
+
 local UnitAura = UnitAura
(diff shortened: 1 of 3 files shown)


Hekili:
I'm getting an infinite loop at line 27 (calling `WA_GetUnitAura` inside `WA_GetUnitAura`).

A quick fix is:

```
if WeakAuras.IsClassic() then
  local WA_GetUnitAuraBase = WA_GetUnitAura
  WA_GetUnitAura = function(unit, spell, filter)
    local name, icon, count, debuffType, duration, expirationTime, source, isStealable, nameplateShowPersonal, spellId, canApplyAura, isBossDebuff, castByPlayer, nameplateShowAll, timeMod = WA_GetUnitAuraBase(unit, spell, filter)
    if spellId then
      local durationNew, expirationTimeNew = LCD:GetAuraDurationByUnit(unit, spellId, source, name)
      if duration == 0 and durationNew then
        duration = durationNew
        expirationTime = expirationTimeNew
      end
    end
    return name, icon, count, debuffType, duration, expirationTime, source, isStealable, nameplateShowPersonal, spellId, canApplyAura, isBossDebuff, castByPlayer, nameplateShowAll, timeMod
  end
```

Note that I also nil-checked `spellId` before using LibClassicDurations (that was the next error I found after using `WA_GetUnitAuraBase`).

PR #1648, opened 2019-08-31
Add Experimental Repair Tool

emptyrivers:
# Description

This adds a repair tool. When the database version changes, then a snapshot of each aura is stored into history before running Modernize on that data. If the user then downgrades their WA version past this migration, then WeakAuras detects the change and asks the user to run the repair tool, which will overwrite the database with the migration snapshots.

The user can also manually run the repair tool via /wa repair. This may be handy for fixing bad migrations caught in alpha (though of course, alpha testers should still keep backups 😉).

For now, I would like to mark this feature as experimental, rather than something fully fledged and ready for the most "user" of users to use. I don't believe this will make things worse, but I also don't want to create the expectation that users can just type /wa repair whenever they like and their problems will be magically solved.

## Type of change

# How Has This Been Tested?

Verified that repair tool works as described above. If user declines to repair then login proceeds as before.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -1503,7 +1503,7 @@ Broker_WeakAuras = LDB:NewDataObject("WeakAuras", {
   iconB = 1
 });
 
-local loginFinished, loginMessage = false, L["Options will finish loading after the login process has completed."]
+local loginFinished, loginMessage = false, L["Options will open after the login process has completed."]
 
 function WeakAuras.IsLoginFinished()
   return loginFinished
@@ -1522,6 +1522,8 @@ function WeakAuras.Login(initialTime, takeNewSnapshots)
   local loginThread = coroutine.create(function()
     WeakAuras.Pause();
     local toAdd = {};
+    loginFinished = false
+    loginMessage = L["Options will open after the login process has completed."]
     for id, data in pairs(db.displays) do
       if(id ~= data.id) then
         print("|cFF8800FFWeakAuras|r detected a corrupt entry in WeakAuras saved displays - '"..tostring(id).."' vs '"..tostring(data.id).."'" );
@@ -2630,7 +2632,7 @@ function WeakAuras.ResolveCollisions(onFinished)
     end);
     popup.editBox:SetText(currentId);
     popup.text:SetJustifyH("left");
-    popup.icon:SetTexture("Interface\\Aweddons\\WeakAuras\\Media\\Textures\\icon.blp");
+    popup.icon:SetTexture("Interface\\Addons\\WeakAuras\\Media\\Textures\\icon.blp");
     popup.icon:SetVertexColor(0.833, 0, 1);
 
     UpdateText(popup);
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -2680,7 +2680,6 @@ function WeakAuras.RepairDatabase(loginAfter)
     WeakAuras.SetImporting(true)
     -- set db version to current code version
     db.dbVersion = WeakAuras.InternalVersion()
-    db.lastUpgrade = time()
     -- reinstall snapshots from history
     for id, data in pairs(db.displays) do
       local snapshot = WeakAuras.GetMigrationSnapshot(data.uid)
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -69,7 +69,9 @@ function WeakAuras.LoadOptions(msg)
 end
 
 function WeakAuras.OpenOptions(msg)
-  if (WeakAuras.IsLoginFinished() and WeakAuras.LoadOptions(msg)) then
+  if WeakAuras.NeedToRepairDatabase() then
+    StaticPopup_Show("WEAKAURAS_CONFIRM_REPAIR", nil, nil, {reason = "downgrade"})
+  elseif (WeakAuras.IsLoginFinished() and WeakAuras.LoadOptions(msg)) then
     WeakAuras.ToggleOptions(msg);
   end
 end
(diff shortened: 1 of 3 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -2596,7 +2630,7 @@ function WeakAuras.ResolveCollisions(onFinished)
     end);
     popup.editBox:SetText(currentId);
     popup.text:SetJustifyH("left");
-    popup.icon:SetTexture("Interface\\Addons\\WeakAuras\\Media\\Textures\\icon.blp");
+    popup.icon:SetTexture("Interface\\Aweddons\\WeakAuras\\Media\\Textures\\icon.blp");
```
InfusOnWoW: Awedddons?

emptyrivers: They’re like addons, except awsome.

WeakAuras/WeakAuras.lua
```diff
@@ -2605,6 +2639,61 @@ function WeakAuras.ResolveCollisions(onFinished)
...
+    for id, data in pairs(db.displays) do
+      local snapshot = WeakAuras.GetMigrationSnapshot(data.uid)
+      if snapshot then
+        db.displays[id] = CopyTable(snapshot)
+        coroutine.yield()
+      end
+    end
+    WeakAuras.SetImporting(false)
+    -- finally, login
+    WeakAuras.Login()
```
InfusOnWoW: This can lead to a double call to Login(). E.g. the user log ons, gets a warning messaging but clicks on cancel. That leads to Login(). Then the user calls /wa repair, okays the repair process, which calls Login() here.

I think if the we are already logged in, we should reload the ui. (I think there's a way for addons to trigger that.)

WeakAuras/WeakAuras.lua
```diff
@@ -1506,64 +1505,94 @@ Broker_WeakAuras = LDB:NewDataObject("WeakAuras", {
...
+      end
+      coroutine.yield();
+      nextCallback = loginQueue[1];
+    end
 
-function WeakAuras.IsLoginFinished()
-  return loginFinished
-end
+    loginFinished = true
+    WeakAuras.ResumeAllDynamicGroups();
```
InfusOnWoW: This is problematic if Login() is called twice, see later comment

WeakAuras/WeakAuras.lua
```diff
@@ -2605,6 +2639,61 @@ function WeakAuras.ResolveCollisions(onFinished)
...
+  button2 = L["Cancel"],
+  OnAccept = function(self)
+     WeakAuras.RepairDatabase()
+  end,
+  OnShow = function(self)
+    if self.data.reason == "user" then
+      self.text:SetText(L["Manual Repair Confirmation Dialog"]:format(WeakAuras.LastUpgrade()))
+    else
+      self.text:SetText(L["Automatic Repair Confirmation Dialog"]:format(WeakAuras.LastUpgrade()))
+      self.button2:SetText(L["Continue Without Repairing"])
```
InfusOnWoW: The buttons didn't appear to resize with the text for me, this was overshot both sides.

WeakAuras/WeakAuras.lua
emptyrivers: I tried that, it doesn't work if we use coroutines Double logins don't seem to have ill effects as best as i can tell.

InfusOnWoW:
From a quick test this worked as expected. A few minor things that I noticed while reviewing.

WeakAuras/WeakAuras.lua
```diff
@@ -1506,64 +1505,94 @@ Broker_WeakAuras = LDB:NewDataObject("WeakAuras", {
...
-    WeakAuras.CheckForPreviousEncounter()
-  end
-  coroutine.yield();
-  WeakAuras.RegisterLoadEvents();
-  WeakAuras.Resume();
-  coroutine.yield();
+    isDatabaseUpgrade = takeNewSnapshots;
+    WeakAuras.AddMany(toAdd);
+    coroutine.yield();
+    WeakAuras.AddManyFromAddons(from_files);
```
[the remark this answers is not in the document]
emptyrivers: I intend to leave it unchanged until I get around to revamping that system.

WeakAuras/WeakAuras.lua
```diff
@@ -1506,64 +1505,94 @@ Broker_WeakAuras = LDB:NewDataObject("WeakAuras", {
...
 
-  -- check in case of a disconnect during an encounter.
-  if (db.CurrentEncounter) then
-    WeakAuras.CheckForPreviousEncounter()
-  end
-  coroutine.yield();
-  WeakAuras.RegisterLoadEvents();
-  WeakAuras.Resume();
-  coroutine.yield();
+    isDatabaseUpgrade = takeNewSnapshots;
```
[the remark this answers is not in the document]
emptyrivers: That's how I was originally going to implement it. After a week, I don't really recall why i decided against it.

WeakAuras/WeakAuras.lua
InfusOnWoW: Actually it might be okay, seems ResumeAllDynamicGroups doesn't decrease below 0, so it shouldn't matter.

Although I question that ResumeAllDynamicGroups, I thought that dynamic groups were created suspended, and that's why we do that. But I can't find any code that suspends newly created dynamic groups, not code that initially suspend it.

emptyrivers:
If you want to restrict it to e.g. alpha testers to reduce the size of the database, I'd be ok with that.

InfusOnWoW:
Ah that's a interesting idea, but looking back at 2.14, while it probably would have benefited a few testers, the main benefit would have been for the actual release.

emptyrivers:
Fair enough. If people complain about SV size then we could always advise them to adjust the time that migration snapshots live before being wiped.

Stanzilla:
@emptyrivers ready to merge then?

emptyrivers:
✔️

PR #1676, opened 2019-09-08
don't package modelpaths for WeakAuras classic

emptyrivers:
since we don't support models currently in classic WA


Stanzilla:
You could add the two files to the ignore list as well

Issue #1700, opened 2019-09-15
Can't open weakaura.


Laukagen:
**Describe the bug**
So earlier today I tried to make a healer mana tracker, and wanted to shorten the amount of decimals in an number connected to this progress bar. I didn't know how, so I looked for it on reddit, but after posting this:
function()

    local p = math.max(0, UnitPower("player")) / math.max(1, UnitPowerMax("player")) * 100;
    return string.format("%.0f%%", p);
end

and using TSU, everything went down and i'm unable to open the optionsmenu. 
When I reboot the game, Weakaura opened in paused mode, but nothing happens when I try to resume.

**Do you have an error log of what happened?**
1x WeakAuras\GenericTrigger.lua:3125: attempt to index local 'variables' (a function value)
WeakAuras\GenericTrigger.lua:3125: in function <WeakAuras\GenericTrigger.lua:3104>
(tail call): ?
WeakAuras\WeakAuras-2.14.6.lua:4813: in function <WeakAuras\WeakAuras.lua:4795>
(tail call): ?
WeakAuras\WeakAuras.lua:4900: in function `GetOverlayInfo'
WeakAuras\RegionTypes\AuraBar.lua:126: in function `propertiesFunction'
WeakAuras\WeakAuras.lua:1005: in function `GetProperties'
WeakAuras\WeakAuras.lua:1135: in function `ConstructConditionFunction'
WeakAuras\WeakAuras.lua:4010: in function <WeakAuras\WeakAuras.lua:3956>
WeakAuras\WeakAuras.lua:4062: in function `Add'
WeakAuras\WeakAuras.lua:3616: in function `load'
WeakAuras\WeakAuras.lua:3623: in function `AddMany'
WeakAuras\WeakAuras.lua:1523: in function <WeakAuras\WeakAuras.lua:1511>

**To Reproduce**
1. Post the code
2. Weakauras stops working.

**Screenshots**
![image](https://user-images.githubusercontent.com/55357330/64924753-18cd4100-d7e8-11e9-8d75-90b35613bfc2.png)

**Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?**
Yes

**Which version of WeakAuras are you using?**
2.14.6 Classic

**Additional context**
Since i'm kinda clueless on what happened, i'm uploading both the weakaura file, and the string i worked on, with my mana bar.

[...]

PR #1701, opened 2019-09-15
Fix lua error on wrong input into variables

InfusOnWoW:
Fixes: #1700


WeakAuras/GenericTrigger.lua
```diff
@@ -3135,7 +3135,7 @@ function GenericTrigger.GetOverlayInfo(data, triggernum)
     if (trigger.custom_type == "stateupdate") then
       local count = 0;
       local variables = events[data.id][triggernum].tsuConditionVariables;
-      if (variables) then
+      if (variables and type(variables) == "table") then
```
emptyrivers:
```suggestion
      if (type(variables) == "table") then
```
if variables is a table, then it will also be truthy

Issue #1723, opened 2019-09-23
An other nil error in WeakAuras.UnloadAll


mrbuds:
When creating an aura using last revision
```
1x WeakAuras\WeakAuras-@project-version@.lua:2138: bad argument #1 to 'pairs' (table expected, got nil)
[C]: in function `pairs'
WeakAuras\WeakAuras-@project-version@.lua:2138: in function `UnloadAll'
WeakAuras\WeakAuras-@project-version@.lua:2130: in function `ReloadAll'
...dOns\WeakAurasOptions\OptionsFrames\OptionsFrame.lua:197: in function <...dOns\WeakAurasOptions\OptionsFrames\OptionsFrame.lua:174>
[C]: in function `Hide'
WeakAurasOptions\WeakAurasOptions-@project-version@.lua:1450: in function <WeakAurasOptions\WeakAurasOptions.lua:1448>

Locals:
(*temporary) = nil
(*temporary) = "table expected, got nil"
 = <function> defined =[C]:-1

```

PR #1724, opened 2019-09-23
Fix ranged swing timer on retail + UnloadAll bugfix

mrbuds:
# Description
- Re-add old behavior for ranged swing timer on retail
- Fix nil error in UnloadAll #1723 

## Type of change

# How Has This Been Tested?
Tested with a lvl 1 hunter on retail

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -1608,14 +1608,22 @@ do
         local currentTime = GetTime();
         local speed = UnitRangedDamage("player");
         if(lastSwingRange) then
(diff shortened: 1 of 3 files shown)


mrbuds:
I forgot to add this little clip showing why i removed the ranged swing timer for retail https://clips.twitch.tv/PrettyGlutenFreeMarrowCorgiDerp
You can see if you are at max range the timing is wrong because of travel time.

mrbuds:
I changed behavior to use UNIT_SPELLCAST_SUCCEEDED like on classic, result is much more accurate.
https://clips.twitch.tv/SpikyMushyTildePhilosoraptor

PR #1761, opened 2019-10-12
Smallerpatches

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -1168,7 +1168,7 @@ WeakAuras.load_prototype = {
       display = L["Zone Name"],
       type = "string",
       init = "arg",
-      events = {"ZONE_CHANGED", "ZONE_CHANGED_INDOORS", "ZONE_CHANGED_NEW_AREA"}
+      events = {"ZONE_CHANGED", "ZONE_CHANGED_INDOORS", "ZONE_CHANGED_NEW_AREA", "VEHICLE_UPDATE"}
```
mrbuds: Registering VEHICLE_UPDATE on classic will raise an error

InfusOnWoW: It's not registered.

mrbuds: Indeed

PR #1816, opened 2019-11-16
Classic: Use LibClassicDurations's UnitAuraDirect instead of UnitAura

### LINKED ISSUE #1809: [Classic] Show Enemy Buffs
Currently, a simple trigger like type:aura, unit:target, aura type:buff does not seem to work on enemies. I would love if this could be changed.

Apparently, LibClassicDurations has the ability to show buffs on enemies. probably by tracking combat log. this works with other addons, for example, i can see enemy buffs on my nameplates (using threatplates together with classicauradurations). WA uses the same lib, but the trigger described above does not seem to work. could this be made to work?

i can not really think of an alternative, other than doing exactly what LibClassicDurations seems to do.

**Additional context**
none

mrbuds:
fixes #1809

UnitAura("target", 1) always return nil if i target an enemy.
LibClassicDurations implement it's own UnitAura replacement that works with ennemies buffs.

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -61,11 +61,13 @@ if not WeakAuras.IsCorrectVersion() then return end
 -- Lua APIs
 local tinsert, wipe = table.insert, wipe
 local pairs, next, type = pairs, next, type
+local UnitAura = UnitAura
 
 local LCD
 if WeakAuras.IsClassic() then
   LCD = LibStub("LibClassicDurations")
   LCD:RegisterFrame("WeakAuras")
+  UnitAura = LCD.UnitAura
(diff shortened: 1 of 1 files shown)


mrbuds:
@rgd87 can you take a look at this PR to make sure it's not using your library wrong ?

Stanzilla:
@mrbuds want to wait for him to respond or merge?

d87:
> UnitAura = LCD.UnitAura

That's wrong. You should use ```UnitAura = LCD.UnitAuraWithBuffs``` or ```LCD.UnitAuraDirect```
LCD.UnitAura is for colon notation

mrbuds:
Indeed, i forget to push that change, thank you!

PR #1824, opened 2019-11-20
Classic Templates

### LINKED ISSUE #1562: Classic Templates
TODO after classic release:

Remove code that merge spells from retail specs
Add these spells to the "classic" section

Needs to be done before next xpac

mrbuds:
# Description
Done
- Priest
- Mage
- Hunter
- Warlock
- Rogue
- Shaman
- Warrior
- Paladin
- Druid

Fixes #1562

### DIFF
--- a/WeakAurasTemplates/TriggerTemplates.lua
+++ b/WeakAurasTemplates/TriggerTemplates.lua
@@ -1,5 +1,3 @@
-if WeakAuras.IsClassic() then return end
-
 -- Special layout for the New Aura Trigger template page
 
 local AceGUI = LibStub("AceGUI-3.0");
@@ -48,6 +46,11 @@ local function changes(property, regionType)
       value = colors[property],
       property = regionColorProperty[regionType],
     };
+  elseif property == "glow" and (regionType == "icon" or regionType == "aurabar") then
+    return {
+      value = true,
+      property = regionType == "icon" and "sub.1.glow" or "sub.2.glow",
+    };
   elseif WeakAuras.regionTypes[regionType].default[property] == nil then
     return nil;
   elseif property == "cooldownSwipe" then
@@ -64,11 +67,6 @@ local function changes(property, regionType)
     return {
       value = false,
       property = "inverse",
-    };
-  elseif property == "glow" then
-    return {
-      value = true,
-      property = "glow",
     };
   end
 end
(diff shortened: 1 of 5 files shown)


mrbuds:
Most abilities were added
Some classes could use more love for their buffs & debuffs categories
The "form" property is not used to make a stance/form trigger; instead a "usable = true" property is added to make it "blue" with conditions if in wrong form. This is not ideal ...
I don't have a druid high level enough on classic to get the right numbers for druid's forms, so i used 1 for cat, and {2, 3} for bear forms, these are placeholder

WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -5051,32 +5586,40 @@ for regionType, regionData in pairs(WeakAuras.regionOptions) do
...
-              args[itemIndex] = item
-            end
-          end
-        end
-      end
+-- remove from tables data for wrong version of the game
+for classId, classData in pairs(templates.class) do
+  for id, data in pairs(classData) do
+    if (tonumber(id) and WeakAuras.IsClassic())
+    or (id == "classic" and not WeakAuras.IsClassic())
```
InfusOnWoW: The structure of the templates looks hacky. As there's a classic table, but not a retail table. And those sections only exist in the class tables

Can you explain why you did it that way?

mrbuds: I wanted to make it easier to re-use existing functions, but i'm re-organizing it to make it less hacky

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -824,63 +883,28 @@ local function subTypesFor(item, regionType)
...
       title = L["Show Only if Buffed"],
       description = L["Only shows the aura if the target has the buff."],
       createTriggers = function(triggers, item)
         createBuffTrigger(triggers, 1, item, "showOnActive", true);
       end,
-      data = { inverse = false },
+      data = data,
     });
+    if subglowindex then
+      data.subRegions = { [subglowindex] = subglow }
```
InfusOnWoW: I don't understand this. Isn't the only difference between data and dataGlow that dataGlow has the glow sub region?
So essentially, aren't after this data and dataGlow identically?

WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -121,4243 +119,4898 @@ local pvpAzeriteTraits = {
...
-        { spell = 287731, type = "buff", unit = "player"}, --Empyreal Ward
-      },
-      icon = 135349
-    },
-    [6] = {},
-    [7] = {
-      title = L["PvP Talents"],
-      args = {
-        { spell = 236186, type="ability", pvptalent = 4},-- Cleansing Light
-        { spell = 247675, type="ability
```
InfusOnWoW: Minor nitpick: A comment here that retail follows would be nice

WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -4616,12 +5268,144 @@ end
...
+  for i = 1, 3 do
+    tinsert(templates.class.DEATHKNIGHT[i][8].args, createSimplePowerTemplate(6));
+
+    tinsert(templates.class.DEATHKNIGHT[i][8].args, {
+      title = L["Runes"],
+      icon = "Interface\\Icons\\spell_deathknight_frozenruneweapon",
+      triggers = {[1] = { trigger = { type = "status", event = "Death Knight Rune", unevent = "auto"}}}
+    });
+  end
+else
```
InfusOnWoW: Same here, the if is 100 lines above.

WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -904,17 +928,19 @@ local function subTypesFor(item, regionType)
       createConditions = function(conditions, item, regionType)
         missingBuffGreyed(conditions, 1, regionType);
       end,
-      data = { inverse = false },
+      data = data,
```
InfusOnWoW: Haven't tested this exact code path "Always shows the aura, grey if buff not active." does indeed include a unused glow sub region. So I believe the above code is a left-over?

PR #1839, opened 2019-12-06
Re-add Range Check to Classic.

nullKomplex:
# Description

Re-implements Range Check library to Classic, as the library has since been updated for Classic compatibility.

## Type of change

# How Has This Been Tested?

- [x] Created a Range Check aura for target. Used Conditions to check several different ranges.

# Checklist:

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -24,26 +24,24 @@ function WeakAuras.IsSpellInRange(spellId, unit)
   return SpellRange.IsSpellInRange(spellId, unit)
 end
 
-if not WeakAuras.IsClassic() then
-  local LibRangeCheck = LibStub("LibRangeCheck-2.0")
+local LibRangeCheck = LibStub("LibRangeCheck-2.0")
 
-  function WeakAuras.GetRange(unit, checkVisible)
-    return LibRangeCheck:GetRange(unit, checkVisible);
+function WeakAuras.GetRange(unit, checkVisible)
+  return LibRangeCheck:GetRange(unit, checkVisible);
+end
+
+function WeakAuras.CheckRange(unit, range, operator)
+  local min, max = LibRangeCheck:GetRange(unit, true);
+  if (type(range) ~= "number") then
+    range = tonumber(range);
   end
-
-  function WeakAuras.CheckRange(unit, range, operator)
-    local min, max = LibRangeCheck:GetRange(unit, true);
-    if (type(range) ~= "number") then
-      range = tonumber(range);
-    end
-    if (not range) then
-      return
-    end
-    if (operator == "<=") then
-      return (max or 999) <= range;
-    else
-      return (min or 0) >= range;
-    end
+  if (not range) then
+    return
+  end
+  if (operator == "<=") then
+    return (max or 999) <= range;
+  else
+    return (min or 0) >= range;
   end
 end
 
@@ -6343,7 +6341,6 @@ if WeakAuras.IsClassic() then
   WeakAuras.event_prototypes["Death Knight Rune"] = nil
   WeakAuras.event_prototypes["Alternate Power"] = nil
   WeakAuras.event_prototypes["Equipment Set"] = nil
-  WeakAuras.event_prototypes["Range Check"] = nil
   WeakAuras.event_prototypes["Spell Activation Overlay"] = nil
   WeakAuras.event_prototypes["Totem"] = nil
 else
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6341,7 +6341,7 @@ WeakAuras.event_prototypes = {
         display = L["Unit"],
         type = "unit",
         init = "unit",
-        values = "actual_unit_types_with_specific",
+        values = "unit_types_range_check",
         test = "true",
         store = true
       },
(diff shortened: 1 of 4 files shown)


emptyrivers:
LGTM, but I want buds to take a look before merging.

mrbuds:
I had same PR ready for this but decided against re-adding the lib because i think it isn't accurate enough for most use case where someone ask for it (like for example get the exact range of the dead zone for hunters in pvp)

The trigger for classic could use a warning explaining it's limits before we get reports on how inaccurate it is

nullKomplex:
Honestly I was considering reworking the number imput to a drop-down that lets you select ranges, but I figured I'd go one step at a time as I figure this stuff out.

mrbuds:
I like this idea

nullKomplex:
I looked into it and could not come up with a solution that I was happy with for a drop down. The library uses 3 separate lists: Friendly, Harmful, and Miscellaneous. From my testing I discovered that friendly NPCs do not generate the same results as friendly players, either. The list appears to be dynamically adjusted based on what it can find from items and _spells_. The latter of which I find rather important, as it means different classes would then have different available ranges from which to check.

I didn't like the idea of restricting a user to what they can currently use, as there are many aura creators that would like to choose settings for more than the class they are currently on. I also disliked the idea of creating separate lists for the different range checks that the library can utilize.

Instead of a convoluted option system I instead opted for a dynamically updated list of ranges listed at the bottom of the trigger:

https://i.imgur.com/oA9DiT0.png

WeakAuras/Prototypes.lua
```diff
@@ -45,6 +45,54 @@ function WeakAuras.CheckRange(unit, range, operator)
...
+  end
+  lastvalue = nil
+  for _, key in pairs(misc) do
+    if lastvalue then
+      miscString = (miscString and (miscString .. ", ") or "") .. (lastvalue .. "-" .. key)
+    end
+    lastvalue = key
+  end
+  RangeCacheStrings.friend, RangeCacheStrings.harm, RangeCacheStrings.misc = friendString, harmString, miscString
+end
```
nullKomplex: There is 0 doubt in my mind that this approach will be denied, but I don't know of a better way to go about it and I don't know how else to learn but to submit my attempt.

WeakAuras/Prototypes.lua
```diff
@@ -6326,6 +6374,12 @@ WeakAuras.event_prototypes = {
           return state and state.show and WeakAuras.CheckRange(state.unit, needle, needle2);
         end,
       },
+      {
+        name = "rangeList",
+        type = "description",
+        display = "|cFFFFA500It is only possible to check if a unit is within the following ranges:",
+        text = function() return L["|cFFAAFFAAFriendly Units:|r|n%s|n|cFFFFAAAAHarmful Units:|r|n%s|n|cFFAAAAFFMiscellanous Units:|r|n%s|n|nThis list is based on your current character."]:format(RangeCacheStrings.friend or "", RangeCacheStrings.harm or "", RangeCacheStrings.misc or "") end
+      },
```
InfusOnWoW: There's already "L["Note: This trigger type estimates the range to the hitbox of a unit. The actual range of friendly players is usually 3 yards more than the estimate."]," above, adding this text to that is prefarable. 

Also, please less colors.

If the text contains the ranges that the lib provides, it should be prefaced with a bit more information. E.g. something like this:

The range checking capabilities depend on the current spec, class, items and type of unit that is checked. Generally the following ranges are always supported: 8, 40 [I don't actually know what this set is].

For your current setup [????], the following ranges can be checked: 
Friendly Units: 1,2,3,4,5,7,8,10, 15, 20, 25, 30, 35, 38, 40, 45, 50, ...
Harmful Units: ...
Other Units: ...
[Isn't the big distinction also players vs npcs?]

I don't think there's any value in value in showing ranges, the input us after all just one range.

nullKomplex: Well I didn't want to add it to the above description because it was long and seemed unsightly. I understand having a warning at the top (is that 3 yard thing even accurate?) but I don't think UI/UX-wise that the list should come before the actual functional part of the trigger personally. Even better that it's located directly under the range option.

The lib itself doesn't have a distinction in its data based on NPC/Player, so I could not generate that from its data. I used its exact internal setup. I would have to either test the values manually or monitor in-game activity to dynamically update the list based on unit type. The goal was to future proof the list, though.

Also, less colors or no colors? D:

WeakAuras/Prototypes.lua
```diff
@@ -45,6 +45,54 @@ function WeakAuras.CheckRange(unit, range, operator)
   end
 end
 
+local RangeCacheRawData = {friend = {}, harm = {}, misc = {}}
```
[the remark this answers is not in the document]
nullKomplex: I was trying to reduce garbage, but that may not be worth having to use `wipe` anyway...

InfusOnWoW:
Both of these notes are about the meaning of the range check, them being separate serves no purpose. If we add that list, then we need to explain what it actually is. Otherwise it's confussing.

Ui wise, the description should come before the setting, as that's just the natural order or reading.

InfusOnWoW:
As far as I'm concerned, it looks fine to me. @mrbuds you are our dedicated chief classic maintainer, so you should have the final say.

PR #1843, opened 2019-12-09
Add "Current Action" trigger for classic fixes #1628

### LINKED ISSUE #1628: Maul/Heroic strike visible toggle
I am unable to get druid maul (and warrior heroic strike) to visibly show that it has been queued/toggled in weakauras

There should be something that looks for this state so that the weak aura icon can visibly change when queued/toggled

I tried many ways to get it to work but was unable to

mrbuds:
# Description

Add a new trigger "Current Action" for queued spells

img

## Type of change

# How Has This Been Tested?

Test on a warrior with heroic strike

before use: https://i.imgur.com/iCK1PYw.jpg
when used: https://i.imgur.com/0Pf4ewa.jpg
queued but no more rage: https://i.imgur.com/3aimgfl.jpg

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6215,6 +6215,47 @@ WeakAuras.event_prototypes = {
     automaticrequired = true
   },
 
+  ["Queued Action"] = {
+    type = "status",
+    events = {
+      ["events"] = {"ACTIONBAR_UPDATE_STATE"}
+    },
+    internal_events = {
+      "ACTIONBAR_SLOT_CHANGED",
+      "ACTIONBAR_PAGE_CHANGED"
+    },
+    name = L["Queued Action"],
+    init = function(trigger)
+      local ret = [=[
+        local spellid = select(7, GetSpellInfo(%q))
+        local button
+        if spellid then
+            local slotList = C_ActionBar.FindSpellActionButtons(spellid)
+            button = slotList and slotList[1]
+        end
+      ]=]
+      return ret:format(trigger.spellName or "")
+    end,
+    args = {
+      {
+        name = "spellName",
+        required = true,
+        display = L["Spell"],
+        type = "spell",
+        test = "true",
+      },
+      {
+        hidden = true,
+        test = "button and IsCurrentAction(button)";
+      },
+    },
+    iconFunc = function(trigger)
+      local _, _, icon = GetSpellInfo(trigger.spellName or 0);
+      return icon;
+    end,
+    automaticrequired = true
+  },
+
   ["Range Check"] = {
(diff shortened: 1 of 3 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -6215,6 +6215,47 @@ WeakAuras.event_prototypes = {
...
+      "ACTIONBAR_SLOT_CHANGED",
+      "ACTIONBAR_PAGE_CHANGED"
+    },
+    name = L["Current Action"],
+    init = function(trigger)
+      local _, _, icon, _, _, _, spellid = GetSpellInfo(trigger.spellName)
+      local button
+      if icon and spellid then
+          icon = icon
+          local slotList = C_ActionBar.FindSpellActionButtons(spellid)
```
emptyrivers: This is a neat API. Do you happen to know how it handles macros?

mrbuds: On my warrior i only have a macro for heroic strike in my action bars and it works

emptyrivers: Cool. We might be able to actually give people a solution when they want their auras to have keybind text.

WeakAuras/Prototypes.lua
```diff
@@ -6215,6 +6215,47 @@ WeakAuras.event_prototypes = {
...
+          button = slotList[1]
```
emptyrivers: This value isn't used

WeakAuras/Prototypes.lua
emptyrivers: Also, add C_ActionBar to .luacheckrc

WeakAuras/Prototypes.lua
mrbuds: ho wait i uploaded wrong version

WeakAuras/Prototypes.lua
```diff
@@ -6215,6 +6215,47 @@ WeakAuras.event_prototypes = {
     automaticrequired = true
   },
 
+  ["Current Action"] = {
```
InfusOnWoW: I think we need a better name.
Discord search is broken, but we should look up what terminology people are using when they ask for that functionality.

InfusOnWoW: It seems "queued" is the right word?

InfusOnWoW: Actually probably best to change this here too, as long as we can still do that.

(Note, also need to change the    WeakAuras.event_prototypes["Current Action"] = nil line)

mrbuds:
I have added the removal of the totem trigger for classic, totem api was removed with last patch

PR #1856, opened 2019-12-14
classic: workaround GetSpellCount change in 1.13.03 #1849

### LINKED ISSUE #1849: spellcount not working since latest patch in classic dec 10th 2019
## Describe the bug

i'll use blind(rogue) as an example, i used spellcount in my text options to show many casts i am able to do(basically # of reagents on me) it was all working perfectly prior to the bg patch, and as soon as i got the update it stopped working.

no only reporting other addons

### To Reproduce

1. import string
2. look at bottom right of square for the text(number)
3. make sure to have blinding powder in inventory
4. see no display of spellcount
5. open /wa and close to see 0 instead of nothing, but still incorrect

**Screenshots**

https://imgur.com/ZD5FFwL

yes

2.15.5 also in screenshot
- [ X] Classic
- [ ] Retail

it was always working fine, just until the BG patch in classic that went live on Dec 10th, 2019 that this started happening.
### Additional Info

```
https://wago.io/nWxm-__3b
```

mrbuds:
# Description
Work around GetSpellCount returning 0 for all spells since last classic update

copy/pasta from https://github.com/tullamods/Dominos/blob/master/Dominos/bars/getActionReagentUses.lua adapted to use localized spell name instead of spell ids

Fixes #1849

ps: it's highly possible a lib doing this will be release

### DIFF
--- a/.pkgmeta-classic
+++ b/.pkgmeta-classic
@@ -22,6 +22,7 @@ externals:
   WeakAuras/Libs/LibGetFrame-1.0: https://github.com/mrbuds/LibGetFrame
   WeakAuras/Libs/!LibTotemInfo: https://github.com/SwimmingTiger/LibTotemInfo
   WeakAuras/Libs/LibRangeCheck-2.0: https://repos.wowace.com/wow/librangecheck-2-0/trunk/LibRangeCheck-2.0
+  WeakAuras/Libs/LibClassicSpellActionCount-1.0: https://github.com/Ennea/LibClassicSpellActionCount-1.0
 
 enable-nolib-creation: no
 
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -73,6 +73,10 @@ local aceEvents = WeakAurasAceEvents
 local WeakAuras = WeakAuras;
 local L = WeakAuras.L;
 local GenericTrigger = {};
+local LCSA
+if WeakAuras.IsClassic then
+  LCSA = LibStub("LibClassicSpellActionCount-1.0")
+end
 
 local event_prototypes = WeakAuras.event_prototypes;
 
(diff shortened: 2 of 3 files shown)


Vardex:
What about this lib? https://github.com/Ennea/LibClassicSpellActionCount-1.0

mrbuds:
slot doesn't work, function need to take spell name

mrbuds:
but yeah that could be added to this lib

InfusOnWoW:
That's ugly....

WeakAuras/Types.lua
```diff
@@ -2223,4 +2223,168 @@ if WeakAuras.IsClassic() then
   end
 
   WeakAuras.glow_types.ACShine = nil
+
+  -- from https://github.com/tullamods/Dominos/blob/master/Dominos/bars/getActionReagentUses.lua
+  WeakAuras.spell_reagents = {
+    -- Druid
+    -- Gift of the Wild
+    [GetSpellInfo(21849)] = 17021,
```
emptyrivers: spell data is asynchronously loaded, I thought. So I would expect that in some cases this'd cause a `table index is nil` error, and break things.

mrbuds:
LibClassicSpellActionCount-1.0 author may add functions suitable for this use case
https://github.com/Ennea/LibClassicSpellActionCount-1.0/issues/1#issuecomment-566003178

InfusOnWoW:
Okay, someone else having to deal with the uglyness is better :)

Though I the more I see the wow classic devs work, the less sure am I that they know what they are doing.

mrbuds:
They change their API like if they didn't know that addons devs will reproduce same behavior in a less effective way.

Stanzilla:
@mrbuds sup with this?

mrbuds:
Lib was updated to add the missing function needed, i'll work on it

mrbuds:
Redid this using the lib, package tested on classic, and it's working fine

PR #1864, opened 2019-12-17
reposition children of dynamic group on unitframes on GETFRAME_REFRES…

mrbuds:
…H callback

# Description
LibGetFrame refresh its list of unitframes 1s after a list of events
PLAYER_REGEN_DISABLED
PLAYER_REGEN_ENABLED
PLAYER_ENTERING_WORLD
GROUP_ROSTER_UPDATE

I added possibility on LibGetFrame to register a callback with CallbackHandler-1.0 on these events to reposition children of dynamic groups with options "group by frame" + "unitframe"
https://github.com/mrbuds/LibGetFrame/commit/f0a0d6bb9e22ff01df2884c99147fdd997a34755

## Type of change

I'd like to test this a bit more, i'm not confident the way it's implemented in DynamicGroup.lua is right

### DIFF
--- a/WeakAuras/RegionTypes/DynamicGroup.lua
+++ b/WeakAuras/RegionTypes/DynamicGroup.lua
@@ -949,9 +949,11 @@ local function modify(parent, region, data)
   end
 
   if data.useAnchorPerUnit and data.anchorPerUnit == "UNITFRAME" then
-    LGF.RegisterCallback("WeakAuras", "GETFRAME_REFRESH", function()
+    LGF.RegisterCallback("WeakAuras" .. data.uid, "GETFRAME_REFRESH", function()
       region:PositionChildren()
     end)
+  else
+    LGF.UnregisterCallback("WeakAuras" .. data.uid, "GETFRAME_REFRESH")
   end
 
   function region:DoPositionChildrenPerFrame(frame, positions)
--- a/WeakAuras/RegionTypes/DynamicGroup.lua
+++ b/WeakAuras/RegionTypes/DynamicGroup.lua
@@ -2,6 +2,7 @@ if not WeakAuras.IsCorrectVersion() then return end
 
 local WeakAuras = WeakAuras
 local SharedMedia = LibStub("LibSharedMedia-3.0")
+local LGF = LibStub("LibGetFrame-1.0")
 
 local default = {
   controlledChildren = {},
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
The WeakAuras part of the patch is fine, but on the LGF side there's no need for a 1s delay.

You can be pretty sure that the unit frames immediately respond to these events. And do that in a OnUpdate handler in the next frame. 

That is try: In your OnEvent handler, install a OnUpdate handler that calls ScanForUnitFrames(true) + deinstalls the OnUpdate handler again.

InfusOnWoW:
The other way to do it, is to give unit frame addons hooks to tell LGF that they have updated themselves.

WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -947,6 +948,12 @@ local function modify(parent, region, data)
     end
   end
 
+  if data.useAnchorPerUnit and data.anchorPerUnit == "UNITFRAME" then
+    LGF.RegisterCallback("WeakAuras", "GETFRAME_REFRESH", function()
+      region:PositionChildren()
+    end)
+  end
```
InfusOnWoW: You need to deregister in the else branch here.

mrbuds: fixed

mrbuds:
> 
> 
> The WeakAuras part of the patch is fine, but on the LGF side there's no need for a 1s delay.
> 
> You can be pretty sure that the unit frames immediately respond to these events. And do that in a OnUpdate handler in the next frame.
> 
> That is try: In your OnEvent handler, install a OnUpdate handler that calls ScanForUnitFrames(true) + deinstalls the OnUpdate handler again.

The unitframe addon i use is vuhdo and it doesn't change the layout of it's frames on GROUP_ROSTER_UPDATE, but instead set a timer to 0.6s, which is handled in an OnUpdate loop, i guess this is done to avoid re-doing the layout many times at once when WoW burst GROUP_ROSTER_UPDATE events when joining a new group.

WeakAuras/RegionTypes/DynamicGroup.lua
InfusOnWoW: I don't see a else branch.

PR #1866, opened 2019-12-20
Add Relative Offset Property Change

### LINKED ISSUE #1561: Condition Property Change to change relative position of auras
It's a common problem (and frustration) that the current Condition Offsets Property Changes move the aura to the exact same place every time. Users often report that when they move an aura they've imported and close the options panel that their aura appears to reset its position to the same place every time, despite its X and Y coordinates being proper. In addition, setting a static location works well in a group, where the position is already relative to where you place the group, but I think that auras outside of a group could use a better option.

Therefore I propose the addition of a Relative Positioning Condition Property Change.

An additional set of Condition Property Changes should be added to adjust the x an y offsets of an aura, relative to the location specified under the Display panel. Could be named Relative X-Offset or Adjust X-Offset.

The setting would simply move the aura the specified number of "pixels" away from its position specified in the Display tab. These effects would likely be additive, but could also be exclusive.

It should be able to work in tandem with the original Offset condition.

An alternative could be also renaming X-Offset in the current Condition Property Changes, to something like Set X Position (or even just Set X-Offset).

**Additional context**
Afenar.

Also, while technically accurate, the name of X and Y Offsets bothers me with the advent of this new potential condition. Potential for rename?

nullKomplex:
# Description

Adds a Property Change that allows the user to adjust the position of a Region relative to its current location. Renames current similar Property Changes to "Absolute X/Y Offset" to help clearly indicate the difference between the two.

Fixes #1561

Initially I wanted to have the Property Changes be additive (such that one Condition could add 10, and another could add 10, for a change in 20) but I realized this was inconsistent with existing interactions (or lack thereof) between Property Changes.

## Type of change

# How Has This Been Tested?

- [x] Used all of the affected Conditions in an aura. Reloads performed.
- [x] Tested interaction with Animations.

# Checklist:

### DIFF
--- a/WeakAuras/RegionTypes/RegionPrototype.lua
+++ b/WeakAuras/RegionTypes/RegionPrototype.lua
@@ -168,7 +168,7 @@ function WeakAuras.regionPrototype.AddProperties(properties, defaultsForRegion)
     type = "customcode"
   }
   properties["xOffset"] = {
-    display = L["X-Offset"],
+    display = L["Absolute X-Offset"],
     setter = "SetXOffset",
     type = "number",
     softMin = -screenWidth,
@@ -176,7 +176,7 @@ function WeakAuras.regionPrototype.AddProperties(properties, defaultsForRegion)
     bigStep = 1
   }
   properties["yOffset"] = {
-    display = L["Y-Offset"],
+    display = L["Absolute Y-Offset"],
     setter = "SetYOffset",
     type = "number",
     softMin = -screenHeight,
--- a/WeakAuras/RegionTypes/RegionPrototype.lua
+++ b/WeakAuras/RegionTypes/RegionPrototype.lua
@@ -183,6 +183,22 @@ function WeakAuras.regionPrototype.AddProperties(properties, defaultsForRegion)
     softMax = screenHeight,
     bigStep = 1
   }
+  properties["xOffsetRelative"] = {
+    display = WeakAuras.newFeatureString .. L["Relative X-Offset"],
+    setter = "SetXOffsetRelative",
+    type = "number",
+    softMin = -screenWidth,
+    softMax = screenWidth,
+    bigStep = 1
+  }
+  properties["yOffsetRelative"] = {
+    display = WeakAuras.newFeatureString .. L["Relative Y-Offset"],
+    setter = "SetYOffsetRelative",
+    type = "number",
+    softMin = -screenHeight,
+    softMax = screenHeight,
+    bigStep = 1
+  }
 
   if (defaultsForRegion and defaultsForRegion.alpha) then
     properties["alpha"] = {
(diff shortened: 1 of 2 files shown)


nullKomplex:
I don't know enough about the inner workings of WeakAuras to know if either there is more I need to change (my tests gave the exact results I was looking for) or if I put anything anywhere that had repercussions beyond the scope of my intended change.

emptyrivers:
This immediately raises the question, what if I make two relative offset conditions? There’s an argument that they “should” compose instead of overwrite.

But then, that would be unique in our conditions. I don’t know what’s best.

InfusOnWoW:
Right, so a couple of thouhgts.

The issue motivates this fix by noting that some auras use conditions to move auras to a different position. (I think Afenar is pretty fond of that.) And that this prevents drag the auras to a different place.

Now, pretty clearly this PR does only help if users can be taught to use relative positioning instead of absolute positioning. And then it's not exactly a good solution. As entering a exact position would still position the aura at some "random" offset to what the user entered. (Random for the user, as they are unaware of the conditions.)

So this is at best a small upgrade. On the other hand, relative offsets is what I probably should have implemented as it's probably much more useful then absolute positioning. And that makes the PR interesting.

In your issue you mentioned making the property changes additive, what rivers suggested too.

That wouldn't be totally unique, if it's understood as a "action". That's how play sound/send chat works. That would be "Move aura by" then combining them would be natural. Obviously resetting needs to happen on aura hiding.

But still I don't think that would really fit into how conditions work and thus I'm not convinced that it would be a good idea. That's mostly to prefering consistency, and that I don't see a huge benefit to being able to compose multiple, and the potential confussion when the user creates them in a way that repeately moves the aura by a certain amount.

So, my suggestion so far:
* Migrate existing absolute position property changes to relative offsets and remove the old property.

nullKomplex:
Wait, so you don't want absolute offsets properties at all? It's not the end of the world to not have them I suppose.

InfusOnWoW:
I do see a big benefit of not having two very similar properties and I believe conversion is possible.

nullKomplex:
At this point I don't understand the system well enough to know if I made the right choices here.

Should I have used the same properties? Should I have used the same functions? Search doesn't reveal them being used anywhere else, but are they? How else are their positions controlled?

When modernizing, this will clear offset conditions used for alpha users that have already modernized to `25`. What should I do in this case? Let the users deal with the broken conditions, progress 2 versions in the same update, or ignore version checking when modernizing (or maybe version == 25)?

emptyrivers:
The way our migrations work, is that you increment the internalVersion value at the top of the main file (shown below) and increment it https://github.com/WeakAuras/WeakAuras2/blob/e09d3bf8d8925b9ee97f58be761009d7327af7ce/WeakAuras/WeakAuras.lua#L1

Now, you have space to apply your data changes in this function: https://github.com/WeakAuras/WeakAuras2/blob/e09d3bf8d8925b9ee97f58be761009d7327af7ce/WeakAuras/WeakAuras.lua#L2786-L3726

This ensures that every data change is applied in a predictable order.

InfusOnWoW:
So, what you were experiencing was most likely due to how conditions are untriggered. If no condition changes the a property, then that property needs to be set to its base value.

That base value is retrieved from data.

Thus if you name your property just XOffset, then if no condition applies, data.xOffset is the base value, and that's incorrect. Thus keeping the new property with a new name is fine.

I'm not 100% convinced that "Relative X Offset" is a particular good name, but I don't have any good suggestion either.

PR #1868, opened 2019-12-23
Custom Options Author Names Tweaks

### LINKED ISSUE #1847: Custom Options - Unnameable option types
## Describe the bug

Unnameable option types like Space and Description show as "Option #5" (etc.) in a collapsed list of settings making it hard to navigate. Would want to see them labelled as their option type. This is a bug report rather than a feature request as the thought was that this should already be the case.

nullKomplex:
# Description

Correctly displays intended fallback name when using a noninteractive display type.
Uses fallback name when name field is an empty string, in Author Mode _only_.
Removes inconsistency between default names and fallback names.

Fixes #1847

## Type of change

# How Has This Been Tested?

- [x] Used every noninteractive Custom Option type.
- [x] Named Custom Options a blank name. Checked in and out of Author Mode.
- [x] Created new options.

# Checklist:

### DIFF
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -196,7 +196,7 @@ local function descType(option)
     L["Used in Auras:"]
   }
   for id, optionData in pairs(option.references) do
-    tinsert(desc, ("%s - Option %i"):format(id, optionData.path[#optionData.path]))
+    tinsert(desc, ("%s - Option #%i"):format(id, optionData.path[#optionData.path]))
   end
   return tconcat(desc, "\n")
 end
@@ -1305,7 +1305,7 @@ typeControlAdders = {
           childOption.subOptions[j] = {
             type = "toggle",
             key = "subOption" .. j,
-            name = L["Sub Option %i"]:format(j),
+            name = L["Sub Option #%i"]:format(j),
             default = false,
             width = 1,
             useDesc = false,
@@ -1696,7 +1696,7 @@ function addAuthorModeOption(options, args, data, order, prefix, i, isSubOption)
           -- mostly because it would have a very non-intuitive effect
           -- the names and keys would likely not match anymore, and so
           -- the merged display would basically explode into a bunch of separate options
-          childOption.name = childOption.name or ("Option %i"):format(i)
+          childOption.name = childOption.name or ("Option #%i"):format(i)
           if not childOption.key then
             local newKey = "option" .. i
             local existingKeys = {}
@@ -2468,7 +2468,7 @@ function WeakAuras.GetAuthorOptions(data, args, startorder)
             childData.authorOptions[i] = {
               type = "toggle",
               key = "option" .. i,
-              name = L["Option %i"]:format(i),
+              name = L["Option #%i"]:format(i),
               default = false,
               width = 1,
               useDesc = false,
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
Rivers is afk for a week, and since this touches mostly his code we'll probably won't merge it until then.

I'll take a look and review it though in the next days.

InfusOnWoW:
As far as I see that's all looks good.

emptyrivers:
All of the `name ~= ""` checks don’t sit well with me. If the goal is to always show something in case the author is insane enough to enter an empty string as an option’s name, then I would instead enforce a non-blank name, than trying to defend and thus getting highly verbose code.

emptyrivers:
(also your defenses fail against an author who puts a space as the name)

InfusOnWoW:
A empty name can happen if the author wants to opt into automatic names after previously naming a option. Thus it's not so much guarding against insane authors, but enabling them to go back to automatic names.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1420,7 +1420,7 @@ local function ensureNonDuplicateKey(option)
   end
 end
 
-function addAuthorModeOption(options, args, data, order, prefix, i)
+function addAuthorModeOption(options, args, data, order, prefix, i, isSubOption)
```
emptyrivers: If we do decide we want this then instead of having a flag, have an integer which represents the depth of recursion. A depth > 0 indicates that this is some sub option, otherwise it is not.

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -1451,8 +1451,8 @@ function addAuthorModeOption(options, args, data, order, prefix, i)
     buttonWidth = buttonWidth + 0.15
   end
   local optionClass = optionClasses[option.type]
-  local optionName = option.class == "noninteractive" and WeakAuras.author_option_types[option.type]
-                     or option.name or L["Option #%i"]:format(i)
+  local optionName = optionClass == "noninteractive" and WeakAuras.author_option_types[option.type]
```
emptyrivers: Since this is fixed, it should be an invariant that the option name we display is never falls back to “Option #i”. So remove that part, it concealed this bug in the first place.

WeakAurasOptions/AuthorOptions.lua
emptyrivers: Also unless I’m mistaken the value is never set anywhere so you may as well remove it.

emptyrivers:
After reading this pull request more carefully I realize that it was a deliberate decision to make that empty string fallback. So, I have to ask, why? Is it really that common to want a blank option name? If so, would it not make more sense to accommodate that need with a setting, for better discovery?

nullKomplex:
I'll review it when I can (on vacation) in case I made a mistake that
changed my intention. The intention was to supply a clean and readable way
for the author to view their options while modifying them if they choose to
not have a visible name for the options for user mode. I did this in the
same spirit that the original bug was posted (the comment about author
mode).

The actual option name should still be whatever the author chooses to
display. And if they want to circumvent the author mode name I figured it
wasn't worth the few extra CPU cycles to stop that. Only author mode was
intended to have changes.

nullKomplex:
My apologies, but I won't be able to do those changes before 8.3, and I know you guys want a release soon so I've reverted all non-bugfix changes except for the one you requested.

Stanzilla:
@emptyrivers Ping

emptyrivers:
Oh right. One moment.

PR #1885, opened 2020-01-04
remember update preferences for each aura

### LINKED ISSUE #1149: Update Auras as default
Currently the default import mode is Create a Copy.
If the Update Auras mode is generally safe to use then I suggest making that the default, as I think the use case is usually more fit to that; its less likely that someone wants to keep a copy of the old auras to tinker with them, they are more likely updating the fast changing auras like Causese's T23 one.

Edit: how do I change the label? :)

emptyrivers:
# Description

Closes #1149

## Type of change

# How Has This Been Tested?

i tested and it worked

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -108,6 +108,16 @@ function GenerateUniqueID()
 end
 WeakAuras.GenerateUniqueID = GenerateUniqueID
 
+local function stripNonTransmissableFields(datum, fieldMap)
+  for k, v in pairs(fieldMap) do
+    if type(v) == "table" and type(datum[k]) == "table" then
+      stripNonTransmissableFields(datum[k], v)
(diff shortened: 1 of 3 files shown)


WeakAuras/Transmission.lua
```diff
@@ -1064,7 +1075,13 @@ local function scamCheck(codes, data)
   end
 end
 
-local ignoredForDiffChecking = WeakAuras.internal_fields
+local ignoredForDiffChecking = {}
```
InfusOnWoW: There's actual a Utility function in Util.lua that can do those loops in one line:

local ignoredForDiffChecking = Mixin({}, WeakAuras.internal_fields, WeakAuras.non_transmissable_fields)

emptyrivers: Right, I forgot that exists.

PR #1915, opened 2020-01-17
Smallpatches

InfusOnWoW:


mrbuds:
I'm having issues with encounter journal getting bugged after i looked at "Azeroth" and looked at encounter_id tooltip.
It's most probably not new, but servers just went down, i'll test more when it's back online.

Otherwise looks good

PR #1923, opened 2020-01-20
Tick SubRegion

### LINKED ISSUE #1656: TickMark Progress Bar
So, there's no way that i know of, to add tickmarks to certain points of a progress bar, without codes. Am i missing something? if not, i would really love this feature. Thanks

nullKomplex:
# Description

Adds a simplistic Tick SubRegion (or Sub Element) to Aura Bar type auras. Feature allows for progress values (optionally in percentage form).

image

Fixes #1656

## Type of change

# How Has This Been Tested?

- [x] Created aura with CD Progress and Custom triggers, tested all orientations and inverse for bar and all tick settings.
- [x] Tested Conditions.

# Checklist:

### DIFF
--- a/WeakAuras/SubRegionTypes/Tick.lua
+++ b/WeakAuras/SubRegionTypes/Tick.lua
@@ -106,7 +106,7 @@ local funcs = {
         offset = (tonumber(percent) / 100) * width
       else
         local pixels = width / (self.parent.state.duration or self.parent.state.total or 1)
-        offset = tonumber(placement) * pixels
+        offset = math.max((tonumber(placement) * pixels), width)
       end
     end
 
--- a/WeakAuras/SubRegionTypes/Tick.lua
+++ b/WeakAuras/SubRegionTypes/Tick.lua
@@ -12,7 +12,6 @@ local default = function(parentType)
     automatic_length = true,
     tick_thickness = 2,
     tick_length = 30,
-    tick_hide_mode = "NEVER",
   }
 end
 
@@ -55,12 +54,6 @@ local properties = {
     min = 0,
     bigStep = 1,
     default = 30,
-  },
-  tick_hide_mode = {
-    display = L["Placement Mode"],
-    setter = "SetTickPlacementMode",
-    type = "list",
-    values = WeakAuras.tick_hide_modes,
   },
 }
 
--- a/WeakAuras/SubRegionTypes/Tick.lua
+++ b/WeakAuras/SubRegionTypes/Tick.lua
@@ -91,6 +91,17 @@ local funcs = {
   ["Update"] = function(self)
     self:SetTickPlacement(self.tick_placement)
   end,
+  ["OrientationChanged"] = function(self)
+    self:SetTickPlacement(self.tick_placement)
+    self:SetTickLength(self.tick_length)
+    self:SetTickThickness(self.tick_thickness)
+  end,
+  ["OnSizeChanged"] = function(self)
+    print("size changed")
+    self:SetTickPlacement(self.tick_placement)
+    self:SetTickLength(self.tick_length)
+    self:SetTickThickness(self.tick_thickness)
+  end,
   SetVisible = function(self, visible)
     if visible then
       self:Show()
@@ -102,12 +113,20 @@ local funcs = {
     self.texture:SetColorTexture(r, g, b, a or 1)
   end,
   SetTickPlacement = function(self, placement)
+    self.parentOrientation = self.parent.effectiveOrientation
+    self.parentInverse = self.parent.inverseDirection
+    if (self.parentOrientation == "VERTICAL") or (self.parentOrientation == "VERTICAL_INVERSE") then
+      self.parentTrueHeight, self.parentTrueWidth = self.parent.bar:GetRealSize()
+    else
+      self.parentTrueWidth, self.parentTrueHeight = self.parent.bar:GetRealSize()
+    end
     local offset, offsetx, offsety = self.tick_placement, 0, 0
     local width = self.parentTrueWidth
-    if self.tick_placement_mode == "NUMERIC" then
+    print(offset, width, self.parentTrueHeight)
+    if self.tick_placement_mode == "ABSOLUTE" then
       local pixels = width / (self.parent.state.duration or self.parent.state.total or 1)
       offset = math.max((placement * pixels), width)
-    elseif self.tick_placement_mode == "PERCENT" then
+    elseif self.tick_placement_mode == "RELATIVE" then
       offset = (placement / 100) * width
     end
 
@@ -120,11 +139,12 @@ local funcs = {
       offsetx = offset
     end
     local side = self.parentInverse and auraBarTrueLeftInverse or auraBarTrueLeft
-
+    print(offset, offsetx, offsety)
     self:ClearAllPoints()
     self:SetPoint("CENTER", self.parent.bar, side[self.parentOrientation], offsetx, offsety)
   end,
   SetTickThickness = function(self, thickness)
+    self.parentOrientation = self.parent.effectiveOrientation
     if (self.parentOrientation == "VERTICAL") or (self.parentOrientation == "VERTICAL_INVERSE") then
       self:SetHeight(thickness)
     else
@@ -132,6 +152,7 @@ local funcs = {
     end
   end,
   SetTickLength = function(self, length)
+    self.parentOrientation = self.parent.effectiveOrientation
     if self.automatic_length then length = self.parentTrueHeight end
     if (self.parentOrientation == "VERTICAL") or (self.parentOrientation == "VERTICAL_INVERSE") then
       self:SetWidth(length)
@@ -144,7 +165,7 @@ local funcs = {
 local function modify(parent, region, parentData, data, first)
 
   region:SetParent(parent)
-  region.parentOrientation = parentData.orientation
+  region.parentOrientation = parent.effectiveOrientation
   region.parentInverse = parentData.inverse
   if (region.parentOrientation == "VERTICAL") or (region.parentOrientation == "VERTICAL_INVERSE") then
     region.parentTrueHeight, region.parentTrueWidth = parent.bar:GetRealSize()
@@ -159,6 +180,7 @@ local function modify(parent, region, parentData, data, first)
   --for k, v in pairs(parent.state) do print(k, v) end
 
   region.parent = parent
+  region.parentData = parentData
   region.tick_visible = data.tick_visible
   region.tick_color = data.tick_color
   region.tick_placement_mode = data.tick_placement_mode
@@ -174,7 +196,8 @@ local function modify(parent, region, parentData, data, first)
   region:SetTickLength(data.tick_length)
 
   parent.subRegionEvents:AddSubscriber("Update", region)
-  --parent.subRegionEvents:AddSubscriber("FrameTick", region)
+  parent.subRegionEvents:AddSubscriber("OrientationChanged", region)
+  parent.subRegionEvents:AddSubscriber("OnSizeChanged", region)
 end
 
 local function supports(regionType)
--- a/WeakAuras/SubRegionTypes/Tick.lua
+++ b/WeakAuras/SubRegionTypes/Tick.lua
@@ -39,6 +39,12 @@ local properties = {
     type = "number",
     validate = WeakAuras.ValidateNumeric,
   },
+  automatic_length = {
+    display = L["Automatic Length"],
+    setter = "SetAutomaticLength",
+    type = "bool",
+    defaultProperty = true,
+  },
   tick_thickness = {
     display = L["Thickness"],
     setter = "SetTickThickness",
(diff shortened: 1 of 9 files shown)


WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,182 @@
...
+  end,
+  SetTickPlacement = function(self, placement)
+    local offset, offsetx, offsety = self.tick_placement, 0, 0
+    local width = self.parentTrueWidth
+    if self.tick_placement_mode == "STATIC" then
+      local percent = string.match(placement, "(%d+)%%")
+      if percent then
+        offset = (tonumber(percent) / 100) * width
+      else
+        local pixels = width / (self.parent.state.duration or self.parent.state.total or 1)
```
nullKomplex: This highlights a fairly substantial flaw, there is no max duration in situations such as a Cooldown Progress trigger that has not fired. This causes (intentionally through 101) it to stick to its last known location. If the trigger fires at least once it will be "fine" (assuming you [i]want[/i] the tick to remain there), but I'm not sure if it's even possible to handle situations where it hasn't fired. Having a default can look awkward in any situation.

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,182 @@
...
+    end
+  end,
+  SetTickHeight = function(self, height)
+    if self.automatic_height then height = self.parentTrueHeight end
+    if (self.parentOrientation == "VERTICAL") or (self.parentOrientation == "VERTICAL_INVERSE") then
+      self:SetWidth(height)
+    else
+      self:SetHeight(height)
+    end
+  end,
```
nullKomplex: I made the conscious decision to keep width always the direction the bar is moving and height the perpendicular. In the aura bar's option changing orientation causes the height and width values to flip, but I don't see a clean way of replicating that behavior in a sub region.

nullKomplex:
Still a WIP. I have some thoughts about changes (ditching the placement modes entirely, for now, and actually implementing hide modes) to make.

WeakAuras/SubRegionTypes/Tick.lua
InfusOnWoW: Then "width" and "height" are not the right words. Both in the user interface and in the code

There are two ways to do this:
Use a generic word, such as "Perpendicular", "Major/Minor", "Across" as the Width/Height

Adjust the naming to whatever the current orientation is. So if its horizontal, the names are height/width, whereas otherwise they are width/height.

For the code, I think widht/height aren't ideal either.

WeakAuras/SubRegionTypes/Tick.lua
InfusOnWoW: If there's no (or a infinite) duration, then imho ticks should be hidden in the STATIC/Absolute mode. So basically a tick is hidden if either it's explicitly or implicitly hidden.

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,182 @@
...
+}
+
+local function modify(parent, region, parentData, data, first)
+
+  region:SetParent(parent)
+  region.parentOrientation = parentData.orientation
+  region.parentInverse = parentData.inverse
+  if (region.parentOrientation == "VERTICAL") or (region.parentOrientation == "VERTICAL_INVERSE") then
+    region.parentTrueHeight, region.parentTrueWidth = parent.bar:GetRealSize()
```
InfusOnWoW: This is not good enough. The aurabar can resize itself via conditions/animations.

You should get the size in SetTickReplacement, and in modify, do:
region:SetAllPoints(parent)
region:SetScript("OnSizeChanged", function(self) self:SetTickPlacement(...) end

Also to your great annoyance, AuraBars can change their orientation via SetOrientation.
You need to introduce a new "event" that is Notified in SetOrientation and subscribed too.

WeakAuras/Types.lua
```diff
@@ -363,6 +363,18 @@ WeakAuras.spark_hide_types = {
   BOTH  = L["Full/Empty"]
 }
 
+WeakAuras.tick_placement_modes = {
+  AUTO = L["Automatic"],
```
InfusOnWoW: I don't understand what AUTO is supposed to do.

mrbuds:
Are you planning to migrate all options from "Spark" to the tick sub region ?

emptyrivers:
I think that migrating the spark options onto this tick subregion would be a mistake, actually. A spark is a texture that happens to be on a moveable anchor point, but ticks are a way of segmenting a progress bar in ways that are meaningful in gameplay terms.

Marrying those two concepts together would limit us in a big way.

emptyrivers:
To expand on this a bit, consider that the ticks are currently supported only on the aurabar region. I can tell you right now that (assuming that ticks are a feature people actually use) we're going to get requests to extend support of this to other region types. So, let's see how many other ways that users might ask for extensions of ticks:

- User asks for ticks support on progress textures.
- User asks for ticks support on circular progress textures.
- User asks for ticks support on icons, since they show progress via the cooldown spiral. This would likely be turned down, I think.
- User asks for ticks to be mask textures, so that they can have transparent space between the non-ticked parts of progress.
- User asks for a way to change the color of progress in sections between ticks that are "full". 
    - At this point, ticks are much more than perhaps we expected them to be with this initial implementation. Perhaps we missed the mark on the true problem we were trying to solve (which happens a lot, honestly. It's part of why development is never actually finished.)

So (assuming you agree with me), all of these requests don't seem so unreasonable for a user, because ticks are not simply a shorthand to decorate an aura with some textures, but rather a way to improve the "at a glance" usability of any aura that displays some kind of progress information, by segmenting the visuals into chunks that are meaningful on a gameplay level. One can imagine that in WeakAuras 2.43.2, we might have the ability to segment a bar, or progress texture, or even circular progress texture, in any of a number of different styles. It makes sense to me to anticipate that to some degree, and leave some space to explore ticks as a way of segmenting progress, instead of just a way to decorate it.

Sparks, on the other hand (as we currently support them, anyways), are a visual aid to show the current progress level. That need would be quite well served by an anchor point option on the moving bit of progress auras. And a texture subregion, of course. This is not to say that it will never make sense to marry ticks and sparks together, but I think that it would be premature to do so in this initial implementation.

WeakAuras/Types.lua
```diff
@@ -366,6 +366,11 @@ WeakAuras.spark_hide_types = {
   BOTH  = L["Full/Empty"]
 }
 
+WeakAuras.tick_placement_modes = {
+  NUMBER = L["Number Value"],
+  PERCENT = L["Percentage"],
```
InfusOnWoW, line 371: I think "Relative" and "Absolute" are better fitting names.

emptyrivers:
```
Checking WeakAuras/RegionTypes/AuraBar.lua        1 warning

    WeakAuras/RegionTypes/AuraBar.lua:976:5: accessing undefined variable 'self'
```

InfusOnWoW:
I haven't forgotten about this, it was just that the last weekend was pretty busy with a lot of new tickets. I'm hoping to get to this soonish though.

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,232 @@
...
+    self.tick_color[1], self.tick_color[2], self.tick_color[3], self.tick_color[4] = r, g, b, a or 1
+    self.texture:SetColorTexture(r, g, b, a or 1)
+  end,
+  SetTickPlacementMode = function(self, placement_mode)
+    if self.placement_mode ~= placement_mode then
+      self.placement_mode = placement_mode
+      self:SetTickPlacement(self.tick_placement)
+    end
+  end,
+  SetTickPlacement = function(self, placement, forced)
```
InfusOnWoW, line 129: Though, I prefer if those funcitons had no forced parameter by splitting the functions into:

SetTickPlacement:
    if self.placement_mode ~= placement_mode then
      self.placement_mode = placement_mode
      self:UpdateTickPlacement()
    end

OrientationChanged()
  self.parentOrientation = self.parent.effectiveOrientation
  self.parentTrueWidth, self.parentTrueHeight = ...
  UpdateTickPlacement()

OnSizeChanged
  self.parentTrueWidth, self.parentTrueHeight = ...
  UpdateTIckPlacement()

UpdateTickPlacement()
    Basically using self's data to place the tick at the right place.

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,232 @@
...
+    if forced or self.tick_placement ~= placement then
+      self.tick_placement = placement
+      self.parentOrientation = self.parent.effectiveOrientation
+      self.parentInverse = self.parent.inverseDirection
```
InfusOnWoW, line 133: Unfortunately inverseDirection is more complex than this:
There are two "inverses" for progress bars, and they cancel out if both are set.
(Ignoring the third inverse via the orientation)

a) The inverse setting set in the options of the aura, this is inverseDirection
b) A inverse setting of the duration information, this is used by e.g. the Cast trigger to automatically inverse the direction of the progress bar for channels. (Whereas by default a cast fills up a cast bar, a channel depletes.)

There are two bugs here:
* This second inverse is currently not taking into account, you can easily figure out whether the duration is inversed by looking copying state.inverse in Update
* The SetInverse function needs to Notify you that the Inverse flag has changed

Also I would prefer it if you introduced simple getters in the progress bar for these properties:
self.parent:GetEffectiveOrientation + self.parent:GetInverseDirection()

(I often forget to do that, but it's cleaner to use functions instead of directly using the variables.)

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,232 @@
...
+      local offset, offsetx, offsety = self.tick_placement, 0, 0
+      local width = self.parentTrueWidth
+      if self.tick_placement_mode == "ABSOLUTE" then
+        local pixels = width / (self.parent.state.duration or self.parent.state.total or 1)
+        offset = math.max(math.min((placement * pixels), width), 0)
+      elseif self.tick_placement_mode == "RELATIVE" then
+        offset = (placement / 100) * width
+      end
+
+      if self.parentInverse ~= ((self.parentOrientation == "HORIZONTAL_INVERSE") or (self.parentOrientation == "VERTICAL")) then
```
InfusOnWoW, line 148: I wonder whether this cleaner as:
local inverse = self.parentInverse
if((self.parentOrientation == "HORIZONTAL_INVERSE") or (self.parentOrientation == "VERTICAL")) then
  inverse = not inverse
end

if inverse then
  offset = -offset
end

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -0,0 +1,232 @@
...
+      self:SetTickPlacement(self.tick_placement)
+    end
+  end,
+  SetTickPlacement = function(self, placement, forced)
+    if forced or self.tick_placement ~= placement then
+      self.tick_placement = placement
+      self.parentOrientation = self.parent.effectiveOrientation
+      self.parentInverse = self.parent.inverseDirection
+      if (self.parentOrientation == "VERTICAL") or (self.parentOrientation == "VERTICAL_INVERSE") then
+        self.parentTrueHeight, self.parentTrueWidth = self.parent.bar:GetRealSize()
```
InfusOnWoW, line 135: True doesn't add any information to the name. And in fact half of the time it's not the true height/width anyway.

WeakAurasOptions/SubRegionOptions/Tick.lua
```diff
@@ -0,0 +1,103 @@
...
+    __delete = function()
+      if (WeakAuras.ApplyToDataOrChildData(parentData, WeakAuras.DeleteSubRegion, index, "subtick")) then
+        WeakAuras.ReloadOptions2(parentData.id, parentData)
+      end
+    end,
+    tick_visible = {
+      type = "toggle",
+      width = WeakAuras.normalWidth,
+      name = L["Show Tick"],
+      order = order + 0.01,
```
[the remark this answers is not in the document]
nullKomplex: I just copied what you had in another one...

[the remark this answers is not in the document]
InfusOnWoW: Oh I'll fix those. Then it probably just works, still I prefer integers.

mrbuds:
Error when i chose trigger "Item Count"

```
1x WeakAuras\SubRegionTypes\Tick.lua:165: Division by zero
[string "@WeakAuras\SubRegionTypes\Tick.lua"]:165: in function `UpdateTickPlacement'
[string "@WeakAuras\SubRegionTypes\Tick.lua"]:107: in function `?'
[string "@WeakAuras\RegionTypes\RegionPrototype.lua"]:33: in function `Notify'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6059: in function <WeakAuras\WeakAuras.lua:6056>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6142: in function <WeakAuras\WeakAuras.lua:6109>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6226: in function `UpdatedTriggerState'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:5996: in function `UpdateFakeStatesFor'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:5969: in function `FakeStatesFor'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:4460: in function <WeakAuras\WeakAuras.lua:4348>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:4480: in function `Add'
[string "@WeakAurasOptions\WeakAurasOptions-@project-version@.lua"]:3476: in function <WeakAurasOptions\WeakAurasOptions.lua:3474>
[string "@WeakAurasOptions\WeakAurasOptions-@project-version@.lua"]:3503: in function <WeakAurasOptions\WeakAurasOptions.lua:3494>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceConfig-3.0-3\AceConfigDialog-3.0\AceConfigDialog-3.0-79.lua"]:51: in function <...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0.lua:49>
[string "@CorruptionTooltips\Libs\AceConfig-3.0-3\AceConfigDialog-3.0\AceConfigDialog-3.0-79.lua"]:843: in function <...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0.lua:664>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:72: in function <...ns\CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0.lua:70>
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:306: in function `Fire'
[string "@CorruptionTooltips\Libs\AceGUI-3.0-41\widgets\AceGUIWidget-DropDown.lua"]:442: in function <...ps\Libs\AceGUI-3.0\widgets\AceGUIWidget-DropDown.lua:433>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:72: in function <...ns\CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0.lua:70>
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:306: in function `Fire'
[string "@AzeriteTooltip\libs\AceGUI-3.0-41\widgets\AceGUIWidget-DropDown-Items.lua"]:351: in function <...s\AceGUI-3.0\widgets\AceGUIWidget-DropDown-Items.lua:341>
```

Error when i chose "Action Usable" trigger
```
1x WeakAuras\SubRegionTypes\Tick.lua:166: attempt to perform arithmetic on local 'percent' (a nil value)
[string "@WeakAuras\SubRegionTypes\Tick.lua"]:166: in function `UpdateTickPlacement'
[string "@WeakAuras\SubRegionTypes\Tick.lua"]:107: in function `?'
[string "@WeakAuras\RegionTypes\RegionPrototype.lua"]:33: in function `Notify'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6059: in function <WeakAuras\WeakAuras.lua:6056>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6142: in function <WeakAuras\WeakAuras.lua:6109>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:6226: in function `UpdatedTriggerState'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:5996: in function `UpdateFakeStatesFor'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:5969: in function `FakeStatesFor'
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:4460: in function <WeakAuras\WeakAuras.lua:4348>
[string "@WeakAuras\WeakAuras-@project-version@.lua"]:4480: in function `Add'
[string "@WeakAurasOptions\WeakAurasOptions-@project-version@.lua"]:3476: in function <WeakAurasOptions\WeakAurasOptions.lua:3474>
[string "@WeakAurasOptions\WeakAurasOptions-@project-version@.lua"]:3503: in function <WeakAurasOptions\WeakAurasOptions.lua:3494>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceConfig-3.0-3\AceConfigDialog-3.0\AceConfigDialog-3.0-79.lua"]:51: in function <...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0.lua:49>
[string "@CorruptionTooltips\Libs\AceConfig-3.0-3\AceConfigDialog-3.0\AceConfigDialog-3.0-79.lua"]:843: in function <...nfig-3.0\AceConfigDialog-3.0\AceConfigDialog-3.0.lua:664>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:72: in function <...ns\CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0.lua:70>
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:306: in function `Fire'
[string "@CorruptionTooltips\Libs\AceGUI-3.0-41\widgets\AceGUIWidget-DropDown.lua"]:442: in function <...ps\Libs\AceGUI-3.0\widgets\AceGUIWidget-DropDown.lua:433>
[string "=[C]"]: ?
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:72: in function <...ns\CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0.lua:70>
[string "@CorruptionTooltips\Libs\AceGUI-3.0\AceGUI-3.0-41.lua"]:306: in function `Fire'
[string "@AzeriteTooltip\libs\AceGUI-3.0-41\widgets\AceGUIWidget-DropDown-Items.lua"]:351: in function <...s\AceGUI-3.0\widgets\AceGUIWidget-DropDown-Items.lua:341>
```


Note:
- I expected a negative value to show a tick relative to the total value
example: -10000 would show a tick at %total - 10000
- In current state of this subregion, i think the "relative" mode could be removed and adding a "%" to the input value would auto-detect it
- I'd expect (without reading tooltip) from a mode called "relative" to show a tick moving relative to %p

InfusOnWoW:
I fixed the lua errors now.

As to the placement, what's the use case for relative to total? 

I'm not too happy about the names either, so suggestions about them are very welcome.

mrbuds:
> As to the placement, what's the use case for relative to total?

An example i can think of is a mana bar with a marker for how much a mana potion restore so you know you should use it when your mana is under that mark

InfusOnWoW:
So, any suggestion how we should name that?

mrbuds:
What do you think of
- Value from start (= current Absolute)
- Value from end
- Percent of total (= current Relative)
?

InfusOnWoW:
I think start is a bit ambigious with how it interacts with timed progress.

E.g. 2.5 for a buff places it at 2.5s remaining time, because at that's 2.5.

I also think that's a false symetry, maybe it's better to have a shorter name for the more natural "value from start", e.g.

* "At Value"
* "At missing Value"
* "Percent of Progress"
* "Percent of Total"

InfusOnWoW:
Also what's the use case for "Percent of Progress" ?

mrbuds:
> 
> 
> Also what's the use case for "Percent of Progress" ?

I don't see one for percent of progress, but i think having a tick mark at progress +/- a static value be useful, like overlays

mrbuds:
Also, i like the names you found better than my suggestions

nullKomplex:
I can't think of a use for "Percent of Progress". The rest sound good. The two things I had in mind but I didn't know how to implement in a good way were:
- Offset from progress (as mentioned by Buds)
- Some form of dynamic, likely custom code fed progress. Could be a custom box or somehow trigger fed (not sure how that would work when multiple ticks could be used).

For the former I was thinking of a progress offset checkbox that enabled a number field. For the latter, there's not really a precedent for either so I didn't know how to proceed.

InfusOnWoW:
So it's 
* "Value Offset from progress"
* "Percent Offset from progress"
?

nullKomplex:
A little long comparatively, but yeah.

InfusOnWoW:
Okay, I'm currently working on something else, but I think I can try to implement this this weekend.

PR #1964, opened 2020-02-05
Cache deflated strings

lqnrd:
# Description

Put serialized tables and their compressed strings in a dictionary for a few minutes. This way we only lag once when the entire raid starts clicking on our chat-linked aura.

Bonus: This also helps to prevent a possible DoS when someone is spam-requesting a single aura hundreds of times.

## Type of change

# How Has This Been Tested?

Tested with version 2.16.2, requesting/sending multiple large auras multiple times.

# Checklist:

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -624,9 +624,28 @@ hooksecurefunc("SetItemRef", function(link, text)
   end
 end);
 
+local compressedTablesCache = {}
+
 function TableToString(inTable, forChat)
   local serialized = Serializer:Serialize(inTable)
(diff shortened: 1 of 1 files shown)


emptyrivers:
I like it, but I did want to ask why 5 minutes? Is that just a ballpark estimate of when the cache is no longer useful, or did you have a specific reason?

lqnrd:
I think at one point in time the lines in chat started to vanish after 5 minutes, which would be the main use case for this patch, and it seemed reasonable. Other than that, I did not think too much about it.

Stanzilla:
I like it! Thank you!

PR #1978, opened 2020-02-09
add 'reset entry' button to arrays in user mode

### LINKED ISSUE #1928: Custom options's "option group" default data
When updating an aura, default behavior for custom option data is to keep user data.
If an aura have a new custom option of the type "option group", for example a spell list, after an update with custom option update disable, that list will be empty (if i understand correctly), because there is no "default" for this field.

I don't know exactly what's the best way to solve this.
Maybe auto-import data for new option group fields ?

Something in author mode to mark current data as default, so when user push the reset button it doesn't empty the list filled by author.

**Additional context**
For <https://wago.io/BFADungeonTargetedSpells> I had a list of spellids, and options for each spells set in a custom function on init.
Updates on custom options from last release let me move this list to custom option UI, but i think/fear most user won't update custom option data when updating aura and will get an empty list of spells.

emptyrivers:
# Description

Fixes #1928

## Type of change

# How Has This Been Tested?

i did the thing and it worked


emptyrivers:
Thanks to @sragia as always for finding artwork for me.

PR #2019, opened 2020-02-26
Disable destName in inverse cast trigger

amakinen:
# Description

The unit trigger rework pushed to alpha a couple of days ago made one of my auras to throw errors when it loaded. Cast trigger's `destName` initialization was calling `UnitName` with a nil value because `destUnit` was disabled when the trigger is inverted. Fix by adding the same enable check `destUnit` has.

Alternatively init could be `UnitName(unit .. '-target')`, but I don't think target's name will be of much interest in a not casting trigger.

## Type of change

# How Has This Been Tested?

- [x] Tested in-game that an inverse casting trigger no longer throws an error on load and works as expected

# Checklist:

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -5846,6 +5846,7 @@ WeakAuras.event_prototypes = {
         display = L["Name of Caster's Target"],
         store = true,
         hidden = true,
+        enable = function(trigger) return not trigger.use_inverse end,
         test = "true",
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
Thanks for the PR, you missed that sourceName should also be disabled and thus I've edited your change and uploaded it in another PR. https://github.com/WeakAuras/WeakAuras2/pull/2022/commits/1fe8f8b8bebebb23341732469dee58e08c6a7c5c

PR #2026, opened 2020-02-29
Allow separators to force a "page break" in custom options

emptyrivers:
I don't know if this is actually worth doing, but it was pretty easy to implement.

Basically, allow separators to force a break in merging, as well as visually. So if you have two complex sections in separate auras, not in a group, you can add a breaking separator to the top of each set of options, and WeakAuras will not attempt to merge them and possibly create a huge mess.

To see the difference:
https://wago.io/x4D57HonC (without separators)
https://wago.io/SZeDvSABI (with separators)

### DIFF
--- a/WeakAurasOptions/AuthorOptions.lua
+++ b/WeakAurasOptions/AuthorOptions.lua
@@ -1092,6 +1092,15 @@ typeControlAdders = {
         return not option.useName
       end
     }
+    args[prefix .. "noMerge"] = {
+      type = "toggle",
+      name = name(option, "noMerge", L["Prevent Merging"]),
+      desc = desc(option, "noMerge", L["If checked, then this separator will not merge with other separators when selecting multiple auras."]),
+      order = order(),
+      width = WeakAuras.doubleWidth,
+      get = get(option, "noMerge"),
+      set = set(data, option, "noMerge"),
+    }
   end,
   group = function(options, args, data, order, prefix, i)
     local option = options[i]
@@ -2275,23 +2284,29 @@ local function mergeOptions(mergedOptions, data, options, config, prepath, paren
     -- find the best place to start inserting the next option to merge
     local nextToMerge = options[i]
     local shouldMerge = false
-    for j = nextInsert, #mergedOptions + 1 do
-      local mergedOption = mergedOptions[j]
-      if not mergedOption then
-        break
-      end -- no more options to check, so must insert
-      local validMerge = true
-      for field in pairs(significantFieldsForMerge) do
-        if nextToMerge[field] ~= mergedOption[field] then
-          validMerge = false
+    if not nextToMerge.noMerge then
+      for j = nextInsert, #mergedOptions + 1 do
+        local mergedOption = mergedOptions[j]
+        if not mergedOption then
+          break
+        end -- no more options to check, so must insert
+        local validMerge = not mergedOption.noMerge
+        if validMerge then
+          for field in pairs(significantFieldsForMerge) do
+            if nextToMerge[field] ~= mergedOption[field] then
+              validMerge = false
+              break
+            end
+          end
+        end
+        if validMerge then
+          shouldMerge = true
+          nextInsert = j
           break
         end
       end
-      if validMerge then
-        shouldMerge = true
-        nextInsert = j
-        break
-      end
+    else
+      nextInsert = #mergedOptions + 1
     end
     -- now we know at what point to add nextToMerge
     if shouldMerge then
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -2017,6 +2017,7 @@ WeakAuras.author_option_fields = {
   header = {
     useName = false,
     text = "",
+    noMerge = false
   },
   group = {
     groupType = "simple",


WeakAurasOptions/AuthorOptions.lua
```diff
@@ -2275,23 +2284,27 @@ local function mergeOptions(mergedOptions, data, options, config, prepath, paren
...
-      for field in pairs(significantFieldsForMerge) do
-        if nextToMerge[field] ~= mergedOption[field] then
-          validMerge = false
+    if not nextToMerge.noMerge then
+      for j = nextInsert, #mergedOptions + 1 do
+        local mergedOption = mergedOptions[j]
+        if not mergedOption then
+          break
+        end -- no more options to check, so must insert
+        local validMerge = mergedOption.noMerge
```
InfusOnWoW: shouldn't this be "         local validMerge = *not* mergedOption.noMerge "

InfusOnWoW:
Apart from what I assume a missing "not" is, I'm wondering whether the "Prevent Merging" option on separators is enough.

Consider this aura (and that's with a added not):
!LFvBtQTnq4FnDM2p0oaHCnFfUZ0JECgkY0K8fdcBzSYzSiwYKd(a)27URTXIxUd6KmPTtNHXJ06vRE8(YZU4303ZNP9z3jw7Zw6ZcXFGO58GNcZuRUvLOYabUnPFTONVHE223TXV8wFdOSkluK1T8i(SUjYTB5zH7M5Pujgji65HrrAHbosL6(ZRw5eUq4ZyFoNNj2nRxEsYUzVpwAaH80Gyv2iLmfUMBDC9CgxDkMCRabu5wBG2GaAdcOniG2eHPrfSwKPLQuF3314nnaOOd4jc81mEGbEHMChgEMbwahrMklxfbl1X0A0uzYflaBvCHE7fGldfZZJI82ScW)9odg1BYa490wEEghqSELijPFOUWY685I1IuddoL8zF20B7W8MY86mgmwoca2Qe(gW47vDuMGuLnYzayCsgCzcEIj2NLYxk0vinpTgAMcrBojwujOTpltSa8dfOFrMkhIDArsuziO7qpVHpoWPhanjKR48mF5QekqjxYrpOTdS8RovLcpdHVDuHPfc1Iavk4caJUKltVET3hiUo9XtKboeicXn8IqnLu1dLw8DYUDSJJBLZOFk5lafbBejxuLjyezP8K)SkdQ1nu8ym5VkdL8Cdy5HRm1IqCiT2NO4ubwUwmniHR1yDaKqeGcxMdvlvrUI3ESunM2FKqmmJrJ7h)7F4Dp8tD72yezG9zNhJRcPaeCbFacGcVwSGJfMmJ4z44OR9rrguB21N9fziKAr1rqT1ENVrTybg(btDNqhGM6jXMYhyEOpdUvSMiIdagFVLLokx8Sf7badsO6lPW5wriPsaXAa2tcEyu()qLK1fBhxgE(I1ZussCpRfEfcFufcms)CZgVyw3tcXQoykJzmMZtjq1LOvoUqHMBWQcYlDqMAP)bIYt1XCWz(bROqzbu(8ISasRIL)GUCBWR2uOGRf1lsHaQxMC7Uz)ropeRaHEchE3xdXDT2F0gPK46V8ot8gADZfH7Ht8g031Pu(APwopHYJO9hKY1311z80ckUX9)T79Qsjp5s)uU2iJ2u7S3FPfDLA2Q6O2E2mLbIgEKV0DiGjk4hlKlIbfUPDn5WxpFGnnovH8cn4SARv3SdwVvPwsGUQCdH9Dsnh8EHycfqPOi34PnvrQOoqM3Uzeh6fZv4jRIVaZS94devNZeV(F8PhgBcEgbocbHf7c6m3ZEtSrZphz8PDgoxxSxO3ID)hRUyxN21DXUk9n75DT9e)NN88BW8mFdip)38Sl)FLf)eUYxNx)7kp(rDs((YIFPj5SMzRAmUdhU7ctYvp(M1mDwJVznu3HtY937pq9v1wP1R0POUzYbTcUyQ(z6wy1w5mZxrTHE4hV5ZF6xvTN0i(0Ul77HEuBgYbSgw1UfUX)Vo

That's a group with:
* Aura 1:
   * Sep No Merge
   * Option

* Aura 2:
   * Sep No Merge
   * Option (same key etc. as Aura 1)

The merged options, doesn't merge the Separators, but that just leads to:

* Group
  * Sep No Merge (Aura 1)
  * Sep No Merge (Aura 2)
  * Option (Aura 1 + Aura 2)

Even adding a additional separator add the end does not work.

Maybe, the right solution is to always append separators with noMerge set?

So:
if nextToMerge.noMerge then
  nextInsert = #mergedOptions + 1
else
  -- look for the insert position


Then the best pratice for aura authors would be:
* Common Options that should be merged first
* Aura specific Options afterwards.

The reverse would obviously not work though.

Or maybe .noMerge should affect not just the separator, but every option until the next separator? The reasoning would be, that separators are a kind of "title" to the following options.

emptyrivers:
Ah, I had intended to have breaking separators always insert at the end. Apparently I forgot to do that.

WeakAurasOptions/AuthorOptions.lua
emptyrivers: You're right, fixed.

emptyrivers:
That is fixed now.

PR #2032, opened 2020-03-05
Supercharge glow actions / conditions

### LINKED ISSUE #1973: Supercharge Glow Action
The current Show Glow actions (OnShow+Conditions) only supports the old Blizzard Glow, we should add the new glows to that.

Additionally we currently only allow specifying a static frame. This could be extended to be as flexible as the Attach to Frame feature, in allowing glowing a unit frame/nameplate or a custom frame.

This would enable creating glows on e.g. group frames, which are quite popular.

mrbuds:
# Description
Action show/hide:
- Add Unit Frame option
- Add Nameplate option
- Add more custom glow options
- Remove all glows on hide
- Conditions

TODO:
- ~~run on hide when opening options~~

Fixes #1973

## Type of change

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -4596,6 +4596,51 @@ function WeakAuras.HandleChatAction(message_type, message, message_dest, message
   end
 end
 
+local function actionGlowStop(actions, frame, id)
+  if actions.glow_type == "buttonOverlay" then
+    LCG.ButtonGlow_Stop(frame)
+  elseif actions.glow_type == "Pixel" then
+    LCG.PixelGlow_Stop(frame, id)
+  elseif actions.glow_type == "ACShine" then
+    LCG.AutoCastGlow_Stop(frame, id)
+  end
+end
+
+local function actionGlowStart(actions, frame, id)
+  if frame:GetWidth() < 1 or frame:GetHeight() < 1 then
+    actionGlowStop(actions, frame)
+    return
+  end
+  local color = actions.use_glow_color and actions.glow_color or nil
+  if actions.glow_type == "buttonOverlay" then
+    LCG.ButtonGlow_Start(frame, color)
+  elseif actions.glow_type == "Pixel" then
+    LCG.PixelGlow_Start(
+      frame,
+      color,
+      actions.glow_lines,
+      actions.glow_frequency,
+      actions.glow_length,
+      actions.glow_thickness,
+      actions.glow_XOffset,
+      actions.glow_YOffset,
+      actions.glow_border,
+      id
+    )
+  elseif actions.glow_type == "ACShine" then
+    LCG.AutoCastGlow_Start(
+      frame,
+      color,
+      actions.glow_lines,
+      actions.glow_frequency,
+      actions.glow_scale,
+      actions.glow_XOffset,
+      actions.glow_YOffset,
+      id
+    )
+  end
+end
+
 function WeakAuras.PerformActions(data, type, region)
   if (paused or WeakAuras.IsOptionsOpen()) then
     return;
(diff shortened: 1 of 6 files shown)


InfusOnWoW:
I haven't looked into it, but you don't need "Auto un-glow on OPTIONS". 

I think the way to solve that involves two actions:
* Fix OnHide to fire on opening the OPTIONS, which I'll look at pretty soon.
* A special OnHide action.
  So you are probably thinking of OnShow and OnHide as being symetrical and offering the same potential actions. And that's mostly fine, but that doesn't need to be.

I think we can add a: "Hide Glows applied by this aura" action, which is just a toggle and if enabled you don't need to specifiy the frame from which to hide the glow.
(That requires keeping track of which frames the aura applied glows too, but that's not hard.)

mrbuds:
> I think we can add a: "Hide Glows applied by this aura" action
Or what about not having this option and make it default, or is it too much magic ?

mrbuds:
About glowing on nameplates my issue so far is to get a generic way to catch best anchor points for health/mana bars compatible with most/all nameplate addons.

C_NamePlate.GetNamePlateForUnit is good enough for anchoring stuff to nameplates, x/y offsets can fix not perfect positioning. This is not good enough for glowing them, frame doesn't mach statusbars in most case.
I could make it work pretty good with a list of hardcoded path found in Exality's explosive orb aura <https://wago.io/explosiveorbs> but that solution was rejected for unitframes, and i have not found a proper generic way to get them.

asaka-wa:
You can anchor to separate objects within the "family" right?
So maaaaybe... make a measurement frame. Anchor bottomright to plate centre. Anchor topleft to each statusbar child. Get the size of the measurement frame. 
Reanchor its topleft to plate centre and measure to status bars' bottomright. GetSize. 
Add 'em up. Anchor glow based on offsets from plate centre    
...or leave your own frame on each plate frame once you've anchored it to the furthest in each direction and use it as your glow frame.

InfusOnWoW:
> 
> 
> > I think we can add a: "Hide Glows applied by this aura" action
> 
> Or what about not having this option and make it default, or is it too much magic ?

That would be backwards incompatible, and I think there are use cases where that is desirable. 

Well, for the anchoring to the personal nameplate, we do have a bit of addon specific code. I think there are drastically less nameplate addons than unit frames addons. 

Or what if LibGetFrame would provide a function?

mrbuds:
Condition is working nice with power trigger, but i'm running into an issue with BuffTrigger2's conditions where region.state.unit is nil on buff removed

trigger: aura > unit = player, type = buff, show = always
condition trigger1 > aura found = false

InfusOnWoW:
The first example, I'll fix. The second can't work because conditions are not run if the aura is hidden.

WeakAurasOptions/ActionOptions.lua
```diff
@@ -425,63 +585,226 @@ function WeakAuras.AddActionOption(id, data)
...
+        width = WeakAuras.normalWidth,
+        desc = function()
+          return (
+            data.actions.finish.glow_frame_type == "UNITFRAME"
+            or data.actions.finish.glow_frame_type == "NAMEPLATE"
+          )
+          and L["Require unit from trigger"] or nil
+        end,
+        name = L["Glow Frame Type"],
+        order = 10.3,
```
InfusOnWoW: This order is wrong, it should be 30.3

WeakAurasOptions/ActionOptions.lua
```diff
@@ -229,19 +229,49 @@ function WeakAuras.AddActionOption(id, data)
...
+        },
+        hidden = function() return not data.actions.start.do_glow end
+      },
+      start_glow_type = {
+        type = "select",
+        width = WeakAuras.normalWidth,
+        name = L["Glow Type"],
+        order = 10.4,
+        values = WeakAuras.glow_types,
+        hidden = function() return not data.actions.start.do_glow end,
```
InfusOnWoW: The glow type should not be visible if the glow action is "Hide".

Same for the OnHide action.

mrbuds:
I have added a new FRAME_GUID_UPDATE callback in LibGetFrame
https://github.com/mrbuds/LibGetFrame/commit/312231f5a7936a062096d05bbdf8582874ec7cfa

This commit https://github.com/WeakAuras/WeakAuras2/pull/2032/commits/2b7f1a3ec1169b2640812e6d29dc733089c766e8 use it to update UNITFRAME glows

InfusOnWoW:
I don't like that each aura has its own callback, that feels to me like it wouldn't scale too well.

I think we should have a registry on our side that keeps track of the units/frames that we might need to reapply and then have one callback.

Also I think you

WeakAuras/WeakAuras.lua
```diff
@@ -4694,44 +4730,38 @@ function WeakAuras.HandleGlowAction(actions, region)
...
+            glow_frame_monitor[glow_guid] = glow_frame_monitor[glow_guid] or {}
+            glow_frame_monitor[glow_guid][id] = {
+              region = region,
+              actions = actions,
+              frame = glow_frame
+            }
+            region.active_glows_hidefunc[glow_frame] = function()
+              actionGlowStop(actions, glow_frame, id)
+              glow_frame_monitor[glow_guid][id] = nil
+              -- TODO: purge glow_frame_monitor[glow_guid] if it's empty
```
InfusOnWoW: The way to check if a table is empty is via next(glow_frame_monitor[glow_guid]) == nil

InfusOnWoW:
I haven't tested your config changes, but imho this looks pretty good now

PR #2042, opened 2020-03-13
Combat Log trigger: fix srcUnit and dstUnit conditions

### LINKED ISSUE #2041: Condition "Destination Unit" doesnt work on classic
When using the condition "Destination Unit" = player or != player, the condition is never true.

image

WA v2.16.6
wow Classic

mrbuds:
# Description
Fixes #2041

## Type of change


mrbuds:
Note that example used in #2041 use condition to whisper to %destUnit which is not replaced by data from trigger.

InfusOnWoW:
There was a small typo in your first commit, which I fixed.

PR #2045, opened 2020-03-16
Smallpatches

InfusOnWoW:


WeakAuras/WeakAuras.lua
```diff
@@ -4610,7 +4609,7 @@ function WeakAuras.PerformActions(data, type, region)
     return;
   end
 
-  if(actions.do_message and actions.message_type and actions.message and not squelch_actions) then
+  if(actions.do_message and actions.message_type and actions.message) then
```
emptyrivers: This squelch_actions was originally to prevent a huge burst of sounds if a user logged in a the wrong time, if i remember correctly. Is that now possible?

mrbuds: thanks for giving context i didn't know what it was

InfusOnWoW: Yes, I've kept it around for sound but removed it for chat and actions.

emptyrivers: Great, thanks.

PR #2058, opened 2020-03-22
Smallpatches

InfusOnWoW:


mrbuds:
It seems there is a bug with additionals fields for 
- names
- exact spell ids
- blacklisted exact spell ids

But work correctly for Blaslisted names
img

InfusOnWoW:
@mrbuds I can't reproduce that, do you have the aura that showed the problem?

mrbuds:
I can't manage to reproduce it anymore

PR #2089, opened 2020-04-05
Smallpatches

InfusOnWoW:


WeakAuras/GenericTrigger.lua
```diff
@@ -2181,6 +2181,11 @@ do
       startTimeCooldown = 0;
     end
 
+    if (startTimeCooldown > (2^31 + 2^30) / 1000) then
```
emptyrivers, line 2184: I assume this is just an arbitrary cutoff?

InfusOnWoW: Yes.

PR #2093, opened 2020-04-08
Glow external element fixes

mrbuds:
# Description
> improve flow of glow external element conditions & actions

show options as you select them

> remove previous glow before applying a new one 

fix an issue when with conditions you change glow type it wouldn't remove previous glow and overwrite the unglow function

### DIFF
--- a/WeakAurasOptions/ConditionOptions.lua
+++ b/WeakAurasOptions/ConditionOptions.lua
@@ -867,17 +867,32 @@ local function addControlsForChange(args, order, data, conditionVariable, condit
     local function anyGlowExternal(property, needle)
       local ref = type(conditions[i].changes[j].value) == "table" and conditions[i].changes[j].value[property]
       if ref then
-        return ref == needle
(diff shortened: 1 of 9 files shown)


WeakAurasOptions/ConditionOptions.lua
```diff
@@ -876,7 +876,7 @@ local function addControlsForChange(args, order, data, conditionVariable, condit
           end
         end
       end
-      return false
+      return needle == nil
```
InfusOnWoW, line 879: This is pretty strange, what are you trying to do here?

mrbuds: this is to make anyGlowExternal("glow_frame_type", nil) return true if glow_frame_type is nil

if glow_frame_type is nil then
`local ref = type(conditions[i].changes[j].value) == "table" and conditions[i].changes[j].value[property]`
will be nil and the function will always return false without this change

PR #2097, opened 2020-04-11
Smallpatches

InfusOnWoW:


InfusOnWoW:
Don't immediately merge, I need to check if the china fix works.

PR #2121, opened 2020-04-17
Add missing migration for action's glow_type

### LINKED ISSUE #2120: No default aura for glow since 2.17
Since 2.17 all my wa don't glow anymore and i have to downgrade to 2.16 to make them glow again.

Weak Aura discord helped me and the issue is that in 2.17 there is no default aura for glow so you have to choose one now. Old weak aura may have the default so glow doesn't work on 2.17 because of that

mrbuds:
fixes #2120 

This add a migration.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -1,4 +1,4 @@
-local internalVersion = 28;
+local internalVersion = 29;
 
 -- WoW APIs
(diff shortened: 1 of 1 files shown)


emptyrivers:
I did guess that this was a migration problem, but it might also not be a common issue. Perhaps it might be safer in the long run to make the code which consumes this value nil-tolerant?

mrbuds:
Ok i'v changed it to use "buttonOverlay" if glow_type is not set

PR #2138, opened 2020-04-21
Add Gradient to Progress Bar Aura Type

asaka-wa:
# Description

This was mostly a project to help me get git set up. Plus I was wondering about how practical it might be to add something like this. So, if you don't think it's worthy of adding to the Addon then that's fine, but I would appreciate any thoughts on how I did it in any case. 

So, it adds an option to show Gradients on Progress Bars. Options for Colour and direction. Setters for Conditions. Works reasonably well with Animations. 

https://i.imgur.com/gHNet7J.gif (gif shows normal gradient first, with animation second)

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -296,7 +296,7 @@ local barPrototype = {
       self.fgFrame:SetHeight(yProgress > 0.0001 and yProgress or 0.0001);
     end
 
-    -- Apply updated gradient 
+    -- Apply updated gradient
     if self.updateGradient then
       self:GetParent():Color(self:GetParent():GetColor())
     end
@@ -940,7 +940,7 @@ local function modify(parent, region, data)
   region.bar.gradient = data.gradient
   region.bar.gradientDirection = data.gradientDirection
   region.bar.gradientColor = data.gradientColor
-  region.bar.updateGradient = data.gradient and data.orientation:match(data.gradientDirection) 
+  region.bar.updateGradient = data.gradient and data.orientation:match(data.gradientDirection)
 
   -- Color update function
   region.Color = region.Color or function(self, r, g, b, a)
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -15,6 +15,9 @@ local default = {
   inverse = false,
   barColor = {1.0, 0.0, 0.0, 1.0},
   backgroundColor = {0.0, 0.0, 0.0, 0.5},
+  gradient = false,
+  gradientColor = {1,1,1,1},
+  gradientDirection = "HORIZONTAL",
   spark = false,
   sparkWidth = 10,
   sparkHeight = 30,
@@ -117,6 +120,22 @@ local properties = {
     display = L["Inverse"],
     setter = "SetInverse",
     type = "bool"
+  },
+  gradientColor = {
+    display = L["Gradient Color"],
+    setter = "SetGradientColor",
+    type = "color"
+  },
+  gradientDirection = {
+    display = L["Gradient Direction"],
+    setter = "SetGradientDirection",
+    type = "list",
+    values = WeakAuras.gradient_direction_types
+  },
+  gradient = {
+    display = L["Gradient Enable"],
+    setter = "SetGradient",
+    type = "bool",
   }
 };
 
@@ -275,6 +294,11 @@ local barPrototype = {
       local yProgress = select(2, self:GetRealSize()) * progress;
       self.fg:SetHeight(yProgress > 0.0001 and yProgress or 0.0001);
       self.fgFrame:SetHeight(yProgress > 0.0001 and yProgress or 0.0001);
+    end
+
+    -- Apply updated gradient 
+    if self.updateGradient then
+      self:GetParent():Color(self:GetParent():GetColor())
     end
 
     -- Stretch texture
@@ -508,7 +532,26 @@ local barPrototype = {
 
   -- Set bar color
   ["SetForegroundColor"] = function(self, r, g, b, a)
-    self.fg:SetVertexColor(r, g, b, a);
+    if self.gradient then
+      local r2, g2, b2 = unpack(self.gradientColor)
+      if self.updateGradient then
+        local value = self.value;
+        value = math.max(self.min, value);
+        value = math.min(self.max, value);
+        local progress = (value - self.min) / (self.max - self.min);
+
+        local r3, g3, b3 = r + (progress * (r2 - r)), g + (progress * (g2 - g)), b + (progress * (b2 - b))
+        if self.directionInverse then
+          self.fg:SetGradient(self.gradientDirection, r3, g3, b3, r, g, b)
+        else
+          self.fg:SetGradient(self.gradientDirection, r, g, b, r3, g3, b3)
+        end
+      else
+        self.fg:SetGradient(self.gradientDirection, r, g, b, r2, g2, b2)
+      end
+    else
+      self.fg:SetVertexColor(r, g, b, a)
+    end
   end,
 
   ["GetForegroundColor"] = function(self)
@@ -893,6 +936,12 @@ local function modify(parent, region, data)
   bar.spark.sparkRotation = data.sparkRotation;
   bar.spark.sparkMirror = data.sparkMirror;
 
+    -- Gradient
+  region.bar.gradient = data.gradient
+  region.bar.gradientDirection = data.gradientDirection
+  region.bar.gradientColor = data.gradientColor
+  region.bar.updateGradient = data.gradient and data.orientation:match(data.gradientDirection) 
+
   -- Color update function
   region.Color = region.Color or function(self, r, g, b, a)
     self.color_r = r;
@@ -1219,6 +1268,22 @@ local function modify(parent, region, data)
   function region:SetOverlayColor(id, r, g, b, a)
     region.bar:SetAdditionalBarColor(id, { r, g, b, a});
   end
+
+  function region:SetGradientColor(r,g,b)
+    self.bar.gradientColor = {r,g,b}
+    self:Color(self:GetColor())
+  end
+
+  function region:SetGradientDirection(direction)
+    self.bar.gradientDirection = direction
+    self:Color(self:GetColor())
+  end
+
+  function region:SetGradient(bool)
+    self.bar.gradient = bool
+    self:Color(self:GetColor())
+  end
+
   -- Update internal bar alignment
   region.bar:Update();
 
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -15,9 +15,12 @@ local default = {
   inverse = false,
   barColor = {1.0, 0.0, 0.0, 1.0},
   backgroundColor = {0.0, 0.0, 0.0, 0.5},
-  gradient = false,
-  gradientColor = {1,1,1,1},
-  gradientDirection = "HORIZONTAL",
+  gradientFG = false,
+  gradientColorFG = {1,1,1,1},
+  gradientDirectionFG = "HORIZONTAL",
+  gradientBG = false,
+  gradientColorBG = {0,0,0,1},
+  gradientDirectionBG = "HORIZONTAL",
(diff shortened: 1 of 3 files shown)


emptyrivers:
cute

emptyrivers:
Though, I am concerned that this is too specific a feature? Like, how many people actually wanted this?

asaka-wa:
Yup. I think people often see something that looks different from what you normally see in WeakAuras, and like it strongly on that level, but not sure if anyone has a hole in their UI that's just this shape.

asaka-wa:
Something prettied up a bit, rather than the stark example above 
https://i.imgur.com/XZ2zaMR.gif

emptyrivers:
@Stanzilla @InfusOnWoW @mrbuds @emptyrivers @nullKomplex @otherpeople thoughts?

InfusOnWoW:
No thoughts at the moment. I'm not sure whether it has the right benefit/complexity ratio, or whether it would be best to put it into a subelement.

asaka-wa:
If it were thought to be worthy of inclusion I'd go about adding it to other Aura Types for which it makes sense. 
...Though I imagine, without having actually looked yet, that Progress Textures would be a difficult/impossible.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -893,6 +936,12 @@ local function modify(parent, region, data)
   bar.spark.sparkRotation = data.sparkRotation;
   bar.spark.sparkMirror = data.sparkMirror;
 
+    -- Gradient
+  region.bar.gradient = data.gradient
+  region.bar.gradientDirection = data.gradientDirection
+  region.bar.gradientColor = data.gradientColor
+  region.bar.updateGradient = data.gradient and data.orientation:match(data.gradientDirection)
```
nullKomplex: This doesn't catch orientation changes via Conditions.

https://i.imgur.com/TbkNYho.gifv

Nightwarden24:
How is this different from using a color animation (including custom code)? By the way, That's a same formula: [link1](https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/Prototypes.lua#L387) and [link2](https://github.com/WeakAuras/WeakAuras2/pull/2138/commits/15ebece8ba50ce441b2bc14e04659f5941e29410#diff-c60dc10284dce7d64cb062f4e40038e5R543)

asaka-wa:
Animations just change the one colour on the Aura. The PR adds a gradient between 2 given colours on the texture.

WeakAuras/RegionTypes/AuraBar.lua
asaka-wa: Good catch! Thanks, I'll look into this.

mrbuds:
I like it!

Things i see that could be improved:
- disable color widget when checkbox is unchecked (like it's done with glow subregion's color)
- add choice between different algorithm of gradients
- add gradient background
- note sure about this one, group the gradients conditions with "Bar Color" using the 2 columns widget we use for "glow"

asaka-wa:
@mrbuds "different algorithm", you mean HSV as well as RGB?
I was playing with that initially before I started a PR <https://discordapp.com/channels/172440238717665280/172440238717665280/702111929501745232>
As you can see, though it's a nicer colour change, it's not what SetGradient will use on the texture so looks odd when mixed.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -277,6 +315,11 @@ local barPrototype = {
       self.fgFrame:SetHeight(yProgress > 0.0001 and yProgress or 0.0001);
     end
 
+    -- Apply updated gradient
+    if self.updateGradient then
+      self:GetParent():Color(self:GetParent():GetColor())
```
InfusOnWoW, line 320: The bar should not know about the parent. 

Instead I think you need to introduce a "UpdateForegroundColor()" in the bar prototype table, which uses the foreground + foreground gradient + gradient settings to apply the gradient.

And call that function whenever any of those properties change. 

That's a better way to manage this instead of calling the aurabar's Color function.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -508,20 +551,43 @@ local barPrototype = {
...
   ["SetBackgroundColor"] = function(self, r, g, b, a)
-    self.bg:SetVertexColor(r, g, b, a);
+    if self.gradientBG then
+      self.bg:SetGradient(self.gradientDirectionBG, r, g, b, unpack(self.gradientColorBG))
+    else
+      self.bg:SetVertexColor(r, g, b, a)
+    end
   end,
 
   ["GetBackgroundColor"] = function(self)
```
InfusOnWoW, line 589: I think we can just remove the function.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -508,20 +551,43 @@ local barPrototype = {
...
+        end
+      else
+        self.fg:SetGradient(self.gradientDirectionFG, r, g, b, r2, g2, b2)
+      end
+    else
+      self.fg:SetVertexColor(r, g, b, a)
+    end
   end,
 
   ["GetForegroundColor"] = function(self)
```
InfusOnWoW, line 576: This function can be just removed.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -508,20 +551,43 @@ local barPrototype = {
 
   -- Set bar color
   ["SetForegroundColor"] = function(self, r, g, b, a)
-    self.fg:SetVertexColor(r, g, b, a);
+    if self.gradientFG then
```
InfusOnWoW, line 554: And here, you'd only store the foreground colors, and then call UpdateForegroundColor()

(And similar for background color)

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1207,6 +1284,7 @@ local function modify(parent, region, data)
 
   function region:SetOrientation(orientation)
     orient(region, data, orientation);
+    region.bar.updateGradient = region.bar.gradientFG and region.bar.orientation:match(region.bar.gradientDirectionFG)
```
InfusOnWoW, line 1287: And here would be a call to region.bar:UpdateForegroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
   function region:SetOverlayColor(id, r, g, b, a)
     region.bar:SetAdditionalBarColor(id, { r, g, b, a});
   end
+
+  function region:SetGradientColorFG(r,g,b)
+    self.bar.gradientColorFG = {r,g,b}
+    self:Color(self:GetColor())
```
InfusOnWoW, line 1303: And here would be a call to region.bar:UpdateForegroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
...
+  end
+
+  function region:SetGradientDirectionFG(direction)
+    self.bar.gradientDirectionFG = direction
+    self:Color(self:GetColor())
+  end
+
+  function region:SetGradientFG(bool)
+    self.bar.gradientFG = bool
+    self:Color(self:GetColor())
```
InfusOnWoW, line 1313: And here would be a call to region.bar:UpdateForegroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
...
   end
+
+  function region:SetGradientColorFG(r,g,b)
+    self.bar.gradientColorFG = {r,g,b}
+    self:Color(self:GetColor())
+  end
+
+  function region:SetGradientDirectionFG(direction)
+    self.bar.gradientDirectionFG = direction
+    self:Color(self:GetColor())
```
InfusOnWoW, line 1308: And here would be a call to region.bar:UpdateForegroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
...
+  end
+
+  function region:SetGradientColorBG(r,g,b)
+    self.bar.gradientColorBG = {r,g,b}
+    region:SetBackgroundColor(region.backgroundColor)
+  end
+
+  function region:SetGradientDirectionBG(direction)
+    self.bar.gradientDirectionBG = direction
```
InfusOnWoW, line 1322: And here would be a call to region.bar:UpdateBackgroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
...
+  end
+
+  function region:SetGradientFG(bool)
+    self.bar.gradientFG = bool
+    self:Color(self:GetColor())
+  end
+
+  function region:SetGradientColorBG(r,g,b)
+    self.bar.gradientColorBG = {r,g,b}
+    region:SetBackgroundColor(region.backgroundColor)
```
InfusOnWoW, line 1318: And here would be a call to region.bar:UpdateBackgroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -1219,6 +1297,35 @@ local function modify(parent, region, data)
...
+  end
+
+  function region:SetGradientDirectionBG(direction)
+    self.bar.gradientDirectionBG = direction
+  end
+
+  function region:SetGradientBG(bool)
+    self.bar.gradientBG = bool
```
InfusOnWoW, line 1326: And here would be a call to region.bar:UpdateForegroundColor()

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -15,6 +15,12 @@ local default = {
   inverse = false,
   barColor = {1.0, 0.0, 0.0, 1.0},
   backgroundColor = {0.0, 0.0, 0.0, 0.5},
+  gradientFG = false,
```
InfusOnWoW, line 18: We don't in general abbreviate.

WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -508,20 +551,43 @@ local barPrototype = {
...
+      local r2, g2, b2 = unpack(self.gradientColorFG)
+      if self.updateGradient then
+        local value = self.value;
+        value = math.max(self.min, value);
+        value = math.min(self.max, value);
+        local progress = (value - self.min) / (self.max - self.min);
+
+        local r3, g3, b3 = r + (progress * (r2 - r)), g + (progress * (g2 - g)), b + (progress * (b2 - b))
+        if self.directionInverse then
+          self.fg:SetGradient(self.gradientDirectionFG, r3, g3, b3, r, g, b)
```
InfusOnWoW, line 564: Shouldn't this be SetGradientAlpha instead of SetGradient?

asaka-wa:
Thanks for the input Infus. I would probably have a couple of follow up questions but most of that makes total sense. 

I must admit though, even I'm not really gonna put up a big fight in favour of this feature so if you're not feeling it then it's probably best left out. As I said in the PR's description, I was interested in getting git properly set up so I would be able to contribute more usefully and this was the first thing I thought of doing, not because there's a lot of people wanting the feature out there. 

It can be used to make some nice things but it doesn't go much beyond aesthetics and everything here could be pushed in via a snippet in an animation rather than in the addon itself.

InfusOnWoW:
It might be, that it is fine, but just has the problem of going against the current "zeitgeist". 

The whole idea behind sub elements is to be able to add more options to the regions, by making optional addons to base auras. And this directly goes against that, as it adds options to the base aura.

This is argueably a much smaller addition than e.g. sparks in terms of options.

So imho, I'm leaning towards adding that.

emptyrivers:
Hey @asaka-wa, there are some merge conflicts to resolve currently.

Also it looks like the repository fork this pull request was opened from no longer exists, so I'm going to go ahead and close it. If you're still interested in submitting this, then please open a new pull request, thanks!

Issue #654, opened 2018-08-18
Point out new features


Stanzilla:
Investigate using the "!" icon Blizzard uses to point out new features in the default UI to point out new stuff in our UI as well

emptyrivers:
`Interface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon`

Stanzilla:
`'|TInterface\\OptionsFrame\\UI-OptionsFrame-NewFeatureIcon:0|t' .. L["Some translation phrase"]`

Stanzilla:
In 2.7.3 I added this to:
* Templates on the "New" screen
* Condition Combinations
* TSU Custom Variables
* War Mode Load Condition
* Group Role Trigger Filter

We'll have to decide on how long to keep those around, I think having them for a fixed period of time is probably better than having them for a fixed number of releases since that is always dependent on how many bug fix releases we do.

Stanzilla:
I suggest just leaving this open to track where we add and remove it.

@mrbuds you added it to a few places in your last commit, please list them here

mrbuds:
Not release yet (2.7.8 ?)

* Templates "Hold CTRL to create multiple auras at once"

Stanzilla:
When did we say we want to remove the indicators again? x releases or time based?

emptyrivers:
I don’t think we ever did specifically. 

I think that 3-4 months is ideal.

Stanzilla:
https://github.com/WeakAuras/WeakAuras2/pull/1087

is removing the indicators on:

* War Mode
* Filter by Group Role
* Condition Combinations
* TSU Custom Variables

I'd say we give Templates another two months or so.

mrbuds:
Date of commits (not release!)
- Author Options:  13/11/2018
- Conditions for glow/cooldown ?
- Action tab => glow 06/10/2018
- Template + hold ctrl in templates: 06/10/2018
- Cooldown swipe & edge on icons: 10/11/2018

emptyrivers:
Bump for features that we can peel the "New!" sticker off of.

emptyrivers:
#2293 puts us on a clean slate up for shadowlands.

InfusOnWoW:
We haven't really kept information in this ticket. So...

PR #2173, opened 2020-05-05
Add more precision types to format large durations neatly 

elvador:
# Description

Currently WeakAuras formats every duration above 60s as mm:ss, as seen here:
https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/Prototypes.lua#L6939-L6942

For larger durations, that's not the the nicest solution imo. 

This pull request adds more precision types, while maintaining old behavior.

Wow_2020-05-06_00-12-20

I also tried to highlight the differences between all options by giving more examples in general. This is what I ended up with. Feedback welcome.

## Type of change

# How Has This Been Tested?
Running tests to ensure the expected output:
https://gist.github.com/elvador/aa842df0b46854fc5df5026ae56ae0e4

According to the WeakAuras Profiling, there is no performance difference between this implementation and the current master branch with my setup.

If this gets merged, I'd recommend an extensive alpha test, as this touches core functionality.

# Checklist:

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -7223,3 +7223,65 @@ end
 function WeakAuras.UntrackableUnit(unit)
   return not trackableUnits[unit]
 end
+
+do
+  local timeFormatter = {}
+  Mixin(timeFormatter, SecondsFormatterMixin)
+  timeFormatter:Init(0, SecondsFormatter.Abbreviation.OneLetter)
+  timeFormatter:SetStripIntervalWhitespace(true)
+
+  --- Formats a positive duration
+  -- @param seconds the seconds to format
+  -- @param[opt] precision precision or id from WeakAuras.precision_types, default 1
+  function WeakAuras.SecondsToTime(seconds, precision)
+    if seconds == math.huge or tostring(seconds) == "nan" then
+      -- This is the original behaviour, no idea why a space
+      return " "
+    end
+    if seconds < 0 then
+      -- If we'd return a space like above, all the fake timers in the options get stuck at 0, without restarting
+      return ""
+    end
+
+    precision = precision or 1
+
+    -- Use Blizzard style above 10 minutes
+    if precision == 6 or precision == 7 then
+      if seconds < 600 then
+        precision = precision == 6 and 0 or 4
+      else
+        local fmt, time = SecondsToTimeAbbrev(seconds)
+
+        -- Remove the space between the value and unit
+        return fmt:gsub(" ", ""):format(time)
+      end
+    end
+
+    -- "Modern" Blizzard style using the mixin
+    if precision == 8 or precision == 9 then
+      if precision == 9 and seconds < 10 then
+        -- Additional precision on style 9
+        precision = 1
+      else
+        return timeFormatter:Format(seconds)
+      end
+    end
+
+    if seconds > 60 then
+      -- Format as mm:ss
+      local remainingStr = string.format("%i:", math.floor(seconds / 60))
+      seconds = seconds % 60
+      return remainingStr .. string.format("%02i", seconds)
+    end
+
+    if precision == 4 or precision == 5 then
+      -- Add precision below 3s
+      if seconds < 3 then
+        return string.format(precision == 5 and "%.2f" or "%.1f", seconds)
+      end
+      return string.format("%d", seconds)
+    end
+
+    return string.format("%.".. precision .."f", seconds)
+  end
+end
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -3223,6 +3223,8 @@ globals = {
 	"SearchLFGJoin",
 	"SearchLFGLeave",
 	"SearchLFGSort",
+	"SecondsFormatter", -- from SecondsFormatterMixin
+	"SecondsFormatterMixin",
 	"SecureCmdOptionParse",
 	"SelectActiveQuest",
 	"SelectAvailableQuest",
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6930,32 +6930,8 @@ WeakAuras.dynamic_texts = {
         if not state.expirationTime or not state.duration then
           return ""
         end
-        local remaining  = state.expirationTime - GetTime();
-        local duration  = state.duration;
 
-        local remainingStr     = "";
-        if remaining == math.huge then
-          remainingStr     = " ";
-        elseif remaining > 60 then
-          remainingStr     = string.format("%i:", math.floor(remaining / 60));
-          remaining        = remaining % 60;
-          remainingStr     = remainingStr..string.format("%02i", remaining);
-        elseif remaining > 0 then
-          local progressPrecision = region.progressPrecision and math.abs(region.progressPrecision) or 1
-          -- remainingStr = remainingStr..string.format("%."..(data.progressPrecision or 1).."f", remaining);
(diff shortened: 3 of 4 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -7223,3 +7223,65 @@ end
...
+  -- @param[opt] precision precision or id from WeakAuras.precision_types, default 1
+  function WeakAuras.SecondsToTime(seconds, precision)
+    if seconds == math.huge or tostring(seconds) == "nan" then
+      -- This is the original behaviour, no idea why a space
+      return " "
+    end
+    if seconds < 0 then
+      -- If we'd return a space like above, all the fake timers in the options get stuck at 0, without restarting
+      return ""
+    end
```
elvador, line 7244: @InfusOnWoW If there is any logic or reasoning behind this, enlighten me please

krazyito65:
I like it, the only problem I have is the way its displayed in the dropdown... Might be a bit difficult for newer users to understand what they are looking at and what the differences are, but I'm not really sure how else to display it.  Took me a good 5m of staring at it to realize what it meant (after realizing the base was mm:ss)

Stanzilla:
I like it. Would change the yellow to the WoW gold though.

And please add a new feature indicator https://github.com/WeakAuras/WeakAuras2/issues/654

elvador:
> I like it, the only problem I have is the way its displayed in the dropdown... 

Yeah, I agree. I also tried giving them arbitrary names and moving the examples in the tooltip, but that looked even worse. Open to suggestions to improve the display. 

If someone wants to play around, just change the strings here:
https://github.com/WeakAuras/WeakAuras2/pull/2173/files#diff-a5afb1bee3b486b406cb28e624592275R71

krazyito65:
I tried to give arbitrary names, and possibly give a tooltip, in my post then i realized it was just as bad haha.. 

I was thinking maybe:
- no decimal
- 1 decimal on seconds
- 2 decimals on seconds
- 3 decimals on seconds
- 1 decimal on single digit
- 2 decimals on single digit
- shortened hours, no decimal
- shortened hours, 1 decimal on single digit
- Shortened hours and minutes, no decimal
- shortened hours and minutes 1 decimal for single digit.


---
I feel its still probably not the best solution though, unless there were tool tips to compliment with examples like you have.

krazyito65:
ehh the more i look at it, i think its still too wordy and the original way is better..

Stanzilla:
> I tried to give arbitrary names, and possibly give a tooltip, in my post then i realized it was just as bad haha..
> 
> I was thinking maybe:
> 
> * no decimal
> * 1 decimal on seconds
> * 2 decimals on seconds
> * 3 decimals on seconds
> * 1 decimal on single digit
> * 2 decimals on single digit
> * shortened hours, no decimal
> * shortened hours, 1 decimal on single digit
> * Shortened hours and minutes, no decimal
> * shortened hours and minutes 1 decimal for single digit.
> 
> I feel its still probably not the best solution though, unless there were tool tips to compliment with examples like you have.

Yeah, no :D

krazyito65:
ok real feedback:  I would change all the 63:42 to 120:11 to match the columns all the way down so people know exactly whats staying the same between each one.

krazyito65:
nvm ignore me :)

elvador:
Also, correct would be
> 1 decimal on single digit at or below 3s

🙂

---


> ok real feedback: I would change all the 63:42 to 120:11 to match the columns all the way down so people know exactly whats staying the same between each one.

They are already showing the same values.
I'm relying on default WoW behavior - which isn't consistent between the different functions.
Type 6 and 7 (counting from 0) are using `SecondsToTimeAbbrev` the threshold.
Type 8 and 9 are using the `SecondsFormatterMixin` above the threshold.

InfusOnWoW:
I'm not really convinced that there should be 10 different formats. Can't we just modify our existing formats to automatically switch to hour formats for > 60 minutes?

elvador:
I've discussed this with @Stanzilla before making this PR. We both agreed that changing default behavior will most likely result in people complaining. Especially considering that you'd have to make a decision on how 63 minutes are displayed: **2h** (blizzard default for buffs, for example) vs **1h 3m** - there are arguments for both. That's why I went for the "adding more options is always better than limiting them" approach.

InfusOnWoW:
Well, you do have only a subset of all potential options, since my favorite is not in there:
"2h, 11m, 3m, 42, 9, 3"

Thus why don't we add that as a option. Because it's plainly wrong that more options are better. We can't add infinite options and we shouldn't.

We need to choose a subset that both a) serves a lot of users, b) doesn't overwhelm them with options. For all the special snowflakes there's %c. The over 1h cooldown problem has come up often enough, that we should add something for that. But adding 4 more options to a already croweded setting, that's adding imho too much complexity for too little gain.

I think we should either, change the existing formats. Or introduce one new format that is Blizzard like.

elvador:
> Well, you do have only a subset of all potential options, since my favorite is not in there:
"2h, 11m, 3m, 42, 9, 3"

I never said I covered all of them - but yeah, of course, this can be added. The reason I didn't add it is that I figured in a game like wow where most cooldowns are within the 10min range, having the seconds below 10min already can make a big difference: Cooldown ready in 2m vs 1m 1s.

> Because it's plainly wrong that more options are better. We can't add infinite options and we shouldn't.

That's why I only added 4 (in my opinion) sensible ones:
6: Like precision 0, but with nicer formatting >10m
7: Like precision 4 ("dynamic precision 1"), but with nicer formatting >10m
8: Nicely formatted throughout
9: Like 8, but with additional prescription below 10s

Please keep in mind that 6 of the options already exit in WeakAuras, whereas at least one is very questionable (3 digit precision).

InfusOnWoW:
I don't want to add it, that is just a example of a reasonable format that is missing. The point is, there's a cost to adding more options to the precission setting.

That cost is clearly visible in your screenshot. How many seconds does a normal user need to parse that kind of mess? Yeah it's functional, and the "dynamic" naming isn't exactly obvious either, but still that UI is ugly and understanding it takes too long.

That cost has to be considered when looking at the benefit. 

There's clearly a benefit, I have seen a few requests for hour longer cooldown formatting and quite a few for a more "blizzard" style formatting.

Since there's a benefit, a solution would be nice. But it can't have the cost you are proposing. 

You can either drastically cut back on the cost side by proposing something simpler, e.g. just adding 8), or you need to convince me that there's some huge benefit to the extra formats, that isn't clear to me.

emptyrivers:
Hey @elvador, you still have some housekeeping to do per this comment: https://github.com/WeakAuras/WeakAuras2/pull/2173#issuecomment-624346021

Additionally, you have some merge conflicts. If you're still interested in submitting this, then please resolve those when you can, thanks!

elvador:
There was no agreement on how it would need to be implemented. I think we all agree that the currently implemented solution needs improving, but apparently my solution wasn't what was needed. I'll gladly update everything as soon as a decision is made on how it should look like.

InfusOnWoW:
The new infrastructure makes it much easier to add as many orthogonal settings as needed. 

From what I see, you proposed essentially three:
* Add a hour format
* Additional precission for <10s (instead of the hardcoded 3s)
* Units for all numbers
* Additional precission for "small" values

Imo:
* We can just format >60m as 1h. I expect pratically no one to even notice.
* I'm fine with making the threshold configrable.
* I think having a this configurable is too niche.
* There's a use case for that, since the difference between 1m 1s and 1m 59 is pretty huge.

What we can't have though, is a trying to solve that matrix of orthogonal options in a flat list.

PR #2228, opened 2020-06-13
Threat Situation trigger rework and enable it for WoW 1.13.5

mrbuds:
…e more informations

If unit is not "At Least One Enemy"
- use UNIT_THREAT_LIST_UPDATE instead of UNIT_THREAT_SITUATION_UPDATE
- use UnitDetailedThreatSituation instead of UnitThreatSituation
- %threatpct, %rawthreatpct and %threatvalue can be use
- duration function use threatpct

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -78,6 +78,13 @@ end
 
 LibRangeCheck:RegisterCallback(LibRangeCheck.CHECKERS_CHANGED, RangeCacheUpdate)
 
+function WeakAuras.UnitDetailedThreatSituation(unit1, unit2)
+  local ok, aggro, status, threatpct, rawthreatpct, threatvalue = pcall(UnitDetailedThreatSituation, unit1, unit2)
+  if ok then
+    return aggro, status, threatpct, rawthreatpct, threatvalue
+  end
+end
+
 local LibClassicCasterino
 if WeakAuras.IsClassic() then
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
I'll take a look soon.

InfusOnWoW:
Well I've been busy, I'm not sure about setting the statesParameter to "unit" yet, that feels a bit strange for a non-cloning trigger.

Stanzilla:
When is 1.13.5 coming out?

InfusOnWoW:
It's already out iirc.

oratory:
It's not out yet, it's coming with AQ and related content. No timeline from Blizz yet but likely 4-5 weeks away still.

mrbuds:
Why did you add the wrapper WeakAuras.UnitDetailedThreatSituation for UnitDetailedThreatSituation ?
I ask because later i'v added
```
  if not UnitDetailedThreatSituation then
    WeakAuras.event_prototypes["Threat Situation"] = nil
  end
```
which can be remove after 1.13.5 release

InfusOnWoW:
Because it's too picky about the passed parameters. Calling UnitDetailedThreatSituation with "player", "any non unit string" results in a error, which makes using specific unit pretty error prone.

Stanzilla:
July 27th is the Raid release

mrbuds:
It doesn't need to wait for next patch to be merged

PR #2235, opened 2020-06-17
Add LibSerialize

emptyrivers:
# Description

We have a new option for serializing arbitrary data, thanks to @rossnichols. From experiments, it looks to be about a 20-25% reduction in the final size of the exported string, which is good enough imo to warrant adoption.

Since this change is not forwards-compatible, it warrants a minor version upgrade to `2.18`, I think. And once it's been released, we should advise people to ensure that they are up to date.

Build artifact will fail to load properly in game, until [this](https://github.com/rossnichols/LibSerialize/pull/1) change is complete.

## Type of change

# How Has This Been Tested

as always, I was lazy and haven't done any thorough testing beyond the initial experiments.

### DIFF
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -20,6 +20,7 @@ externals:
   WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk/LibDBIcon-1.0
   WeakAuras/Libs/LibGetFrame-1.0: https://github.com/mrbuds/LibGetFrame
   WeakAuras/Libs/Archivist: https://github.com/emptyrivers/Archivist
+  WeakAuras/Libs/LibSerialize: https://github.com/emptyrivers/LibSerialize
 
 enable-nolib-creation: no
 
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -20,7 +20,9 @@ externals:
   WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk/LibDBIcon-1.0
   WeakAuras/Libs/LibGetFrame-1.0: https://github.com/mrbuds/LibGetFrame
   WeakAuras/Libs/Archivist: https://github.com/emptyrivers/Archivist
-  WeakAuras/Libs/LibSerialize: https://github.com/rossnichols/LibSerialize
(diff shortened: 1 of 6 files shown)


.pkgmeta
```diff
@@ -20,6 +20,7 @@ externals:
   WeakAuras/Libs/LibDBIcon-1.0: https://repos.wowace.com/wow/libdbicon-1-0/trunk/LibDBIcon-1.0
   WeakAuras/Libs/LibGetFrame-1.0: https://github.com/mrbuds/LibGetFrame
   WeakAuras/Libs/Archivist: https://github.com/emptyrivers/Archivist
+  WeakAuras/Libs/LibSerialize: https://github.com/rossnichols/LibSerialize
```
rossnichols, line 23: As mentioned elsewhere, you should ingest version 1.0.0 so I don't have to worry about breaking WeakAuras :)

Stanzilla:
I missed a lot of the talk about this on Discord but was the possibility of adopting this in Ace3 discussed at all? @Nevcairiel

Nevcairiel:
I don't think there is any real advantage to doing that. Nothing wrong with standalone libraries for specific use-cases.

WeakAuras/Transmission.lua
```diff
@@ -663,11 +665,26 @@ function TableToString(inTable, forChat)
...
+  local encoded = ""
+  if inString:find("^%!") then
+    local header
+    inString:gsub("^%!WA%:%d+!", function(m) header = m return "" end)
+    if not header then
+      encodeVersion = 1
+    else
+      encodeVersion = tonumber(header:match("%d+"))
+    end
+  end
```
oratory, line 683: `encoded` is an empty string after getting the `encodeVersion`, I suggest the following:
```lua
  local _, _, encodeVersion, encoded = inString:find("^(!WA:%d+!)(.+)$")
  if encodeVersion then 
    encodeVersion = tonumber(encodeVersion:match("%d+"))
  else
    encoded, encodeVersion = inString:gsub("^%!", "")
  end
```

Wago is ready and will decode strings with the new serialization format. Some time after the next version of WA is released and distributed I can have Wago start using the new lib when generating the strings as well.

InfusOnWoW:
I think we should do that. 20% smaller is a clear win, even if the backwards compability break is a bit annoying.

PR #2262, opened 2020-07-09
Threat trigger update

### LINKED ISSUE #2257: Missing % of Progress in Animation for Threat Situation trigger
I'd like to add a custom animation based on % of Progress for a Progress WA with a Threat Situation trigger. However the only option is Duration (s).

Here's what it looks like right now when default Threat Situation trigger:
https://i.imgur.com/xF2toqq.png

Here's what it looks like when I use a custom trigger to get threat situation (what I want):
https://i.imgur.com/5Xu5Tfh.png

### LINKED ISSUE #2261: Threat Situation status as condition
Please add support for accessing Threat Situation status in conditions. My use case is a progress bar, showing my current threat percentage, and then playing a sound ("aggro!") if the threat status changes to 1 (higher than tank) or greater. Therefore, not only accessing the value, but comparing it like "if threat status > higher than tank (1)" is important to me.

Currently, the alternative is to create five triggers; one for the naked threat situation, and one for each of the four threat statuses with additional constraints on the threat status, and then checking "if trigger 2 is active" etc, but this doesn't allow for greater/less than comparisons.

mrbuds:
# Description

Use threat value and calculated total threat as value & total
Store Aggro & Status for use in conditions fixes #2261
Set canHaveDuration fixes #2257

## Type of change

- [x] Bug fix
- [x] Breaking change => change %p & %t values

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -5864,17 +5864,19 @@ WeakAuras.event_prototypes = {
       local ret = [[
         local unit = %s
         local ok = true
-        local aggro, status, threatpct, rawthreatpct, threatvalue
+        local aggro, status, threatpct, rawthreatpct, threatvalue, threattotal
         if unit then
           aggro, status, threatpct, rawthreatpct, threatvalue = WeakAuras.UnitDetailedThreatSituation('player', unit)
+          threattotal = (threatvalue or 0) * 100 / (threatpct or 1)
         else
           status = UnitThreatSituation('player')
           aggro = status == 2 or status == 3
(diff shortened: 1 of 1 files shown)


mrbuds:
> 
> 
> I assume you have a good reason for that backwards incompatible change to %p?

I'v seen examples of threat bar where they want to know exactly how much threat they can do before pulling aggro from tank,.
While this change still require minimal maths do get the differential, it will make the trigger work with the new tick mark option to show a mark at a value from total.

InfusOnWoW:
Ah that's good enough imho and makes it more in line with how other triggers work.

PR #2289, opened 2020-07-17
Make CD Widget functional when using a static progress type on an icon aura.

nullKomplex:
# Description

A user in discord reported that icon auras with a static progress type didn't display the CD widget. Upon investigation, it was never implemented. It currently simply hides the widget if the progress type is static. This PR would make the CD widget work as expected, showing a static CD widget set to the given progress values.

## Type of change

# How Has This Been Tested

- [x] Used an aura with multiple triggers to test switching between multiple progress types and values.

# Checklist


InfusOnWoW:
I think this is good, but we can't do this in a backwards compatible way.

So I think we should merge it to the shadowlands branch

PR #2305, opened 2020-07-18
Shadowlands templates

mrbuds:
# Description

- [x] remove azerite stuff
- [x] priest skills update
- [ ] more classes
- [ ] covenant abilities

### DIFF
--- a/WeakAurasTemplates/TriggerTemplatesData.lua
+++ b/WeakAurasTemplates/TriggerTemplatesData.lua
@@ -2041,9 +2041,11 @@ else
           { spell = 527, type = "ability"}, -- Purify
           { spell = 586, type = "ability", buff = true}, -- Fade
           { spell = 2061, type = "ability", overlayGlow = true}, -- Flash Heal
+          { spell = 8092, type = "ability", requiresTarget = true}, -- Mind Blast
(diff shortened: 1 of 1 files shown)


WeakAurasTemplates/TriggerTemplatesData.lua
```diff
@@ -5185,6 +5185,7 @@ if not WeakAuras.IsClassic() then
     for className, class in pairs(templates.class) do
       for specIndex, spec in pairs(class) do
         for _, section in pairs(spec) do
+          if section.args then
```
InfusOnWoW, line 5196: Don't be afraid of fixing the indentation

InfusOnWoW: Ah you did in the follow up commit

InfusOnWoW:
I don't think we need to fix all classes in one PR. It's good enough if you are happy enough with priests, we should just merge it.

Stanzilla:
Do we want to kill the templates right away? Because we're publishing this version of WA with the prepatch which still does have Azerite and Essences.

InfusOnWoW:
Yes, the few people that want to create auras for essences can do it manually.

mrbuds:
i'v rebased, and i agree that it can be merged now and do more updates later

PR #2307, opened 2020-07-18
Set an absolute minimum to border size of 1

emptyrivers:
Because otherwise we risk a division by 0 error in the backdrop mixin


InfusOnWoW:
I think the right thing might be to change what we pass to SetBackdrop for edgeSize = 0. I think we are supposed to pass edgeFile = nil, edgeSize = nil then.

InfusOnWoW:
And just for documentation purposes: I then tried how it behaved on BfA with a manual setting of 0. None of Model, Group or the Border sub element showed anything that looked good.

PR #2310, opened 2020-07-19
add fullCircle option

### LINKED ISSUE #2053: Unclear behavior or bug in the Arc length option
There is a dynamic group with t (t>=2) children. Value of the Grow option is counter clockwise or clockwise; Constant factor - radius. If I set the Arc length to n (n<360) degrees, t items will be at n/t degrees apart instead of n/(t-1). As a result, arc with an angle of less than n degrees will be formed by these items.

From discussion in Discord:

>- Nightwarden: rivers 
>1. Could you rename the Arc length option? It's an angle, and length is not measured in degrees.
>2. You made mistake in line 608, 651 in the file WeakAuras/RegionTypes/DynamicGroup.lua. It should be divided by (numVisible-1) because the number of spaces/gaps between n objects is (n-1), not n.
>- rivers: I’m not normally opposed to changing names, but I don’t know about this one. So far, I haven’t seen anyone not understand what the option is supposed to do, where that misunderstanding could only be attributed to its name.
>- rivers: Additionally, since we don’t support arbitrary curves (without writing code, anyways), it is reasonable to speak of the angle as an arc length, just as we usually use kilograms when talking about weight even though it should be some scaling of newtons instead
>- rivers: As for your second point, if I went with what you say, then the first and last element would always fully overlap when the user uses a constant radius arc, and allows the curve to span the full circle
>- rivers: e.g. with 3 items visible, we would expect them to be 120 degrees apart. If we used your suggestion then they would be 360/(3-1) = 180 degrees apart, which would be a bug. Users would complain that “one of the items is always missing”
>- rivers: So, if you did notice a bug in the behavior, you have either mis-identified the source, or were not specific enough
>- rivers: Hmm, well at least in the Anglo-sphere we use kilograms to measure weight instead of newtons. I don’t know about other languages
>- Nightwarden: Case of 360 degrees (full circle) ​​can be considered separately. Right now, if I set arc length to 90 degrees, 2 items will be at 45 degrees apart instead of 90.
>- rivers: Ok, then the correct solution is to grant 2π-θ radians (where θ is the arc length that the user chose) to the gap between the last and first element
>- rivers: And evenly size the remaining gaps
>- rivers: Can you write a ticket for that?
>- Nightwarden: No :slight_smile:
>- Nightwarden: ctr+c ctr+v from discord chat
>- rivers: Hmm I suppose that my code already implicitly does that
>-  rivers: So, then this needs to be a new option

Link in code - [WeakAuras/RegionTypes/DynamicGroup.lua](https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/RegionTypes/DynamicGroup.lua#L608); lines 567, 608

emptyrivers:
#Description

Closes #2053

## Type of change

# How Has This Been Tested

it work

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -1,4 +1,4 @@
-local internalVersion = 33;
+local internalVersion = 34;
 
 -- Lua APIs
 local insert = table.insert
@@ -4125,6 +4125,18 @@ function WeakAuras.Modernize(data)
       end
     end
 
+  end
+
+  -- Introduced in July 2020 in Shadolands
+  if data.internalVersion < 34 then
+    if data.regionType == 'dynamicgroup'
+    and (data.grow == "CIRCLE" or data.grow == "COUNTERCIRCLE") then
+      if data.arcLength == 360 then
+        data.fullCircle = true
+      else
+        data.fullCircle = false
+      end
+    end
   end
 
 
--- a/WeakAuras/RegionTypes/DynamicGroup.lua
+++ b/WeakAuras/RegionTypes/DynamicGroup.lua
@@ -26,6 +26,7 @@ local default = {
   yOffset = 0,
   radius = 200,
   rotation = 0,
+  fullCircle = true,
   arcLength = 360,
   constantFactor = "RADIUS",
   frameStrata = 1,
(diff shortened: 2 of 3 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -4127,6 +4127,19 @@ function WeakAuras.Modernize(data)
 
   end
 
+  -- Introduced in July 2020 in Shadolands
+  if data.internalVersion < 34 then
+    if data.regionType == 'dynamicgroup'
+    and (data.grow == "CIRCLE" or data.grow == "COUNTERCIRCLE")
+    and data.fullCircle == nil then
```
InfusOnWoW, line 4134: data.fullCircle will always be true here. 

We unfourtunately validate the settigns of a aura against the defaults before Modernize. (Which looks like a bug to me.) And that Modernize will add the missing fullCircle bool.

I think just removing the data.fullCircle == nil here should work though.

emptyrivers: Ah, right. I forgot about that.

InfusOnWoW: You should test it once against old settings :)

PR #2313, opened 2020-07-20
Tick improvements

nullKomplex:
# Description
Improves upon the tick feature, adding the ability to use a texture along with many features that bring ticks to parity with sparks, including: Mirroring, desaturation, offsets, blend mode, and rotation.

## Type of change

# How Has This Been Tested

- [x] Used all settings in display.
- [x] Tested all new Conditions.

# Checklist


nullKomplex:
I would like to specifically request additional feedback on the following:

- The features that form parity with sparks.
- If they aren't overkill, should we be looking to migrate sparks to ticks?
- Localization string handling in the Options code. I have a feeling that wasn't optimal.

Additionally, I can't seem to crack why my texture chooser doesn't update the texture immediately.

WeakAuras/SubRegionTypes/Tick.lua
```diff
@@ -314,6 +321,15 @@ local function modify(parent, region, parentData, data, first)
   region.automatic_length = data.automatic_length
   region.tick_thickness = data.tick_thickness
   region.tick_length = data.tick_length
+  region.use_texture = data.use_texture
+  region.tick_texture = data.tick_texture
+  region.tick_blend_mode = data.tick_blend_mode
+
+
+  if data.use_texture then
+    WeakAuras.SetTextureOrAtlas(region.texture, data.tick_texture)
```
InfusOnWoW: I think we should default to CLAMPTOBLACKADDITIVE as the wrap mode (3. and 4. argument)

InfusOnWoW:
I've rebased this on top of the current shadow lands branch. It looks good, except that the options code could use some reordering and has a few other small issues.

With you current layout, the "Blend Mode" text, that is title of that option pushes down the Choose button.

I think a better ordering is:

1) Use Texture (upper case T) + Blend Mode
2) Texture (remove the Spark)
3) Desaturate / Choose

As to your questions:
Well, I would like to move over sparks to ticks. Sparks have two additional options:
* Automatic rotation: I think this one can be removed. The Modernize code needs to set the right rotation on modernizing, but no one will miss the automatic rotation.
* Hide on (Empty, Full, Full/Empty): I think that's a usefull option. I think what would work, is if that option is only offered on the "Offset from Progress" mode

Comparing sparks to ticks, I just noticed that combination of "Offset from Progress" + Smooth Progress doesn't work "correctly", that needs to be fixed before we can move sparks to ticks, though doesn't need to be part of this PR.

As to the texturePicker not working correctly, that's due to it not bein able to handle sub elements yet, I'll figure out how to fix that in the best way.

InfusOnWoW:
I do have a change on my branch that should make the TexturePicker usable for you.
.

InfusOnWoW:
I think this is ready now.

PR #2330, opened 2020-07-27
Shadowlands

InfusOnWoW:


Stanzilla:
can we do StopMotion.lua instead

InfusOnWoW:
> can we do StopMotion.lua instead or ideally come up with a better name :D

Done, the former. The name is by now well-established, all 3 users know it.

PR #2338, opened 2020-07-29
Smallpatches

InfusOnWoW:


Stanzilla:
What's the reason for the private space? Currently not a fan of the name/casing but naming things is hard.

InfusOnWoW:
> 
> 
> What's the reason for the private space? Currently not a fan of the name/casing but naming things is hard.

It's not accessible outside of the addon. (That does include our options code atm, but I know how to fix that.)

I want to move various functions that the user should not call to that table. We currently have roughly 700 members in WeakAuras, I want to reduce that significantly. Most of our options code e.g. should not be exposed at all. It'll be a long-term project though.

PR #2340, opened 2020-07-29
fix pvp talents

emptyrivers:


emptyrivers:
Current shadowlands build errors in the middle of the login thread (thus rendering the entire addon unusable), so we might want to release quickly-ish.

Stanzilla:
On it

PR #2358, opened 2020-08-03
guard against div/0

### LINKED ISSUE #2353: Shadowlands Beta: Divide by zero error in groupCountFunc when dead
## Describe the bug

Died on Shadowlands beta.  I have an aura that triggers on <= 0.8 of the group missing Battle Shout.

Expected nothing special.

Instead, WA started throwing this errors repeatedly (up to 8k as I write this, ~3m since dying)

```
Message: [string "--[==[ Error in 'Unknown' ]==]     return f..."]:2: Division by zero
Time: Sun Aug  2 21:30:58 2020
Count: 10071
Stack: [string "--[==[ Error in 'Unknown' ]==]     return f..."]:2: Division by zero
[string "--[==[ Error in 'Unknown' ]==]     return function(count, max)
      local fraction = count/max
      if(fraction <= 0.8) then
        return true
      else
        return false
      end
    end
  "]:2: in function `groupCountFunc'
[string "@Interface\AddOns\WeakAuras\BuffTrigger2.lua"]:989: in function <Interface\AddOns\WeakAuras\BuffTrigger2.lua:988>
[string "@Interface\AddOns\WeakAuras\BuffTrigger2.lua"]:1031: in function <Interface\AddOns\WeakAuras\BuffTrigger2.lua:1009>
[string "@Interface\AddOns\WeakAuras\BuffTrigger2.lua"]:1355: in function <Interface\AddOns\WeakAuras\BuffTrigger2.lua:1350>
[string "@Interface\AddOns\WeakAuras\BuffTrigger2.lua"]:1674: in function <Interface\AddOns\WeakAuras\BuffTrigger2.lua:1667>

Locals: count = 0
max = 0
(*temporary) = nil
(*temporary) = nil
(*temporary) = "Division by zero"
```

### To Reproduce

Died.  The errors are raised while waiting to release and while a ghost.  They stop the moment you resurrect.

1. Install aura from the linked gist below.
2. Die
3. Profit?

3.0.0-alpha4

- [] Classic
- [] Retail
- [X] Shadowlands beta

The aura is unchanged from the one I use on retail, where this does not happen.

### Additional Info

Export string for the aura:  https://gist.github.com/jf647/813013aab273cabbf5100ff739409bee

emptyrivers:
we shouldn't have to do this...thanks blizz
fixes #2353

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -327,7 +327,10 @@ WeakAuras.function_strings = {
   ]],
   count_fraction = [[
     return function(count, max)
-      local fraction = count/(max ~= 0 and max or math.huge) -- because PTR always complains about div/0
+      if max == 0 then
+        return false
+      end
+      local fraction = count/max
       if(fraction %s %s) then
         return true
       else
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -327,7 +327,7 @@ WeakAuras.function_strings = {
   ]],
   count_fraction = [[
     return function(count, max)
-      local fraction = count/max
+      local fraction = count/(max ~= 0 and max or math.huge) -- because PTR always complains about div/0
       if(fraction %s %s) then
         return true
       else


WeakAuras/Prototypes.lua
```diff
@@ -327,7 +327,7 @@ WeakAuras.function_strings = {
   ]],
   count_fraction = [[
     return function(count, max)
-      local fraction = count/max
+      local fraction = count/(max ~= 0 and max or math.huge) -- because PTR always complains about div/0
```
InfusOnWoW, line 330: I'd use max == 0 and 0 or count/max, which is imho easier to read.

emptyrivers, line 330: I don't know if infinity is the correct option, or if we should go with 1. Technically I think no matter what we do, this would make unintuitive results whenever blizzard is giving us silly data.

InfusOnWoW: Having thought about it, imho the best was to handle this is that the count_fraction method returns false for max == 0.
so basically.

emptyrivers: I did that now.

PR #2379, opened 2020-08-10
re-enable model region for classic

mrbuds:
# Description
After seeing WeakAuras was forked on a chinese forum with models support https://bbs.nga.cn/read.php?tid=22060034&rand=116, my tests show that wrong models does not crash the game anymore.

I use the same ModelPaths.lua found on nga.cn forum for classic

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -31,7 +31,7 @@ jobs:
         path: .release/
 
     - name: Create Classic Package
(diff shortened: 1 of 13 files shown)


mrbuds:
I need to test the generated classic package to make sure i'v not forget anything

Vardex:
The fork also add some options for `Model` region and `barModel` subregion like portrait zoom and cam distance scale.
It might be a good idea (maybe in a different pr) to add those.

mrbuds:
should be ready now

mrbuds:
woops it's not ready found something

mrbuds:
need to fix model subregion picker for classic

PR #2392, opened 2020-08-14
Smallpatches

InfusOnWoW:


WeakAuras/Transmission.lua
```diff
@@ -1231,7 +1231,7 @@ end
 
 function WeakAuras.MatchInfo(data, children, target)
   -- match the parent/single aura (if no children)
-  local oldParent = target or findMatch(data)
+  local oldParent = target or findMatch(data, children)
```
emptyrivers, line 1234: Ah, interesting. When did that get changed?

Makes me look like an ignorant fool in another ticket, though 😄

InfusOnWoW: I think it accidentally worked before 04e45acee2d6884b2ef386a20b99d5e4f4d39edd

PR #2416, opened 2020-08-23
Add combat timer trigger

asaka-wa:
# Description

Combat Timers are a very frequently requested Aura. There are Custom Options available that will likely remain popular as they provide functionality like tying in to Exorsus Notes and so on. But a generic trigger that provides a simple value the the time spent in combat, and a way to trigger/Condition based on that time would be useful for various applications I think.

## Type of change

# How Has This Been Tested

- [ ] Test A
- [ ] Test B

# Checklist

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -3376,6 +3376,64 @@ do
   end
 end
 
+-- Combat Timer
+local combatTimerWatchFrame
+function WeakAuras.WatchCombatTimer(timerType)
+  if not combatTimerWatchFrame then
+    combatTimerWatchFrame = CreateFrame("frame")
+    combatTimerWatchFrame:SetScript("OnEvent",
+      function(self, event, ...)
+        if event == "PLAYER_REGEN_DISABLED" then
+          self.combatStart = GetTime()
+        elseif event == "PLAYER_REGEN_ENABLED" then
+          self.combatStart = nil
+          self.combatTimer = nil
+          WeakAuras.ScanEvents("COMBAT_TIMER_UPDATE")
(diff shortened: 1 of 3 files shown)


asaka-wa:
Well what you describe is what I was going for. But, of course, Buds is right and sending a ScanEvent onUpdate isn't good. But the onUpdate has got to happen somewhere, since it's a timer. And since it's a timer that goes up it can't be handled by the normal timed progress systems. 

I've started moving things over to the watchframe only firing status update events and will incorporate ScheduleScan next.

asaka-wa:
I feel like I've reached a bit of an impasse here. 
I understand the advice being given, though they're opposing views in important ways. But the problem as I see it, is that this timer counts *up* and therefore can't reuse the mechanisms in place to handle regular timed progress info. So I need to have a value on the trigger that updates every frame while the trigger is active, without having to recheck whether the trigger should be active every frame. I've looked through how various other triggers are built but don't see a ready example to guide me on this. 

If this whole idea is better to just wait for the inclusion of the "Active for x seconds" Condition, which would presumably introduce handling of timers going up in a much wider way, then that's fine.

InfusOnWoW:
Right, that seems to require a bit more work. Changing the state each frame is certainly not what I want. Instead that might require a new progress type.

asaka-wa:
Yup. 
So without adding that kind of thing we could produce a trigger that actives at a given point in combat/encounter, and carries a start time that the user could format in custom text if they needed a current fight duration. It does feel like having the current fight duration as a basic text replacement is core to what the trigger should provide though. Stripping it down to a start time doesn't feel good enough. 
Shall I close this PR for now then?

PR #2457, opened 2020-09-09
Update Actions

Stanzilla:
- Now uses a GitHub Action for Discord status posts
- Sets git fetch depth to a slightly more sane 250 instead of 0, v2 of the action is still buggy with tags so this can probably be removed in the future
- Try to add the PR number to the name of the archive that is generated


InfusOnWoW:
I have absolutely no clue what you are doing in this PR....

Stanzilla:
Great, right?

PR #2461, opened 2020-09-10
adds bleed icon

Causese:
adds bleed icon

# Description

adds bleed icon to the existing encounter journal icon table
## Type of change

# How Has This Been Tested

- [x] returned it ingame in %c trigger update

# Checklist

- [x] "might" have added whitespace

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -2794,6 +2794,7 @@ WeakAuras.EJIcons = {
   poison =    "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:39:57:39:57|t",
   disease =   "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:71:89:39:57|t",
   enrage =    "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:103:121:39:57|t",
+  bleed =     "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:158:192:32:64|t",
 }
 
 WeakAuras.reset_swing_spells = {}


WeakAuras/Types.lua
```diff
@@ -2794,6 +2794,7 @@ WeakAuras.EJIcons = {
   poison =    "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:39:57:39:57|t",
   disease =   "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:71:89:39:57|t",
   enrage =    "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:103:121:39:57|t",
+  bleed =     "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:158:192:32:64|t",
```
InfusOnWoW, line 2797: Does UnitAura return bleed for bleeds now?

Also this should target the shadowlands branch.

Causese: No UnitAura doesn't return bleed, just wanted to complete the ej icons table.

InfusOnWoW: Well we only use it for the debuff icon in BuffTrigger2, so what's the benefit withotu that?

Causese: weakauras does something special regarding the icon size when returning the icon via the global instead of using the escape sequence. (the size is bigger by default while maintaining the position)

e.g if you return WeakAuras.EJIcons.bleed in %c with text size 20 then the icon will be bigger than if you returned "|TInterface\\EncounterJournal\\UI-EJ-Icons:::::256:64:158:192:32:64|t"

I wanted to hardcore all the bleed icons in my weakauras without having to adjust text size & x/y coordinates and in order to keep it consistent with the debuffType size etc, I decided to add to the table.

but you're right, there's also other stuff in the table that is not used by the bufftrigger. it's just one of those things that make life easier

InfusOnWoW: Hmm right we have includes the tank/healer/dps icons in there, so we can add Bleed.

InfusOnWoW (approved):

PR #2492, opened 2020-09-19
Shadowlands

InfusOnWoW:


WeakAurasOptions/TriggerOptions.lua
```diff
@@ -115,6 +117,34 @@ local function AddOptions(allOptions, data)
...
+  triggerOptions["addTriggerOption"] = {
+    __title = L["Add Trigger"],
+    __order = 5000,
+    __withoutheader = true,
+    __topLine = true,
+    __collapsed = false,
+    addTrigger = {
+      type = "execute",
+      width = WeakAuras.normalWidth,
+      name = L["Add Trigger"],
```
mrbuds, line 129: At first look i though there was an issue with the add button missing, I think this button should include the green "+" texture Interface\\AddOns\\WeakAuras\\Media\\Textures\\add

mrbuds (commented):
About the auto collapse on triggers options, if i select an aura with more than 1 trigger, they are collapsed as wanted, but then if i select an other aura with only 1 trigger this trigger is not expanded, and if i create a new aura, the default "aura" trigger is collapsed on creation

WeakAurasOptions/TriggerOptions.lua
InfusOnWoW: All the other Add buttons don't have that. I think that's just somethign to get used to.

InfusOnWoW:
The collapse/expand state is shared by all triggers. I previously had a slightly different approach, but there are too many edge cases to consider with e.g. multi selection and trigger merging that I currently don't want to do that.

It's something that I need to eventually tackle.

mrbuds:
Using this PR `aura > unit = nameplate > type = debuff > autoclone checked` doesn't seem to show anything on classic

InfusOnWoW:
Fixed the aura bug, and improved the trigger expand/collapse behaviour.

mrbuds (approved):

PR #2636, opened 2020-10-26
Classic: add support for MAINTANK and MAINASSIST filter

### LINKED ISSUE #2635: Support for "maintank" tag in triggers
Hello,

I'm aware that WeakAuras2 is not ment for creating stand alone unit frames, but it is a very effective tool to attach additional information to them!

For example I created a WeakAura, that displays all incoming heals on my target as progress bars. Here is a screenshot:
target incoming heals

With the Blizzard UI and several other unit frame addons (for example Pitbull, or ElvUI) you can tag units as "main tanks" and they will appear in a seperate frame. 
From a healers perspective you want to have more information about whats happening with the tanks, so you might make this frame bigger and display more information on it than your normal raid frames. Also you will likely heal with mouse over casting and not have the tank in your target.

Here is a little concept what I want to do:
Concept

All of this can already be done with WeakAuras for a specific UID like raid1, party3, or target, but not for the main tank frames. So i kindly ask for an option in the trigger setup to relate a WA to a main tank :-)

Since there is no UID like "maintank1", but only a tag, I was thinking of something like getting the UnitID, or UnitName from the parent frame? But what the heck do I know :) Maybe there is a better approach.

Best regards
Admyra

mrbuds:
# Description
Add MAINTANK and MAINASSIST filter instead of UnitGroupRolesAssigned filters for classic
- bufftrigger2
- various status triggers
- Load condition

fixes #2635

## Type of change

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -74,6 +74,8 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
+local UnitGroupRolesAssigned = not WeakAuras.IsClassic() and UnitGroupRolesAssigned or function() return "DAMAGER" end
+
 -- keyed on unit, debuffType, spellname, with a scan object value
 -- scan object: id, triggernum, scanFunc
 local scanFuncName = {}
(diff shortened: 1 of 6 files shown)


InfusOnWoW:
I don't think we should overload the group role assigned in that way. I'd rather have it as a separate setting, since technically it's also in retail. I don't think we should enable it for retail, but if anyone ever wants it, enabling it is easy and pain free. If we mix it in that way, enabling it for retail becomes much harder.

mrbuds:
You'r right, i have updated PR to make MAINTANK/MAINASSIST role distinct from healer/tank/dps role

InfusOnWoW (approved):

PR #2713, opened 2020-11-15
Add "Tooltip on Mouseover" support to Text

isker:
# Description

Aura and Icon can both display tooltips on mouseover in certain situations.  Add the same feature to Text.

I am trying to create "data text" WeakAuras like the old [SLDataText](https://www.wowinterface.com/downloads/info8539-SLDataText.html).  These are Texts displaying basic information, sometimes with additional information in a tooltip on mouseover.  For example, the "ping" text shows one value returned from `GetNetStats`.  When mousing over the displayed value, all of the values returned from `GetNetStats` are displayed in a tooltip.

The [TSU docs](https://github.com/WeakAuras/WeakAuras2/wiki/Trigger-State-Updater-(TSU)#tooltips) make it seem like they would be an appropriate way to accomplish this by stating: "TSU triggers have the 'Tooltip on Mouseover' option available by default in the Display tab."  However, this is not true for Text.  There does not seem to be a reason why it couldn't be.

This PR copies code and option definitions straight from Icon into Text.

An alternative solution is to use an Icon WA with a fake 0% alpha icon providing the tooltip, overlaid on the text.  However, just supporting tooltips in Text seems like a better solution.

## Type of change

- [x] New feature (non-breaking change which adds functionality

## How Has This Been Tested

https://wago.io/TiTa2ecqo in Classic.

Without my changes, no tooltip on mouseover.  With my changes, there is an option in Display that, when checked, shows a tooltip configured in the TSU on mouseover.

InkedWoWScrnShot_111520_141445_LI

WoWScrnShot_111420_200703

## Checklist

### DIFF
--- a/WeakAurasOptions/RegionOptions/Text.lua
+++ b/WeakAurasOptions/RegionOptions/Text.lua
@@ -247,6 +247,14 @@ local function createOptions(id, data)
       }
     },
 
+    useTooltip = {
+      type = "toggle",
+      width = WeakAuras.normalWidth,
+      name = L["Tooltip on Mouseover"],
+      hidden = function() return not OptionsPrivate.Private.CanHaveTooltip(data) end,
+      order = 51
+    },
+
     endHeader = {
       type = "header",
       order = 100,
--- a/WeakAuras/RegionTypes/Text.lua
+++ b/WeakAuras/RegionTypes/Text.lua
@@ -95,6 +95,21 @@ local function modify(parent, region, data)
   region:SetWidth(region.width);
   region:SetHeight(region.height);
 
+  local tooltipType = Private.CanHaveTooltip(data);
+  if(tooltipType and data.useTooltip) then
+    if not region.tooltipFrame then
+      region.tooltipFrame = CreateFrame("frame", nil, region);
+      region.tooltipFrame:SetAllPoints(region);
+      region.tooltipFrame:SetScript("OnEnter", function()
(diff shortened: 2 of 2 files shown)


WeakAurasOptions/RegionOptions/Text.lua
```diff
@@ -35,6 +35,15 @@ local function createOptions(id, data)
         OptionsPrivate.ResetMoverSizer();
       end,
     },
+
+    useTooltip = {
+      type = "toggle",
+      width = WeakAuras.normalWidth,
+      name = L["Tooltip on Mouseover"],
+      hidden = function() return not OptionsPrivate.Private.CanHaveTooltip(data) end,
+      order = 11
```
InfusOnWoW, line 44: That's certainly the wrong place to put it. It should be much further down.

isker: Can you be more specific?  This option is pretty high up in the other two regions that have it.

InfusOnWoW: You put it before the custom function that belongs to the the text edit above.

It should be below Font Flags.

isker:
@InfusOnWoW how's this look?

InfusOnWoW:
There's a argument that we should factor out the tooltip functionality into a sub element. But your PR argueably doesn't make that more complicated, so I've merged it.

isker:
Thanks!

PR #2773, opened 2020-12-02
Expand [status - crowd controlled] trigger

asaka-wa:
# Description

The current CC trigger is entirely binary. People often want to know if they're controlled by a specific CC type for which they have a counter. 

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6650,7 +6650,7 @@ Private.event_prototypes = {
     args = {
       {
         name = "controlType",
-        display = L["Specific CC Type"],
+        display = L["Specific Type"],
         type = "select",
         values = "loss_of_control_types",
         conditionType = "select",
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
   ["Crowd Controlled"] = {
     type = "status",
     events = {
-      ["unit_events"] = {
-        ["player"] = {"UNIT_AURA"}
+      ["events"] = {
+        --"PLAYER_CONTROL_LOST",
+        --"PLAYER_CONTROL_GAINED",
+        --"LOSS_OF_CONTROL_ADDED",
+        "LOSS_OF_CONTROL_UPDATE"
       }
     },
-    force_events = "UNIT_AURA",
+    force_events = "LOSS_OF_CONTROL_UPDATE",
     name = L["Crowd Controlled"],
+    canHaveDuration = true,
+    statesParameter = "one",
+    init = function(trigger)
+      local ret = [=[
+          local show = false
+          local use_controlType = %s
+          local controlType = %s
+          local inverse = %s
+          local use_interruptSchool = %s
+          local interruptSchool = tonumber(%q)
+          local duration, expirationTime, spellName, icon, spellName, spellId, locType, lockoutSchool, controlText, _
+          for i = 1, C_LossOfControl.GetActiveLossOfControlDataCount() do
+            local data = C_LossOfControl.GetActiveLossOfControlData(i)
+            if data then
+              if (not use_controlType)
+              or (data.locType == "PACIFYSILENCE" and (controlType == "SILENCE" or controlType == "PACIFY"))
+              or (data.locType == "STUN_MECHANIC" and controlType == "STUN")
+              or (data.locType == controlType and (controlType ~= "SCHOOL_INTERRUPT" or ((not use_interruptSchool) or data.lockoutSchool == interruptSchool)))
+              then
+                spellId = data.spellID
+                spellName, _, icon = GetSpellInfo(data.spellID)
+                duration = data.duration
+                expirationTime = data.startTime + data.duration
+                locType = data.locType
+                lockoutSchool = data.lockoutSchool
+                controlText = data.displayText
+                show = true
+                break
+              end
+            end
+          end
+      ]=]
+      ret = ret:format(
+        trigger.use_controlType and "true" or "false",
+        type(trigger.controlType) == "string" and "[["..trigger.controlType.."]]" or [["STUN"]],
+        trigger.use_inverse and "true" or "false",
+        trigger.use_interruptSchool and "true" or "false",
+        trigger.interruptSchool or 0
+      )
+      return ret
+    end,
     args = {
       {
-        name = "controlled",
-        display = L["Crowd Controlled"],
-        type = "tristate",
-        init = "not HasFullControl()"
-      }
+        name = "controlType",
+        display = L["Specific CC Type"],
+        type = "select",
+        values = "loss_of_control_Types",
+        conditionType = "select",
+        test = "true",
+        default = "STUN",
+      },
+      {
+        name = "interruptSchool",
+        display = L["Interrupt School"],
+        type = "select",
+        values = "main_spell_schools",
+        conditionType = "select",
+        default = 1,
+        conditionType = "select",
+        test = "true",
+        enable = function(trigger) return trigger.controlType == "SCHOOL_INTERRUPT" end,
+      },
+      {
+        name = "inverse",
+        display = L["Inverse"],
+        type = "toggle",
+        test = "true",
+      },
+      {
+        name = "name",
+        display = L["CC Name"],
+        hidden = true,
+        conditionType = "string",
+        init = "controlText",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "spellName",
+        display = L["CC Spell Name"],
+        hidden = true,
+        conditionType = "string",
+        init = "spellName",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "locType",
+        display = L["CC Type"],
+        hidden = true,
+        conditionType = "string",
+        init = "locType",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "spellId",
+        display = L["CC Spell Id"],
+        hidden = true,
+        conditionType = "number",
+        init = "spellId",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "lockoutSchool",
+        display = L["Interrupted School"],
+        hidden = true,
+        conditionType = "string",
+        init = "lockoutSchool and lockoutSchool > 0 and GetSchoolString(lockoutSchool) or nil",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "icon",
+        hidden = true,
+        init = "icon",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "duration",
+        hidden = true,
+        init = "duration",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "expirationTime",
+        init = "expirationTime",
+        hidden = true,
+        store = true,
+        test = "true",
+      },
+      {
+        name = "progressType",
+        hidden = true,
+        init = "'timed'",
+        store = true,
+        test = "true",
+      },
+      {
+        hidden = true,
+        test = "(inverse and not show) or (not inverse and show)",
+      },
     },
-    automaticrequired = true
+    automaticrequired = true,
   },
   ["Cast"] = {
     type = "status",
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6603,9 +6603,6 @@ Private.event_prototypes = {
     type = "status",
     events = {
       ["events"] = {
-        --"PLAYER_CONTROL_LOST",
-        --"PLAYER_CONTROL_GAINED",
-        --"LOSS_OF_CONTROL_ADDED",
         "LOSS_OF_CONTROL_UPDATE"
       }
     },
@@ -6626,8 +6623,6 @@ Private.event_prototypes = {
             local data = C_LossOfControl.GetActiveLossOfControlData(i)
             if data then
               if (not use_controlType)
-              or (data.locType == "PACIFYSILENCE" and (controlType == "SILENCE" or controlType == "PACIFY"))
-              or (data.locType == "STUN_MECHANIC" and controlType == "STUN")
               or (data.locType == controlType and (controlType ~= "SCHOOL_INTERRUPT" or ((not use_interruptSchool) or bit.band(data.lockoutSchool, interruptSchool) > 0)))
               then
                 spellId = data.spellID
@@ -6661,6 +6656,8 @@ Private.event_prototypes = {
         conditionType = "select",
         test = "true",
         default = "STUN",
+        init = "locType",
+        store = true,
       },
       {
         name = "interruptSchool",
@@ -6671,6 +6668,8 @@ Private.event_prototypes = {
         default = 1,
         test = "true",
         enable = function(trigger) return trigger.controlType == "SCHOOL_INTERRUPT" end,
+        init = "lockoutSchool",
+        store = true,
       },
       {
         name = "inverse",
@@ -6695,13 +6694,6 @@ Private.event_prototypes = {
         test = "true",
       },
       {
-        name = "locType",
-        display = L["Type"],
-        hidden = true,
-        store = true,
-        test = "true",
-      },
-      {
         name = "spellId",
         display = L["Spell Id"],
         hidden = true,
@@ -6712,7 +6704,7 @@ Private.event_prototypes = {
       },
       {
         name = "lockoutSchool",
-        display = L["Interrupted School"],
+        display = L["Interrupted School Text"],
         hidden = true,
         init = "lockoutSchool and lockoutSchool > 0 and GetSchoolString(lockoutSchool) or nil",
         store = true,
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6621,14 +6621,14 @@ Private.event_prototypes = {
           local inverse = %s
           local use_interruptSchool = %s
           local interruptSchool = tonumber(%q)
-          local duration, expirationTime, spellName, icon, spellName, spellId, locType, lockoutSchool, controlText, _
+          local duration, expirationTime, spellName, icon, spellName, spellId, locType, lockoutSchool, name, _
           for i = 1, C_LossOfControl.GetActiveLossOfControlDataCount() do
             local data = C_LossOfControl.GetActiveLossOfControlData(i)
             if data then
               if (not use_controlType)
               or (data.locType == "PACIFYSILENCE" and (controlType == "SILENCE" or controlType == "PACIFY"))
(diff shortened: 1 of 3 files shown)


Nightwarden24:
- Could you add C_LossOfControl.GetActiveLossOfControlDataCount and C_LossOfControl.GetActiveLossOfControlData to  to [.luacheckrc](https://github.com/WeakAuras/WeakAuras2/blob/master/.luacheckrc#L721) file?

- I guess there are no these functions in Classic.
   Yeah, they have different names
   
   
   I can't say anything about structure of return value

- List of loss_of_control_Types is not complete. What about at least these?

   + BANISH (Cyclone - Druid; Imprison if use Detainment PvP-talant - Demon hunter)
   + DISORIENT (Dragon's Breath - Mage; Scatter Shot - Hunter; Blind - Rogue)
   + INCAPACITATE (Repentance - Paladin; Paralysis - Monk?; Gouge - Rouge; Incapacitating Roar - Druid)
   + POLYMORPH (Polymorph - Mage; Hex - Shaman)
   + SLEEP (Song of Chi-Ji - Monk; Hibernate - Druid)

   Does the function return data for them? Or am I completely wrong about return type for control abilities? And existence of types needs to be checked in Classic  ¯\\\_(ツ)_/¯
   
   P.S. Not sure about these types
   + SAP (Sap - Rouge [SAP or INCAPACITATE?]; Imprison without PvP-talent - Demon Hunter [SAP or INCAPACITATE?])
   + SNARE (Slow - Mage; Thunder Clap - Warrior)
   + HORROR (Mortal Coil - Warlock; Vampiric Touch - Priest {If it's dispelled, the dispeller flees in Horror for 3 sec})

mrbuds:
>     * I guess there are no these functions in Classic.

This trigger is disabled on classic https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/Prototypes.lua#L7851

Nightwarden24:
> > ```
> > * I guess there are no these functions in Classic.
> > ```
> 
> This trigger is disabled on classic https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/Prototypes.lua#L7851

I apologize for that

asaka-wa:
There's no API or table of types on Blizzard's side so I could certainly be missing some types. I took the list I'm using from an addon. I know that Hex, for example, is actually a "PACIFYSILENCE" effect with the displayText "Polymorphed" so just because some types you might intuitively think must exist aren't listed that doesn't mean they're missing.

Nightwarden24:
Ah, I thought you checked types manually (at least the most typical abilities or "ambiguous" cases) :disappointed:

I didn't rely on intuition or display text. I was looked at mechanic of abilities on wow.tools and wowhead.com and thought it had something to do with locType :flushed:

asaka-wa:
My guess is that it's like the debuffTypes on auras, and is more about what spells can remove the effects than it is about what the effects themselves are.

mrbuds:
I found this with a google search that lead me to <https://www.curseforge.com/wow/addons/d4hp/issues/5>
```
AllowedLOCTypes = <table> {
1 = "STUN_MECHANIC"
2 = "STUN"
3 = "PACIFYSILENCE"
4 = "SILENCE"
5 = "FEAR"
6 = "CHARM"
7 = "PACIFY"
8 = "CONFUSE"
9 = "POSSESS"
10 = "SCHOOL_INTERRUPT"
11 = "ROOT"
12 = "FEAR_MECHANIC"
}
```

asaka-wa:
Yeah, I handle all of those.

mrbuds:
I see, mhm i would have preferred re-affect the value of locType, and make it's condition of type "select" instead of "string"

asaka-wa:
To be fair, the controlType value is using the "select" conditionType so locType can probably just have its conditionType removed. 
One would be the version as they select it, while the other would include those combined ones like "PACIFYSILENCE" but I don't think that's really going to be relevant.

WeakAuras/Types.lua
```diff
@@ -1199,6 +1199,29 @@ for i = 1, 5 do
   Private.totem_types[i] = totemString:format(i);
 end
 
+Private.loss_of_control_Types = {
```
InfusOnWoW, line 1202: lowercase "T"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+      },
+      {
+        name = "inverse",
+        display = L["Inverse"],
+        type = "toggle",
+        test = "true",
+      },
+      {
+        name = "name",
+        display = L["CC Name"],
```
InfusOnWoW, line 6684: Just "Name"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        hidden = true,
+        conditionType = "string",
+        init = "controlText",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "spellName",
+        display = L["CC Spell Name"],
```
InfusOnWoW, line 6693: Just "Spell Name"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        hidden = true,
+        conditionType = "string",
+        init = "spellName",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "locType",
+        display = L["CC Type"],
```
InfusOnWoW, line 6702: Just "Type"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        hidden = true,
+        conditionType = "string",
+        init = "locType",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "spellId",
+        display = L["CC Spell Id"],
```
InfusOnWoW, line 6711: Just "Spell Id"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        hidden = true,
+        conditionType = "number",
+        init = "spellId",
+        store = true,
+        test = "true",
```
InfusOnWoW, line 6716: operator_types == "only_equal"

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+          local use_interruptSchool = %s
+          local interruptSchool = tonumber(%q)
+          local duration, expirationTime, spellName, icon, spellName, spellId, locType, lockoutSchool, controlText, _
+          for i = 1, C_LossOfControl.GetActiveLossOfControlDataCount() do
+            local data = C_LossOfControl.GetActiveLossOfControlData(i)
+            if data then
+              if (not use_controlType)
+              or (data.locType == "PACIFYSILENCE" and (controlType == "SILENCE" or controlType == "PACIFY"))
+              or (data.locType == "STUN_MECHANIC" and controlType == "STUN")
+              or (data.locType == controlType and (controlType ~= "SCHOOL_INTERRUPT" or ((not use_interruptSchool) or data.lockoutSchool == interruptSchool)))
```
InfusOnWoW, line 6631: Did you check what the lookoutSchool is if a spell with multiple spell schools is interrupted?

E.g. frostfire, chaos bolt or whatever else tehre might be?

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        hidden = true,
+        conditionType = "string",
+        init = "lockoutSchool and lockoutSchool > 0 and GetSchoolString(lockoutSchool) or nil",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "icon",
+        hidden = true,
+        init = "icon",
```
InfusOnWoW, line 6730: The init here should be unnecessary.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        store = true,
+        test = "true",
+      },
+      {
+        name = "duration",
+        hidden = true,
+        init = "duration",
```
InfusOnWoW, line 6737: The init here should be unnecessary.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        store = true,
+        test = "true",
+      },
+      {
+        name = "expirationTime",
+        init = "expirationTime",
```
InfusOnWoW, line 6743: The init here should be unnecessary.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+      {
+        name = "progressType",
+        hidden = true,
+        init = "'timed'",
+        store = true,
+        test = "true",
+      },
+      {
+        hidden = true,
+        test = "(inverse and not show) or (not inverse and show)",
```
InfusOnWoW, line 6757: You could write that as "inverse ~= show", which is a bit less readable. Though it's fine in this case.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        conditionType = "number",
+        init = "spellId",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "lockoutSchool",
+        display = L["Interrupted School"],
+        hidden = true,
+        conditionType = "string",
```
InfusOnWoW, line 6722: I think this should not be available as a  condition check.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        init = "spellName",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "locType",
+        display = L["CC Type"],
+        hidden = true,
+        conditionType = "string",
+        init = "locType",
```
InfusOnWoW, line 6705: You don't have to use init here.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        init = "controlText",
+        store = true,
+        test = "true",
+      },
+      {
+        name = "spellName",
+        display = L["CC Spell Name"],
+        hidden = true,
+        conditionType = "string",
+        init = "spellName",
```
InfusOnWoW, line 6696: You don't have to init here.

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,162 @@ Private.event_prototypes = {
...
+        default = "STUN",
+      },
+      {
+        name = "interruptSchool",
+        display = L["Interrupt School"],
+        type = "select",
+        values = "main_spell_schools",
+        conditionType = "select",
+        default = 1,
+        conditionType = "select",
```
InfusOnWoW, line 6672: you set conditionType twice

nullKomplex:
What is the difference between interrupted school and interrupt school?

WeakAuras/Prototypes.lua
asaka-wa: How does the value get stored without it?

WeakAuras/Prototypes.lua
asaka-wa: Hmm, yeah I'll see if I can test this later.

Nightwarden24: I checked: for example, lockoutSchool is [124](https://wow.gamepedia.com/COMBAT_LOG_EVENT#Spell_School) when interrupting [Chaos Bolt](https://www.wowhead.com/spell=116858)

asaka-wa: Thanks for checking. 
That would mean that they would be locked out of all [Fire, Nature, Frost, Shadow, Arcane], right?
So if I aim to provide a dropdown for the base school lockouts I need to associate the higher level school values with each of the base types. Or, actually looks like I can go via a bitmask.

WeakAuras/Prototypes.lua
InfusOnWoW: It should work with just store = true if you have a local variable of that name.

asaka-wa: Ho! nice.

WeakAuras/Prototypes.lua
Nightwarden24: Another example: lockoutSchool is [72](https://wow.gamepedia.com/COMBAT_LOG_EVENT#Spell_School) when interrupting [Full Moon](https://www.wowhead.com/spell=274283)
Yes, all my abilities of nature and arcane school are locked out =(

asaka-wa: Yup `bit.band(data.lockoutSchool, interruptSchool)` should handle it. I'll test tonight to make sure.

asaka-wa:
Okay, changes made 👍

Nightwarden24:
I'm totally against your list of control types. So I've checked class abilities that affect control, but have not considered spells from other sources. Result in file.

Interesting:

- Slow, knockback, taunt don't count as loss of control (captain obvious) and, accordingly, are not returned by the function.
- Events fired when using some abilities have more than one locType. Classic example: priest's [Mind Bomb](https://www.wowhead.com/spell=205369) (FEAR_MECHANIC and ROOT). I don't know whether you consider and handle it.
- displayText is just value from LOSS_OF_CONTROL_DISPLAY_X variable (Global string), not a localized string returned by client or server.

I got the following list of locTypes:

+ CHARM
+ CONFUSE
+ DISARM
+ FEAR
+ FEAR_MECHANIC
+ NONE - I don't know what to do with that. Ignore?
+ PACIFY
+ PACIFYSILENCE
+ POSSESS
+ ROOT
+ SCHOOL_INTERRUPT
+ SILENCE
+ STUN
+ STUN_MECHANIC

Types shouldn't be combined or omitted. They are too different by its nature. Don't take their names literally. I'll try to explain how I see logic of their assignment

* CHARM - ? I don't know
* CONFUSE - You are incapacitated and walking in a circle (i.e. there are small movements). Any damage will instantly interrupt the effect.
* DISARM - You can't use weapon or shield, but can still attack with hands
* FEAR - It's same as FEAR_MECHANIC, but lasts significantly less time (2-3 seconds)
* FEAR_MECHANIC - You are incapacitated and fleeing in an uncontrolled direction. Damage may cancel the effect.
* PACIFY - You can't attack, but can cast spell and move?
* PACIFYSILENCE - You can't attack or cast any spell, but can still move
* POSSESS - You can't control your actions. Someone else does it.
* ROOT -  You can't move, but can attack or cast spell
* SCHOOL_INTERRUPT - You can't cast only spell of specific school
* SILENCE - You can't cast any spell, but can move and attack
* STUN - You are incapacitated and don't move. Damage will break the effect or you are invulnerable 
* STUN_MECHANIC - Any stun. You are incapacitated except using special spell (example [Enraged Regeneration](https://www.wowhead.com/spell=184364)). Damage won't cancel effect.

Сheck out the file and see what abilities belong to each category. Perhaps, you will also notice the difference between them.

Your loss_of_control_types list will generate such questions:

- I want to track situation when I'm stunned and should use Dispersion to reduce incoming damage. But icon appers when Cyclone or Paralysis is applied.
- More exotic. Why does aura trigger on [Hex](https://www.wowhead.com/spell=51514)? I can't deal physical damage (i.e. attack) when I'm silenced? Opposite situation. I want to track situation when I can't attack and I'm silenced (there is bunch of other spells with this type of control), but aura is active when on [Garrote](https://www.wowhead.com/spell=703).

I suggest the following option:

+ CHARM - no idea, Charm?
+ CONFUSE = Confuse or Disorientation (closer to spell description for me)
+ DISARM = Disarm
+ FEAR = Horror
+ FEAR_MECHANIC = Fear
+ PACIFY = Pacify
+ PACIFYSILENCE - don't know
+ POSSESS = Possess
+ ROOT = Root
+ SCHOOL_INTERRUPT = Interrupt
+ SILENCE = Silence
+ STUN = Incapacitation
+ STUN_MECHANIC = Stun

P.S. Multiple select in cases of control types and spell schools would be really convenient.

[Loss of control.xlsx](https://github.com/WeakAuras/WeakAuras2/files/5660637/Loss.of.control.xlsx)

asaka-wa:
Well, FEAR_MECHANIC is new to me, I'll incorporate that. I would've handled it in the same was as STUN, rolling both together but I don't really care about that, if anyone thought that they were important to be separate then, sure, whatever. 

PACIFYSILENCE though, sure could be handled separately – my code only gets easier with this change, but it feels like if people are under a PACIFY effect they want to know about PACIFYSILENCE too. 

All these names are too abstract for users to just *get*, as you've showed above it isn't clear even to us. WA sometimes just throws API names at people and lets them sink or swim, but that only works in oft-used parts of the API where collectively the user community interacts with it enough to just learn the way to refer to things. In this case I think that it would just mystify people and they'd just go ahead and pick "STUN", for example, and hope it does what they need. 

So, I dunno. I could accept your advice and simplify the trigger. I'm just not sure that best serves the users.

Nightwarden24:
To pick stun, user has to select STUN_MECHANIC. STUN isn't a stun. Sorry, I don't know how else to explain this to you.

Give users all control types. Don't make decisions for them.

By the way, you still haven't added "new" functions to [.luacheckrc](https://github.com/WeakAuras/WeakAuras2/blob/master/.luacheckrc#L721) file.

asaka-wa:
Sure, thanks for your thoughts. I'll bear them in mind.

InfusOnWoW:
@Nightwarden24 are you happy with asakawa's latest changes?

WeakAuras/Prototypes.lua
```diff
@@ -6602,21 +6602,145 @@ Private.event_prototypes = {
...
+    end,
     args = {
       {
-        name = "controlled",
-        display = L["Crowd Controlled"],
-        type = "tristate",
-        init = "not HasFullControl()"
-      }
+        name = "controlType",
+        display = L["Specific CC Type"],
```
InfusOnWoW, line 6653: Change to "Specific Type".

Nightwarden24:
Good. Filling the list with types exactly as they are is not a bad idea. But I'm afraid that some description will still need to be added later.

One thing before the merge that you forgot about again: luacheck.

asaka-wa:
Unless something official can be found then I see nothing to be gained by adding some ad hoc guesswork. 

I'm not forgetting, I have no intention of doing it. It's not a system I know and I'm sure the addon's maintainers can adjust that if need be.

mrbuds (approved):

PR #2835, opened 2020-12-21
Spell/Item in Range triggers

nullKomplex:
# Description
Adds triggers for spell and item ranges. These are commonly requested on discord and the current options require the use of the conditions of another trigger.

## Type of change

## How Has This Been Tested

- [x] Tested various spells on allied and enemy targets
- [x] Tested an item on allied targets
- [x] Created a new aura to avoid the mishap from the last PR

## Checklist

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -1030,7 +1030,7 @@ function GenericTrigger.LoadDisplays(toLoad, loadEvent, ...)
               if not genericTriggerRegisteredEvents["COMBAT_LOG_EVENT_UNFILTERED"] then
                 eventsToRegister["COMBAT_LOG_EVENT_UNFILTERED"] = true;
               end
-            elseif (event == "FRAME_UPDATE") then
+            elseif (event == "FRAME_UPDATE") or (event == "WA_SPELL_RANGECHECK") then
               register_for_frame_updates = true;
             else
               if (genericTriggerRegisteredEvents[event]) then
@@ -1386,6 +1386,7 @@ do
   local update_frame = nil
   WeakAuras.frames["Custom Trigger Every Frame Updater"] = update_frame;
   local updating = false;
+  local rangeThrottle = 0
 
   function Private.RegisterEveryFrameUpdate(id)
     if not(update_clients[id]) then
@@ -1399,6 +1400,10 @@ do
       update_frame:SetScript("OnUpdate", function()
         if not(WeakAuras.IsPaused()) then
           WeakAuras.ScanEvents("FRAME_UPDATE");
+          if (not rangeThrottle or GetTime() - rangeThrottle > 0.2) then
+            rangeThrottle = GetTime()
+            WeakAuras.ScanEvents("WA_SPELL_RANGECHECK")
+          end
         end
       end);
       updating = true;
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -7897,6 +7897,129 @@ Private.event_prototypes = {
     automaticrequired = true
   },
 
+  ["Spell in Range"] = {
+    type = "status",
+    events = {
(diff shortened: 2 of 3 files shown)


WeakAuras/GenericTrigger.lua
```diff
@@ -1399,6 +1400,10 @@ do
       update_frame:SetScript("OnUpdate", function()
         if not(WeakAuras.IsPaused()) then
           WeakAuras.ScanEvents("FRAME_UPDATE");
+          if (not rangeThrottle or GetTime() - rangeThrottle > 0.2) then
+            rangeThrottle = GetTime()
+            WeakAuras.ScanEvents("WA_SPELL_RANGECHECK")
+          end
```
nullKomplex, line 1406: Is this an alright way to handle this throttled range check?

nullKomplex: Also this incidentally allows users to use this event as a throttled trigger. I know this was explicitly denied in the past but I would not know of a proper method to make this unavailable to the users.

WeakAuras/Prototypes.lua
```diff
@@ -7897,6 +7897,228 @@ Private.event_prototypes = {
...
+        "PLAYER_TARGET_CHANGED",
+        "WA_SPELL_RANGECHECK",
+      }
+    },
+    name = L["Item in Range"],
+    init = function(trigger)
+      trigger.unit = trigger.unit or "target"
+      trigger.itemName = trigger.itemName or 0
+      local itemName = type(trigger.itemName) == "number" and trigger.itemName or "[["..trigger.itemName.."]]"
+      -- IsItemInRange returns nil for no target, true for out of range, and true for in range.
```
InfusOnWoW, line 8037: Soemething about this is wrong :)

WeakAuras/GenericTrigger.lua
InfusOnWoW: We usually factor it out into a component thatthen sends the events, but in your case that might be a bit overkill as the event sending is absolutely trivial.

There's some code for the existing range checking trigger that does send a event on a periodic timer already, maybe using that works. I haven't looked at it in detail.

I wouldn't worry about hiding that.

WeakAuras/Prototypes.lua
```diff
@@ -7897,6 +7897,228 @@ Private.event_prototypes = {
     automaticrequired = true
   },
 
+  ["Spell in Range"] = {
+    type = "status",
+    events = {
+      ["events"] = {
+        "PLAYER_TARGET_CHANGED",
```
InfusOnWoW, line 7904: There are events missing for other types of unit changes.

WeakAuras/GenericTrigger.lua
nullKomplex: Range Check is actually fired every frame. I recalled it being throttled as well but that isn't the case. The only other range checking is done via Conditions which uses its own system instead of events (https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAuras/Conditions.lua#L688) and you have informed me in the past that Conditions don't receive ScanEvents.

nullKomplex: (I did also originally test with that event to verify it wasn't being received. I'm using the same event as the Condition.)

InfusOnWoW: Interesting, that is something that should be changed though.

WeakAuras/Prototypes.lua
mrbuds: I think Infus refers to 
https://github.com/WeakAuras/WeakAuras2/blob/72c8bb70b69920010525121ace197e5572517aac/WeakAuras/Prototypes.lua#L2057-L2064

mrbuds: minus AddUnitRoleChangeInternalEvents line

PR #2836, opened 2020-12-22
Bonus Id Trigger fixes and inverse option.

nullKomplex:
# Description

Adds an inverse option to the Item Bonus Id trigger. Will show Legendary Power Name and Icon if the option is selected (now renamed to cover both the icon and the fallback name).

## Type of change

## How Has This Been Tested

- [x] Thoroughly tested all bugged behaviors
- [x] Thoroughly tested equipping and unequipping a legendary with inverse toggled both on and off
- [x] Created a brand new aura and selected this trigger to make sure the previous breaking bug no longer exists with this new fix

## Checklist

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6394,7 +6394,7 @@ Private.event_prototypes = {
       {
         name = "icon",
         hidden = "true",
-        init = "icon or 'Interface\\Icons\\INV_Misc_QuestionMark'",
+        init = "icon or 'Interface/Icons/INV_Misc_QuestionMark'",
         store = "true",
         test = "true",
       },
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -6356,12 +6356,21 @@ Private.event_prototypes = {
     init = function(trigger)
       trigger.itemBonusId = trigger.itemBonusId or ""
       local ret = [=[
-        local useLegendaryIcon = %s
-        local itemBonusId, itemId, itemName, itemIcon, itemSlot, itemSlotString = WeakAuras.GetBonusIdInfo(%q)
+        local fetchLegendaryPower = %s
+        local item = %q
+        local inverse = %s
+        local useItemSlot, slotSelected = %s, %d
+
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -6379,10 +6388,10 @@ Private.event_prototypes = {
       },
       {
         name = "legendaryIcon",
-        display = L["Use Legendary Power Icon"],
+        display = L["Fetch Legendary Power"],
         type = "toggle",
         test = "true",
-        desc = L["Displays the icon for the Legendary Power that matches this bonus id."],
+        desc = L["Fetches the name and icon of the Legendary Power that matches this bonus id."],
```
nullKomplex, line 6394: I checked all the modernize code and there was only one instance of renaming variables and this change did not fit that scenario so I left the variable name alone instead of changing it and modernizing.

InfusOnWoW: Yeah it's usually not worth doing that. Because:
* We (devs/alpha testster) don't usually test the modernize code as well as the rest.
* The cost of a mistake in it are much higher. There's always the potential to make mistakes.

WeakAuras/Prototypes.lua
```diff
@@ -6354,21 +6354,31 @@ Private.event_prototypes = {
...
+        local useItemSlot, slotSelected = %s, %d
+
+        local itemBonusId, itemId, itemName, icon, itemSlot, itemSlotString = WeakAuras.GetBonusIdInfo(item)
         local itemBonusId = tonumber(itemBonusId)
-        local icon = useLegendaryIcon and WeakAuras.GetLegendaryIcon(itemBonusId) or itemIcon
+        if fetchLegendaryPower then
+          itemName, icon = WeakAuras.GetLegendaryData(itemBonusId or item)
+        end
+
+        local slotValidation = ((useItemSlot and itemSlot == slotSelected) or (not useItemSlot))
```
InfusOnWoW, line 6370: There's a extra () here compared to the style we usually use.

nullKomplex: Oops, that was left over from a lot of tinkering with inverse. I fix.

PR #2855, opened 2020-12-30
Add Talent NOT selected Load Option

### LINKED ISSUE #2780: "Talent NOT selected" Loading
I have a couple auras which load when I have a certain talent selected and another aura that loads when I don't have that aura. (The talent is Glacial Spike for frost mage. My icicle tracker needs to behave differently when I have Glacial Spike so I use separate auras depending on whether it's selected.) Since there is no "talent NOT selected" option, I set it to load when either of the other two are selected. This works fine normally, but in Torghast, I've had the situation where I have both Glacial Spike and one of the other two talents. This makes it so BOTH sets of auras are loaded.

A way to specify that an aura is to load if a talent is NOT selected would solve this issue. (Perhaps other missing "NOT" logic could also be implemented, e.g. "Spell NOT known.")

Xiiph:
Fixes #2780

## Type of change

## How Has This Been Tested
- [ ] In **RETAIL** only.

## Checklist

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -1274,6 +1274,14 @@ Private.load_prototype = {
       events = WeakAuras.IsClassic() and {"CHARACTER_POINTS_CHANGED"} or {"PLAYER_TALENT_UPDATE"}
     },
     {
+      name = "ignoretalent",
+      display = L["|cFFFF0000Not|r Talent selected"],
+      type = "multiselect",
+      values = valuesForTalentFunction,
+      test = "(not WeakAuras.CheckTalentByIndex(%d))",
+      events = WeakAuras.IsClassic() and {"CHARACTER_POINTS_CHANGED"} or {"PLAYER_TALENT_UPDATE"}
+    },
+    {
       name = "pvptalent",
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
While that fixes that ticket, it's not a change I necessarily want.

The wow api provides enough information to distinguish between "known" and "selected" talents. With the former being a superset of the later.

Since Blizzard keeps on reusing the talent gained mechanic and quite likely some class will get a talent gain legendary in this expansion, WA should support that properly.

That is, allow for all of:
1) Talent selected
2) NOT Talent selected
3) Talent known
4) NOT Talent known.

2) is not really necessary since 1) would cover that with the multi-selection. 
But 1) is currently also not possible.

There are a few options that could be done:
1) Just add all Talent selected/NOT Talent known and have 3 differnt kind of talent load options
2) Add some special widget that allows toggling how the talent selected/known load option works
3) Some other solution

The problem with the first solution is that tripling the amount of talent load options is not free. Especially since the load page is already quite overloaded.

So I'd rather have a 2) or 3) solution.

InfusOnWoW:
I've had time to work on the talent load option and there's now a pull request for that which obsoletes this.

PR #2875, opened 2021-01-06
update WeakAurasModelPaths from wow.tools csv export

### LINKED ISSUE #2829: Model Path Module Missing Shadowlands listings.
## Describe the bug

Not much else to say except the current expansion doesn't show in the 3D Model Paths module when creating an aura.

```plaintext
PASTE ERROR HERE
```

### To Reproduce

**Screenshots**

- [ ] Classic
- [ ] Retail

mrbuds:
# Description

To get source csv files from https://wow.tools/files/ click on "Listfile" >Community CSV Retail and/or ClassicRetail

image

Fixes #2829

### DIFF
--- a/.github/scripts/.gitignore
+++ b/.github/scripts/.gitignore
@@ -1,0 +1,2 @@
+*.csv
+.wowtools.json
--- a/.github/scripts/csv_to_lua.lua
+++ b/.github/scripts/csv_to_lua.lua
@@ -1,0 +1,132 @@
+
+-- usage:
+-- # lua.exe ./list_to_table.lua <release>
+
+local releases = {
+  classic = {
+    input = "classic_list.csv", -- get it from https://wow.tools/casc/listfile/download/csv/build?buildConfig=9ad6ad5306deb8eed364b64cc628ac98
+    output = "ModelPathsClassic.lua",
+    generate = true
+  },
+  retail = {
+    input = "retail_list.csv", -- get it from https://wow.tools/casc/listfile/download/csv/build?buildConfig=26291f284f42494375d511d1fc120216 (last retail build when i write that)
+    output = "ModelPaths.lua",
+    generate = true
+  }
+}
+
+require "table"
+
+if not arg[1] or not releases[arg[1]] then
+  print(arg[0], "<classic|retail>")
+  return
+end
+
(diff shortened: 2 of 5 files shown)


StanzillaManticore:
Two things:

1) We could just run that in the CI to automate it
2) We could ask @marlamin for a ?latest API so we don't have to hardcode version numbers

Stanzilla:
Two things:

1) We could just run that in the CI to automate it
2) We could ask @marlamin for a ?latest API so we don't have to hardcode version numbers

mrbuds:
> 
> 
> Two things:
> 
>     1. We could just run that in the CI to automate it
> 
>     2. We could ask @Marlamin for a ?latest API so we don't have to hardcode version numbers

The files are not small and the script take quite some time to run on my computer.
Could CI run an automated update only if target's build has changed instead of running it for every new commit ?

Marlamin:
Do you just need (named) m2s for this?

mrbuds:
> 
> 
> Do you just need (named) m2s for this?

Hey Marlamin :)
Yes, only named m2 with their fileId

Marlamin:
>We could ask @Marlamin for a ?latest API so we don't have to hardcode version numbers  

A latest build per branch listfile endpoint would need to keep track of the current build for a branch and that's kind of a pain for me to do right now in the CASC backend that generates the listfile. I added a quick separate API that returns the current build config for a certain branch that you can use: https://wow.tools/api.php?type=currentbc 

> Yes, only named m2 with their fileId  

I added an optional `typeFilter` GET to the listfile downloads that returns only filenames ending in the specified string. 
e.g. `https://wow.tools/casc/listfile/download/csv/build?buildConfig=26291f284f42494375d511d1fc120216&typeFilter=m2` will only returns filenames for that build ending in m2. Hopefully that saves a bit of processing on your end of things.

A note I always give people who want to rely on WoW.tools for things, I give no guarantee of uptime/API stability

Stanzilla (approved):
Did you mark the .sh file as +x in git?

mrbuds:
> 
> 
> Did you mark the .sh file as +x in git?

yes with `git update-index --chmod=+x ./wowtools.sh`

PR #2879, opened 2021-01-09
Some improvements in sandbox

TimoshN:
aura_env in C_Timer
proxify _G access
virtual Blizz-like frames in sandbox


emptyrivers:
The failing check is for a lot of trailing whitespace, fyi.

Also there's a lot of commented debug code added, can you remove those comments, please?

PR #2900, opened 2021-01-24
Cast trigger: add latency overlay

mrbuds:
# Description

New Overlay Latency option for Cast trigger using delay between CURRENT_SPELL_CAST_CHANGED and UNIT_SPELLCAST_START events

This is the method use by popular cast bar addon Quartz.

## Type of change

## How Has This Been Tested

expected:
- stop casting a self heal before progress reach latency overlay = no heal
- stop casting a self heal after progress reach latency overlay and before full progress = heal

results: seems good :)

image

## Additional comment

I don't like the hack in GenericTrigger.lua to make RunTriggerFunc not return false because of `statesParameter = "unit"` require an arg1, but i have not found a better way to do it

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -729,9 +729,6 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
     local updateTriggerState = false;
     for triggernum, data in pairs(triggers) do
       local allStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
-      if event == "CURRENT_SPELL_CAST_CHANGED" then -- hack for Cast trigger latency
-        arg1 = "player"
-      end
       if (RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2, ...)) then
         updateTriggerState = true;
       end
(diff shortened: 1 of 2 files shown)


WeakAuras/GenericTrigger.lua
```diff
@@ -729,6 +729,9 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
     local updateTriggerState = false;
     for triggernum, data in pairs(triggers) do
       local allStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
+      if event == "CURRENT_SPELL_CAST_CHANGED" then -- hack for Cast trigger latency
```
InfusOnWoW, line 732: The way this is usually done is:
1) Make a fake event WA_EVENT
2) Create a frame watching for CURRENT_SPELL_CAST_CHANGED
3) Which calls ScanEvents with WA_EVENT, "player"

See for example the UNIT_PET -> PET_UPDATE code

mrbuds: pushed an update fixing that thank you

WeakAuras/Prototypes.lua
```diff
@@ -6850,6 +6850,11 @@ Private.event_prototypes = {
         LibClassicCasterino.RegisterCallback("WeakAuras", "UNIT_SPELLCAST_INTERRUPTED", WeakAuras.ScanUnitEvents)
       end
       AddUnitEventForEvents(result, unit, "UNIT_TARGET")
+      if trigger.use_showLatency and unit == "player" then
+        WeakAuras.WatchForCastLatency()
```
InfusOnWoW, line 6854: That should be in the loadFunc.

PR #2944, opened 2021-02-18
Fix target type check to compare against a string

### LINKED ISSUE #2943: WeakAuras.Import checks type against nil instead of 'nil'
## Describe the bug
The `WeakAuras.Import` API call has the following check:
```lua
if type(target) ~= nil then
```

This is always `true`, even when the optional `target` parameter has not been specified or is `nil`. This causes the `.Import` call to always return `false, "Invalid update target. Importing as an unknown payload."` even when the import is successful.

The fix for this is simply changing the above check to:
```lua
if type(target) ~= 'nil' then
```

### To Reproduce
1. Add a `print` statement at the end of the `Import` function, or at the place where `.Import` is being called logging the status and message.
2. Open the aura import popup
3. Enter a valid import string
4. See the import info popup appear and a message in the chat frame with `false` `"Invalid update target. Importing as an unknown payload."`

**Screenshots**
image (5)

Yes

3.1.9

- [ ] Classic
- [X] Retail

Looks like it's been broken for at least two years: https://github.com/WeakAuras/WeakAuras2/commit/b831e888d2b15ed2e708c3cd5d55d6e34b98efdd

Snakybo:
# Description

This is a fix for the issue submitted in #2943, it simply changes the following check:
```lua
if type(target) ~= nil then
```

to:
```lua
if type(target) ~= 'nil' then
```

Fixes #2943

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested

I ran the reproduction steps listed in #2943:

1. Add a print statement at the end of the Import function, or at the place where .Import is being called logging the status and message.
2. Open the aura import popup
3. Enter a valid import string
4. ~See the import info popup appear and a message in the chat frame with false "Invalid update target. Importing as an unknown payload."~

and confirmed that step 4 is no longer the case and the function properly returns `true, ""` if the import was successful.

## Checklist

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -1684,7 +1684,7 @@ function WeakAuras.Import(inData, target)
     return nil, "Invalid import data."
   end
   local status, msg = true, ""
-  if type(target) ~= nil then
+  if type(target) ~= 'nil' then
     local targetData
     local uid = type(target) == 'table' and target.uid or target
     if type(uid) == 'string' then


emptyrivers:
Thanks for noticing that :3

PR #2948, opened 2021-02-20
Default options

InfusOnWoW:
# Description

Fixes #(issue)

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist


Stanzilla:
One file is ´DefaultOptions` the other `DefaultsOption` we should settle on one, I think I prefer `DefaultOptions` and I think it's more grammatically correct as well` @emptyrivers?

InfusOnWoW:
So, I've rebased this on top of current main and checked that some basic things work. I'm a bit suprised how much actually worked. The biggest missing thing is templates, where the introduction of defaults break the old way region defaults are mixed into templates. And the UI. E.g. the new UI clips the settings frame. And it's quite ugly. I guess it would much better if there was a bit of identation, but there's also a lot of options

Anyway a quick overview of the code:
Observe that in https://github.com/WeakAuras/WeakAuras2/blob/13f790d038f8dc6a5aa47012513478064a803514/WeakAuras/SubRegionTypes/SubText.lua#LL13C26-L13C26
the defaults for the SubText depend on the parent's region type. So depending on whether a subtext is added to an icon or an aurabar, e.g. the position, font flags, shadow settings are different.

This is a source of complexity, in this PR that is kept, thus:
a) There are still **base** defaults in SubText depending on the parent type:
https://github.com/InfusOnWoW/WeakAuras2/blob/44ffbd891ea3481e398464fd4078119681d339a1/WeakAuras/SubRegionTypes/SubText.lua#L10

b) There are **user** defaults settings that allow overriding these:
https://github.com/InfusOnWoW/WeakAuras2/blob/44ffbd891ea3481e398464fd4078119681d339a1/WeakAurasOptions/SubRegionOptions/SubText.lua#L528

So there's subtext_pb_font for overriding the font for progress bars, and subtext_icon_font for overriding the font on icons

c) How these override settings (that is subtext_pb_font) should be applied to the base defaults to get the effective defaults is the **mapping**
https://github.com/InfusOnWoW/WeakAuras2/blob/44ffbd891ea3481e398464fd4078119681d339a1/WeakAuras/SubRegionTypes/SubText.lua#L88

Each mapping has a name, that is the top-level key, a base, and a map that e.g. maps global settings paths to keys in the defaults table. So the values in the map, are the keys of the default table that are potentially overriden.

That's the core idea, some more random things:
Each user override has two settings, first if this override should apply to a) new auras/subelements or b) on imports or c) on both,
so Private.GetDefault/ Private.SetDefault return two values: a) the setting for when a override should be applied, and b) the actual override value. Th

The effective defaults tables are kept in a cache, the "defaults cache", which takes care of not recreating the defaults all the time.

There are a bunch of TODOs in the code that indicate areas where I though more work was needed.

PR #2960, opened 2021-03-03
fix combat log trigger's icon for classic

### LINKED ISSUE #2959: Aura Icon now showing up on Combat Log Aura Event Trigger
## Describe the bug

I've tried to create a WA which shows the icon of an aura applied to the player, but the correct icon does not appear on show, only the "question mark" icon.

### To Reproduce

Trigger
 - Trigger Type: Event
 - Event: Combat Log
 - Message Prefix: Spell
 - Message Suffix: Aura Applied
 - Destination Unit: Player

Display
 - Icon Source: Dynamic Information

**Screenshots**

https://streamable.com/h15am7

Yes

3.1.4

- [X] Classic
- [ ] Retail

Haven't tried.

mrbuds:
# Description

Use spellName instead of spellId to get icon id for classic

Fixes  #2959

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -3471,7 +3471,7 @@ Private.event_prototypes = {
       {
         hidden = true,
         name = "icon",
-        init = "spellId and select(3, GetSpellInfo(spellId)) or 'Interface\\\\Icons\\\\INV_Misc_QuestionMark'",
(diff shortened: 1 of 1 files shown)


MarkoRacz:
At the moment (3.1.9) using spellName instead of spellId in a Custom Event Trigger based WA only returns the Icon if the Player knows the spell in question (has it in Spellbook)

PR #2962, opened 2021-03-04
handle DBM & BW pause/resume callbacks

### LINKED ISSUE #2951: Add support for DBM pausing timers
DBM triggers are still firing even though the timer has paused.

I would like to see WA updated to support DBM paused timers.

**Additional context**
https://github.com/DeadlyBossMods/DeadlyBossMods/commit/ed1ae7ba2ae58d282de3e6134718fd36313fab20

mrbuds:
# Description
As I don't think we have a way to pause a timed aura, this will handle DBM_TimerPause like a DBM_TimerStop event, and DBM_TimerResume like a DBM_TimerStart event
and same for BigWigs_PauseBar and BigWigs_ResumeBar events

fixes #2951

I have not tested the change since it seems dbm will only fire them for a mythic boss i won't do

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -54,6 +54,7 @@ globals = {
 	"BigWigsLoader",
 	"CUSTOM_CLASS_COLORS",
 	"DBM",
+	"DBT",
 	"ElvUIPlayerNamePlateAnchor",
 	"GTFO",
 	"IndentationLib",
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -2846,13 +2846,15 @@ do
     local now = GetTime()
     nextExpire = nil
     for id, bar in pairs(bars) do
-      if bar.expirationTime < now then
-        bars[id] = nil
-        WeakAuras.ScanEvents("BigWigs_StopBar", id)
-      elseif nextExpire == nil then
-        nextExpire = bar.expirationTime
-      elseif bar.expirationTime < nextExpire then
-        nextExpire = bar.expirationTime
+      if not bar.paused then
+        if bar.expirationTime < now then
+          bars[id] = nil
+          WeakAuras.ScanEvents("BigWigs_StopBar", id)
+        elseif nextExpire == nil then
+          nextExpire = bar.expirationTime
+        elseif bar.expirationTime < nextExpire then
+          nextExpire = bar.expirationTime
+        end
       end
     end
 
@@ -2900,6 +2902,34 @@ do
         bars[text] = nil
         WeakAuras.ScanEvents("BigWigs_StopBar", text)
       end
+    elseif event == "BigWigs_PauseBar" then
+      local addon, text = ...
+      local bar = bars[text]
+      if bar then
+        bar.paused = true
+        bar.remaining = bar.expirationTime - GetTime()
+        WeakAuras.ScanEvents("BigWigs_PauseBar", text)
+        if recheckTimer then
+          timer:CancelTimer(recheckTimer)
+        end
+        recheckTimers()
+      end
+    elseif event == "BigWigs_ResumeBar" then
+      local addon, text = ...
+      local bar = bars[text]
+      if bar then
+        bar.paused = nil
+        bar.expirationTime = GetTime() + bar.remaining
+        bar.remaining = nil
+        WeakAuras.ScanEvents("BigWigs_ResumeBar", text)
+        if nextExpire == nil then
+          recheckTimer = timer:ScheduleTimerFixed(recheckTimers, bar.expirationTime - GetTime())
+        elseif bar.expirationTime < nextExpire then
+          timer:CancelTimer(recheckTimer)
+          recheckTimer = timer:ScheduleTimerFixed(recheckTimers, bar.expirationTime - GetTime())
+          nextExpire = bar.expirationTime
+        end
+      end
     elseif event == "BigWigs_StopBars"
     or event == "BigWigs_OnBossDisable"
     or event == "BigWigs_OnPluginDisable"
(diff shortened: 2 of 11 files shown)


mrbuds:
Bigwigs implemented pause/resume of timers as well, i need to handle it

InfusOnWoW:
I don't quite understand what pause+resume does. 

Is the bar still visible while paused?

mrbuds:
> 
> 
> I don't quite understand what pause+resume does.
> 
> Is the bar still visible while paused?

Both addon call Pause() and Resume() functions in libcandybar
```
--- Pauses a running bar
function barPrototype:Pause()
	if not self.paused then
		self.updater:Pause()
		self.paused = true
	end
end
--- Resumes a paused bar
function barPrototype:Resume()
	if self.paused then
		self.exp = GetTime() + self.remaining
		self.updater:Play()
		self.paused = nil
	end
end
```

So it seems yes bars stay on screen but stop their animation.
My PR is not doing that, it shortcut pause/resume to make them a stop/start, i don't know how complex it is to have same behavior as they do

InfusOnWoW:
Well, currently the auras don't support being paused/resumed at all. 

Adding support for it shouldn't be impossible, though currently state is suposed to contain a expirationTime. We would need to support e.g. having paused + remaining in the state, which if set are used instead of expirationTime.

Looking at the code, that would require changes in ProgressTexture/AuraBar/Icon/Tick and for %p.

That doesn't look too bad. Do you want to do that?

mrbuds:
That sound like a nice little project i can work on it

mrbuds:
So i have a working patch for pausing auras as you can see in this clip https://clips.twitch.tv/AmericanSlipperyMeerkatOneHand-iedtMRl60OJD-2rw that needs a bit more thoughts.

But after looking at dbm & bw timer triggers implementation in weakauras, especially for no-clones that use a lot of recheck scheduling to detect shortest duration matching bar, it's not going to be that easy

InfusOnWoW:
Well for the rechecking scheduling, you have to skip over all paused clones. 

And obviously recheck the scheduling on both Pause and Resume.

mrbuds:
i'v pushed the state.paused commit so you can have a look

mrbuds:
Bigwigs & DBM pause/resume is implemented
I used this group for tests https://wago.io/GUJCuatgb

MysticalOS:
For time being i added a clone of DBT.Options to DBM.Bars.options for backwards compat.

as for testing dbm pause/resume. you don't need to do boss, the one boss that does it has debug functions built in. just load castle mods then run the scripts from lua file.

for example

`/run DBM:GetModByName(2426):TestRestart(30)`
to start timers
`/run DBM:GetModByName(2426):TestPause()`
to pause them
`/run DBM:GetModByName(2426):TestResume()`
to resume them

also if whlie add it you want to teset add and remove time objects tests for those exist too.

mrbuds:
For dbm i used this dirty snippet to test pause/resume (included in https://wago.io/GUJCuatgb)
```
function(e)
    local bossName = "WA-DBM-TEST"
    local mod
    for k, v in pairs(DBM.Mods) do
        if v.id == bossName then
            mod = v
            break
        end
    end
    if not mod then
        mod = DBM:NewMod(bossName)
    end
    local cd = mod:NewCDTimer(5, 347350, "dbm wa test", nil, nil, 1)
    if e == "PLAYER_TARGET_CHANGED" then
        cd:Start()
    elseif e == "PLAYER_STARTED_MOVING" then
        cd:Pause() 
    elseif e == "PLAYER_STOPPED_MOVING" then
        cd:Resume() 
    end
end
```

mrbuds:
@emptyrivers can you confirm if this is the correct fix for the tocversion field issue that was brought up a few days ago ?

InfusOnWoW:
I had a quick look through the Pause code and it looks pretty good. I'll have to check a few details on some of the changes, but the general gist of it is very much how I think this should work.

MysticalOS:
Slightly off topic but I did some talking with a bunch of users about the recent freezing in raids related to Council and at first I was confused because the new string of DBT anchor errors in DBM was annoying but never caused freezes. The anchoring issues themselves were annoying because they couldn't be reproduced by myself or QartemisT. But when a subset of a subset of users were getting a repeating freeze i was even more stumped.

Then I noticed a common demoninator. The DBT anchor lua errors could happen anytime during fight but the freeze the answer was always "I froze during dark ritual". Why was the subset of subset of users freezing there?

I put that out of mind though and rushed updated DBT code out that fixed anchoring. Anchoring all fixed. no one was getting DBT errors anymore about anchoring. They were getting some callback errors instead because weak auras couldn't find DBM.Bars.options . Whatever little details...But then they froze again at dark ritual. That's when it finally hit me the freeze and weak auras failing were related.

They were all running the popular dark ritual weak aura that triggers off DBM timer. Because DBT was failing to create/anchor the dark ritual timer correctly the weak aura deadlocking and that's what caused freeze. Then when I pushed fix to anchoring they still froze because now weak aura was still failing but for a totally different reason.

I imagine in both scenarios it was spitting out hundreds of errors. Whether the freeze happened because WA code execution didn't know wtf to do with that, or failure in DBM Callback code executing callbacks on our end I don't know. This is something I want to look into. Regardless the freezing should be resolved i believe with the rewrite code in DBT combined with updated weak aura code knowing where new options are, but I still want to get bottom of where the freeze occurs so that if timers fail, no matter how spectacularly neither weak auras or DBM callback code deadlocks from it. On your end I'd just maybe add a safe exit that goes "what is this shit data? that's not supposed to be nil or broken, ignore this trash" somewhere (as long as it isn't expensive) and I'll look at callback code to make sure it also doesn't execute stupid shit unthrottled.

mrbuds:
@MysticalOS Do you have a link for this Dark Ritual aura and a copy of the error they get ?

Justw8:
> Bigwigs implemented pause/resume of timers as well, i need to handle it
> 
> cf https://github.com/BigWigsMods/BigWigs/blob/b359a4470cf8d7a3ccb0730ecb014a73b33d4175/CastleNathria/TheCouncilofBlood.lua#L611

I know this is closed and done now, but since we had someone report an issue today which was fixed in a bugfix, I came across this and just wanted to add: Bigwigs has had pause/resume bars for over 6 years now. https://github.com/BigWigsMods/BigWigs/commit/8dea709b1798acbacba9813888a39cef6cbc14e1

PR #2986, opened 2021-03-18
ensure a few "percentX" property doesn't return a boolean

mrbuds:
fix type error when they are use in conditions

fixes an error reported on discord https://pastebin.com/j8CHBhpi using this aura https://wago.io/8KnH9fmY3

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2061,7 +2061,7 @@ Private.event_prototypes = {
         name = "percentXP",
         display = L["Experience (%)"],
         type = "number",
-        init = "total ~= 0 and (value / total) * 100",
+        init = "total ~= 0 and (value / total) * 100 or nil",
         store = true,
         conditionType = "number"
       },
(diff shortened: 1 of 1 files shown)


mrbuds:
i'm not sure what is best value to return if total == 0 between 0 or 100, i made the choice to use 100, it may not be correct

WeakAuras/Prototypes.lua
```diff
@@ -2061,7 +2061,7 @@ Private.event_prototypes = {
         name = "percentXP",
         display = L["Experience (%)"],
         type = "number",
-        init = "total ~= 0 and (value / total) * 100",
+        init = "total ~= 0 and (value / total) * 100 or 100",
```
InfusOnWoW, line 2064: "or nil" should work.

mrbuds: that sound more sane, it's updated

InfusOnWoW:
Well now the commit message is wrong...

WeakAuras/Prototypes.lua
mrbuds: > 
> 
> Well now the commit message is wrong...

indeed, fixed

PR #2992, opened 2021-03-21
Smallpatches

InfusOnWoW:


WeakAuras/Modernize.lua
```diff
@@ -1225,5 +1225,68 @@ function Private.Modernize(data)
     end
   end
 
+  if data.internalVersion < 45 then
+    print(data.id)
```
mrbuds, line 1229: debug print left

WeakAuras/WeakAuras.lua
```diff
@@ -1356,6 +1356,14 @@ local function GetInstanceTypeAndSize()
     local difficultyInfo = Private.difficulty_info[difficultyIndex]
     if difficultyInfo then
       size, difficulty = difficultyInfo.size, difficultyInfo.difficulty
+    elseif size == "arena" then
+      if C_PvP.IsRatedArena() and not IsArenaSkirmish() then
+        size = "ratedarena"
+      end
+    elseif size == "pvp" then
+      if C_PvP.IsRatedBattleground() then
```
mrbuds, line 1364: IsRatedArena & IsRatedBattleground are not in C_PVP table on classic

InfusOnWoW: Oh. There is no rated arena in classic, right?

mrbuds: There is no arena at all (there will be in TBC), but shouldn't it be checked in case someone import an aura from retail ?

mrbuds (approved):

PR #3005, opened 2021-04-02
TBC/Classic Swing timer fixes

NilkemoryaOryhara:
# Description

Enable ranged swing timer for TBC. Add new TBC spell ranks and Steady Shot for swing timer resets. Add warrior/rogue ranged shots to reset melee swing timer.

## Type of change

## How Has This Been Tested

Testing swing timer speeds with a warrior character on Shattrath dummies in WoW TBC Beta client.

## Checklist

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -1911,7 +1911,7 @@ Private.swing_types = {
   ["off"] = SECONDARYHANDSLOT
 }
 
-if WeakAuras.IsClassic() then
+if WeakAuras.IsClassic() or WeakAuras.IsBC() then
   Private.swing_types["ranged"] = RANGEDSLOT
 end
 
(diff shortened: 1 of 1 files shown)


mrbuds:
Great thank you!
There was a few error, i edited it a bit

NilkemoryaOryhara:
> 
> 
> Great thank you!
> There was a few error, i edited it a bit

Thank you for looking this over and getting it added.

There is one small aspect of the changes that I think wasn't clearly communicated and was lost in your changes. Which is that there is a small inaccuracy with melee swing timers because mechanically warriors and rogues have their melee swing timers reset by shooting their ranged weapons. I was trying to include a fix for that for both classic era and TBC with the new combined shoot spell.

mrbuds:
That change looked wrong to me, i'd like a proof that the behavior is like you describe before commit a change to melee swing timer triggered by ranged weapon

NilkemoryaOryhara:
> 
> 
> That change looked wrong to me, i'd like a proof that the behavior is like you describe before commit a change to melee swing timer triggered by ranged weapon

Sure, no problem. Here is a clip of me testing Shoot, as well as mounting, resetting the main hand swing timer in TBC Beta with a WeakAuras Main Hand swing timer and a 2.9 speed main hand with no offhand equipped for clarity's sake. The behavior is the same on live classic, so it's not just a TBC beta bug.

https://streamable.com/nyjecp

The mounting behavior works mostly correctly, except for the first time where I moved before the cast bar fully finished and it didn't quite register as casting properly.

---

Additionally, while testing these sorts of things I've found that WeakAuras incorrectly counts dynamite casts as pausing or resetting the swing timer, when in reality it has no effect (you can even melee swing mid dynamite cast bar). But I'm not sure where the actual code to include dynamite or add it as an exception would be.

https://streamable.com/d2rk1y

mrbuds:
@NilkemoryaOryhara i pushed another commit to fix that, i'v only added Dense Dynamite to casts that do not trigger reset of swing timer, there is probably more

NilkemoryaOryhara:
@mrbuds Nice, seems to be working well for Dense Dynamite. I believe this can be applied to all bomb/grenade type explosives with a cast time, although it's a bit difficult to check all of them individually to be 100% sure. Fel Iron Bomb has identical swing time behavior for sure.

PR #3015, opened 2021-04-11
Delete moan.ogg

SpencerMalone:
# Description
Removal of rather obnoxious / inappropriate moan sound. If there is a particular way to notify users to automatically update this to another sound if this one is absent, more than happy to bring it in. Myself and others have found this to be off putting, and it's 2021, and as far as I can tell this was inherited from an ancestor of the addon, can we just remove this file?

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

(unsure if this require a doc update or not)

## How Has This Been Tested

I will admit to not running tests on this, was hoping to rely in CI

## Checklist


Stanzilla:
Why do you want to delete it? You can just not use it?

InfusOnWoW:
This needs to be at least a backwards compatible.

SpencerMalone:
Backwards compatibility would be nice, is there an already existing function that we can use, or a replacement file somebody could think of? I was kind of looking at https://commons.wikimedia.org/wiki/File:En-us-moan.ogg, although we would have to include the licensing.

Re: 
> Why do you want to delete it?
In the code of conduct is listed...
> The use of sexualized language or imagery and unwelcome sexual attention or advances
I think this is very clearly audio intended as sexual in nature, and IMO is an unwelcome thing to distribute as a default sound.

SpencerMalone:
https://freesound.org/people/AntumDeluge/sounds/417827/ isn't an awful sub, at 1.5s vs the original 2s, and is license free

emptyrivers (changes requested):
> Backwards compatibility would be nice, is there an already existing function that we can use to redirect to another media file, or a replacement file somebody could think of?

Since the breakage that this change produces is at the data level, you will need to add a migration case [here](https://github.com/WeakAuras/WeakAuras2/blob/ebd49a6ca7225b52e2f04e12176a1b27e226b115/WeakAuras/Modernize.lua#L5-L979) to fix invalidated data. Additionally in the current state, the addon code is also broken - you will need to remove the [registration to LibSharedMedia](https://github.com/WeakAuras/WeakAuras2/blob/f7ac1cb2c855b9656e1f93910ce5909947d075b4/WeakAuras/Types.lua#2385) of the deleted file (unless said file is replaced, that is).

InfusOnWoW:
I don't think we should replace nor delete the sound.

The Code of Conduct is concerned with "fostering an open and welcoming environment", which I doubt is affected by us shipping that sound. 

We certainly wouldn't add the sound today. But breaking existing auras is something that we don't do intentionally. And in this case I don't find the sound so inappropriate, that we should make a exception from that.

SpencerMalone:
I personally found it to be revolting and unappealing to have that shipped with the addon, and makes it feel unwelcoming to keep around. I think the fact that...
> We certainly wouldn't add the sound today

means we should probably replace it?

I do agree that replacing the sound is probably a better user experience than wholly removing it, as you still have a unique sound and don't have to update anything.

InfusOnWoW:
No, as I already explained, removing things breaks user's auras. Which is not something we do intentionally. We even have code, whose only purpose is to work around custom auras that use undocumented hacks.

There might be a cultural misunderstanding or some kind of language problem. I find the sound inappropriate and unprofessional. But crucially I fail to see how it is offensive.

PR #3051, opened 2021-04-26
Add Ready for Update section + fixes to Install section

mrbuds:
# Description

Add a section on top with all auras that can be updated called "Ready for Update"
The buttons for this section are not directly linked to target's aura, if there is multiple target you can choose with right click menu which aura to update.

A goal of this PR is to regroup all update functions to the top of the UI, to achieve this **all wago update related options are removed from normal auras buttons and right click menu**.
Ignoring/un-ignoring aura updates was set per aura, this pr completely remove the ability to do that in WeakAuras UI, Companion app and its clones can implement it on their side

image

image

## Type of change
- [x] Bug fix
- [x] New feature
- [x] Loss of functionality

## Checklist

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -418,6 +418,26 @@ StaticPopupDialogs["WEAKAURAS_CONFIRM_DELETE"] = {
   preferredindex = STATICPOPUP_NUMDIALOGS,
 }
 
+StaticPopupDialogs["WEAKAURAS_CONFIRM_IGNORE_UPDATES"] = {
+  text = L["Do you want to ignore all future updates for this aura"],
+  button1 = L["Yes"],
+  button2 = L["Cancel"],
+  OnAccept = function(self)
+    if self.data then
+      local auraData = WeakAuras.GetData(self.data)
+      if auraData then
+        for child in OptionsPrivate.Private.TraverseAll(auraData) do
+          child.ignoreWagoUpdate = true
+        end
+      end
+      WeakAuras.SortDisplayButtons(nil, true)
+    end
+  end,
+  OnCancel = function(self) end,
+  whileDead = true,
+  preferredindex = STATICPOPUP_NUMDIALOGS,
+}
+
 function OptionsPrivate.ConfirmDelete(toDelete, parents)
   if toDelete then
     local warningForm = L["You are about to delete %d aura(s). |cFFFF0000This cannot be undone!|r Would you like to continue?"]
@@ -625,8 +645,9 @@ local function LayoutDisplayButtons(msg)
   --if(frame.addonsButton) then
   --  frame.buttonsScroll:AddChild(frame.addonsButton);
   --end
-  if WeakAurasCompanion and WeakAurasCompanion.stash and next(WeakAurasCompanion.stash) then
-    frame.buttonsScroll:AddChild(frame.pendingImportButton);
+  if WeakAurasCompanion then
+    frame.buttonsScroll:AddChild(frame.pendingInstallButton);
+    frame.buttonsScroll:AddChild(frame.pendingUpdateButton);
   end
   frame.buttonsScroll:AddChild(frame.loadedButton);
   frame.buttonsScroll:AddChild(frame.unloadedButton);
@@ -663,10 +684,6 @@ local function LayoutDisplayButtons(msg)
       for id, button in pairs(displayButtons) do
         if(OptionsPrivate.Private.loaded[id] ~= nil) then
           button:PriorityShow(1);
-        end
-        if WeakAurasCompanion and not button.data.parent then
-          -- initialize update icons on top level buttons
-          button:RefreshUpdate()
         end
       end
     end
@@ -874,6 +891,7 @@ end
 
 local previousFilter;
 local pendingUpdateButtons = {}
+local pendingInstallButtons = {}
 function WeakAuras.SortDisplayButtons(filter, overrideReset, id)
   if (OptionsPrivate.Private.IsOptionsProcessingPaused()) then
     return;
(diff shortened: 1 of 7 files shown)


WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -874,6 +875,7 @@ end
 
 local previousFilter;
 local pendingUpdateButtons = {}
```
Stanzilla, line 877: left over from copy pasta?

Stanzilla:
I'm not sure if I like it also being green

mrbuds:
> 
> 
> I'm not sure if I like it also being green

i agree

mrbuds:
> 
> 
> I'm not sure if I like it also being green

what about blue ?
image

Stanzilla:
I think I like yellow, in things like git it is always green = new/added, yellow=changed

StanzillaManticore:
I like it, you?

Stanzilla:
I like it, you?

mrbuds:
That yellow color is good enough.

I'd like to mention that have not implemented any other action on the button other than the "click to update" logo button, there is no right click menu for them.

Previous behavior on auras with an update is still the same, if you want to skip/ignore updates you have to right click on them, not on the button of "ready for update" section.

There is room for improvement, it feels convoluted right now, i would like to remove the update buttons on each auras and keep all operations related to update to that new "Ready for Update" section. I'd like to work on it on next iteration.

Stanzilla:
That makes sense yes, so you think this can go in already?

mrbuds:
Yes, or if it's postpone i'd like to cherry-pick <https://github.com/WeakAuras/WeakAuras2/pull/3051/commits/6548fe73baf7343c692f876108b5a11a500b4f7e> that fix an error with outdated version of companion, and an error with already merged "Ready for Import"

Stanzilla:
Let's do the cherry pick and flesh the experience out in this PR

Stanzilla:
One bug atm is that when you push an aura into companion that is already installed, it will show up as ready to import, since we do have the wago versions saved, could we prevent them from showing up or do you rather see that as the duty of the app pushing it to check?

Also I think "Ready to Import" could be "Ready to Install" now that we have update as well, since technically, updating is also an import that might be a bit weird?

mrbuds:
> One bug atm is that when you push an aura into companion that is already installed, it will show up as ready to import, since we do have the wago versions saved, could we prevent them from showing up or do you rather see that as the duty of the app pushing it to check?

I'm not sure yet, i can find use case where i would want an aura always show in "ready for import" after clicking the "install in companion" button on wago even if i already have it installed


> Also I think "Ready to Import" could be "Ready to Install" now that we have update as well, since technically, updating is also an import that might be a bit weird?

yes

Stanzilla:
1. Install old version of aura from Wago
2. Run companion so it offers an update in game
3. Push different aura from wago to companion
4. Error

```
1x WeakAurasOptions\WeakAurasOptions-@project-version@.lua:913: attempt to index local 'child' (a nil value)
[string "@WeakAurasOptions\WeakAurasOptions.lua"]:913: in function `SortDisplayButtons'
[string "@WeakAurasOptions\WeakAurasOptions.lua"]:660: in function <WeakAurasOptions\WeakAurasOptions.lua:635>

```

Stanzilla:
When you install an aura from "Ready to Install" can we then hide it in that section? Also we might want a remove button as well. Could be in the right click menu of the the round import button

mrbuds:
> When you install an aura from "Ready to Install" can we then hide it in that section? Also we might want a remove button as well. Could be in the right click menu of the the round import button

The life cycle of *ready to installed* auras is handled by companion, when you click on the "ok" button of notification it get's removed from companion data

Stanzilla:
Hrm that's a bit weird

mrbuds:
If you let WeakAuras handle when an import string it adds some complications
- when will companion stop delivering the data ? should it keep it forever ?
- if i tag an import string to not be shown, so how can i untag it to be shown again if i want to push a new install of this aura ?

import strings are not tied 1 for 1 with auras, you can have duplicate auras / multiple versions of an aura

mrbuds:
I don't have more change in mind for this PR

AcidWeb:
There is any ETA for rollout of new companion and this new install-aura-from-page feature?

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -892,38 +890,102 @@ function WeakAuras.SortDisplayButtons(filter, overrideReset, id)
...
-  if not pendingImportButtonShown and frame.pendingImportButton then
-    frame.pendingImportButton.frame:Hide()
+  if not pendingInstallButtonShown and frame.pendingInstallButton then
+    frame.pendingInstallButton.frame:Hide()
+  end
+
+  local pendingUpdateButtonShown = false
+  if CompanionData then
+    local handled = {}
+    for id, aura in pairs(WeakAurasSaved.displays) do
```
InfusOnWoW, line 932: Doing this just on initialization does miss two things:

New auras that are imported that match a existing pending import. (Frankly not that likely)
Auras being deleted which match a pending import. (I haven't tested what happens in that case.)

InfusOnWoW (commented):
You are the maintainer, so you got to decide what to do with the existing ignore update functionality.

But I think a reasonable function would be to have a "ignore update" button on the pending update button, that sets the ignore flag on all auras that are linked.

For the common case that there is a 1:1 mapping between pending updates and existing aura, that seems to work just like users expect.

And for the various cases where a update is for multiple auras, that seems to be reasonable too.

E.g. it fails if a user ignores a update but then later imports a aura that matches the update, but that's rather slim edge-case.

mrbuds:
> 
> 
> You are the maintainer, so you got to decide what to do with the existing ignore update functionality.
> 
> But I think a reasonable function would be to have a "ignore update" button on the pending update button, that sets the ignore flag on all auras that are linked.
> 
> For the common case that there is a 1:1 mapping between pending updates and existing aura, that seems to work just like users expect.
> 
> And for the various cases where a update is for multiple auras, that seems to be reasonable too.
> 
> E.g. it fails if a user ignores a update but then later imports a aura that matches the update, but that's rather slim edge-case.

I re-added a ignore update option to right click menu, and all your comments should be fixed

InfusOnWoW:
Can you squash all the pending updates/install commits?

It looks fine now.

mrbuds:
> 
> 
> Can you squash all the pending updates/install commits?
> 
> It looks fine now.

done [space] !

Issue #1881, opened 2020-01-03
Publish auras natively from WeakAuras


emptyrivers:
**Is your feature request related to a problem? Please describe.**
We have a few fields in aura data which don't really *do* anything, and just exist as meta data for the user to know more information about who made the aura, where it came from, what version it is, etc. We don't really provide access to set any of these values, which leads to some questions sometimes like, "how can i set the url of the aura?". The answer is to export to https://wago.io, and then reimport from wago. But that's not a great answer, in my opinion.

**Describe the solution you'd like**
Provide an option on exporting an aura to publish it instead (or some variation on that wording). Publishing an aura would pop up some interface that would let you set various pieces of information like description, version, URL, contact info, etc on your local data before preparing an export string.

The exact flow of how to get to the publish interface is not something I've settled on.

**Describe alternatives you've considered**
- Perhaps we could add another tab to the right hand pane with various kinds of information. But certain people like to complain that we have too many tabs already, so I don't expect this alternative to go down well.
- Provide some kind of debug interface where users who know what they're doing would edit the data directly. This might actually be useful for other reasons too (so long as the option is not easily accessible by people who don't know what they're doing).

emptyrivers:
@oratory if this does end up getting implemented, consider pulling from the imported data updated meta data to try and prefill version notes.

InfusOnWoW:
I think there some overlap with <https://github.com/WeakAuras/WeakAuras2/issues/1512>.

And yes we probably should have some way to set/clear the meta data of a aura.

emptyrivers:
Ah, yes good point. I had forgotten about that ticket. Maybe if i keep writing tickets that dance around the missing pieces I'll eventually figure out the perfect solution.

InfusOnWoW:
I think having the meta data visible (and having some explanatory text around the uid on it, would help for #1512, though it might not be enough.

I think we adding a tab is the solution that we should try first.

jec324:
I believe we should also consider letting the original author "lock" the meta data of the original to a degree.  Such as, prohibiting the URL or Name from being edited on publish if they choose, perhaps having an "original Author" field.  This could be done in a Custom Option or something, but I'm not sold on that particular solution.  

[...]

PR #3056, opened 2021-04-28
Add option to modify aura version

Vardex:
# Description

This add an option in the information tab, to modify the version that is shown on the tooltip of an aura.

## Type of change

## How Has This Been Tested

- [x] Changed version of a single aura
- [x] Changed version of a group

## Checklist

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -1096,7 +1096,7 @@ local methods = {
 
     local hasDescription = data.desc and data.desc ~= "";
     local hasUrl = data.url and data.url ~= "";
-    local hasVersion = (data.semver and data.semver ~= "") or (data.version and data.version ~= "");
+    local hasVersion = (data.semver and data.semver ~= "") ;
 
     if(hasDescription or hasUrl or hasVersion) then
       tinsert(namestable, " ");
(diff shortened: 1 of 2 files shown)


mrbuds:
I don't get what it solve, why not leave it to wago ?

Vardex:
Don't really know but recently, description and url modification was added. Version is the only things that the user see on their aura and that they can't change.

I only made that PR because someone asked how to change that version and I had to tell them to edit their sv.

InfusOnWoW:
Well, that's intentionally, it's data that is maintained on wago's side.

Vardex:
You could say the same for url, it was something that was added by wago and couldn't be modified in game.

InfusOnWoW:
You are right. I relented on url because too many people wanted to change the url. I guess the same rational is true for version too.

What was the difference between semversion and version? semversion is what the user sets and version just the incremented number?

mrbuds:
> What was the difference between semversion and version? semversion is what the user sets and version just the incremented number?

companion & weakauras use "semver" for display/information for user only
"version" is use by both to compare which aura for same "slug" is the newest

emptyrivers:
Related #1881

InfusOnWoW:
This needs:

* Coordination with @oratory so that if a export with a semVer is pasted into the "Import new version" dialog, the semver from the import is used. (If valid)
* A validation in WeakAuras to ensure that the format stays a dotted 3 number list or whatever the right format is.

oratory:
> * Coordination with @oratory so that if a export with a semVer is pasted into the "Import new version" dialog, the semver from the import is used. (If valid)

I can add something to wago to look for a manually changed version.

> * A validation in WeakAuras to ensure that the format stays a dotted 3 number list or whatever the right format is.

It should also check that the new version is greater than the current version, or wago won't use it.

InfusOnWoW:
Nah, we don't want a "increase" check in wa. As one of the use cases is to "reset" the version to 1.0.0.

oratory:
Ah that makes sense, Wago would just ignore it if it's not an increase then.

InfusOnWoW:
Closing for now, since it wasn't update in a long time.

PR #3081, opened 2021-05-12
Support "Action Queued" for classic/tbc Templates

### LINKED ISSUE #2995: Missing icon
## Describe the bug
I was making simple icons for my spells and can't find Heroic Strike anywhere. It should be next to Hamstring, but is simply missing in the list.

There is no error message.

### To Reproduce

1. Go to 'New Aura'
2. Click on 'From Template'
3. Click on 'Icon'. Any size will do
4. Click on 'Abilities'
5. See 'Heroic Strike' missing from the list.

**Screenshots**
NoHS

No, but this seems unlikely to be the issue.

WeakAuras-3.2.2-classic

- [X] Classic
- [ ] Retail

Don't know

mrbuds:
# Description

Fixes #2995 & an error in toc file

### DIFF
--- a/WeakAurasTemplates/TriggerTemplates.lua
+++ b/WeakAurasTemplates/TriggerTemplates.lua
@@ -133,6 +133,10 @@ local checks = {
   enchantMissing = {
     variable = "enchanted",
     value = 0
+  },
+  queued = {
+    variable = "show",
+    value = 1,
   }
 }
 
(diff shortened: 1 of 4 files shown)


WeakAurasTemplates/TriggerTemplates.lua
```diff
@@ -404,6 +411,22 @@ local function createWeaponEnchantTrigger(triggers, position, item, showOn)
   }
 end
 
+local function createQueuedActionTrigger(triggers, position, item)
+  triggers[position] = {
+    trigger = {
+      type = WeakAuras.GetTriggerCategoryFor("Queued Action"),
+      event = "Queued Action",
+      use_enchant = true,
```
InfusOnWoW, line 419: That looks unnecessary?

mrbuds: removed

Issue #3088, opened 2021-05-14
BC - Steady Shot/Aimed Shot resetting melee swing timer


samhalli:
### Description

Steady Shot and Aimed Shot are resetting the melee swing timer when using trigger Player/Unit Info - Swing Timer - Main Hand

Also Aimed Shot is -not- resetting the ranged swing timer when using trigger Player/Unit Info - Swing Timer - Ranged

### WeakAuras Version

bc-alpha-5

### World of Warcraft Flavor

The Burning Crusade

### Tested with only WeakAuras

- [X] Yes
- [ ] No

### Reproduction Steps

Create progress bar aura with trigger of Player/Unit Info - Swing Timer - Main Hand, and cast stead/aimed shot

### Export String

[...]

PR #3089, opened 2021-05-14
fix scrolling with Ready for Install/Update

mrbuds:
AddChild should only be done on init
shown buttons always need to be added to frame.buttonsScroll.children


InfusOnWoW (approved):

mrbuds:
added a fix for #3088

InfusOnWoW (approved):

PR #3090, opened 2021-05-14
Development proxifier

TimoshN:
# Description
Mostly copy paste of Interface/FrameXML/RestrictedInfrastructure.lua. Attepts to control frame creation and buttons's Click methods

## Type of change

## How Has This Been Tested

Daily use. After some time custom code may throw nil errors

## Checklist

### DIFF
--- a/WeakAuras/AuraEnvironment.lua
+++ b/WeakAuras/AuraEnvironment.lua
@@ -556,6 +556,25 @@ do
     if doOutput then print(...) end
   end
 
+  local function modificateArg(arg)
+    if ( not arg ) then return end
+    return proxifier(arg)
+  end
+
+  local function nextArgs(arg, ...)
+    local numPayload = select('#', ...)
+    if numPayload == 0 and not arg then return end
+    return modificateArg(arg), nextArgs(...)
+  end
+
+  local function captureReturn(success, result, ...)
+    if not success then
+      error(result)
+    end
+
+    return modificateArg(result), nextArgs(...)
+  end
+
   local function proxifier_proxy_table(var)
     if not proxifierCache[var] then
       proxifierCache[var] = setmetatable({}, {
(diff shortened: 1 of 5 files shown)


caffeineaddiction:
```lua
local _GG = {}
for k,v in pairs(_G) do
  _GG[k] = v
end
-- _G ~= _GG
```

caffeineaddiction:
```lua
metaProxy = function(var)
  if type(var) == 'function' then
    return function(...) return var(...) end end
  if type(var) == 'table' then
    return setmetatable({}, { __index = function(self,k) return metaProxy(var[k]) end})
  end
  return var
end

local _GG = metaProxy(_G)
-- _G ~= _GG
-- _G.print ~= _GG.print
```

Stanzilla:
@caffeineaddiction what are you trying to say?

caffeineaddiction:
> @caffeineaddiction what are you trying to say?

"Attepts to control frame creation and buttons's Click methods"

an attempt was made?

Stanzilla:
> > @caffeineaddiction what are you trying to say?
> 
> "Attepts to control frame creation and buttons's Click methods"
> 
> an attempt was made?

Okay let's try full sentences next time. So you are just posting ways to circumvent the mechanism that was added?

caffeineaddiction:
I am saying I don't see a value in adding this complexity when it can be bypassed so easily.

https://www.google.com/search?q=security+through+obscurity+is+not+security

TimoshN:
Thanks for pointing, very appreciated!

InfusOnWoW:
@caffeineaddiction you need to seriously consider how you treat others. Linking that google search is clearly not appropriate and that kind of conduct is not acceptable.

ResultsMayVary:
While caffeine's comments perhaps came across curtly I think his point still has value. What precisely does this try to protect and how does it go about it.

WeakAuras/AuraEnvironment.lua
```diff
@@ -435,7 +758,7 @@ local exec_env = setmetatable({},
     elseif overridden[k] then
       return overridden[k]
     else
-      return _G[k]
+      return _G[k] -- proxifier(_G[k])
```
ResultsMayVary, line 761: Right now this proxifier does nothing.

TimoshN: Because it is not needed. Virtual regions is a thing

PR #3100, opened 2021-05-18
Multitoc for BCC & Classic

mrbuds:
# Description

Package Classic and BCC libs & specific files for both packages, and let the new multi toc feature handle which files to load.

The goal is to have a single package work for both "classic era" and BCC clients, so if we have too much issues with people using wrong version of the addon we can use this.

For reference: https://github.com/Stanzilla/WoWUIBugs/issues/68


Stanzilla:
So if we want to move to a single package release, to avoid people downloading the wrong one, we would remove the extra `.pkgmeta` files and move everything back into one. Then also rename the default toc and add `-Mainline` to it.

I don't quite see the benefit, apart from the user issue, though.

mrbuds:
A benefit for me is i will be able to use  my local git repo directly on bcc, currently i use alpha builds from curseforge (packaged), do my changes and copy the changes to my local copy of the repo

InfusOnWoW:
Yeah it would be nice to have that work.

Stanzilla:
Going to look into it more, thanks guys!

mrbuds:
> 
> 
> So if we want to move to a single package release, to avoid people downloading the wrong one, we would remove the extra `.pkgmeta` files and move everything back into one. Then also rename the default toc and add `-Mainline` to it.
> 
> I don't quite see the benefit, apart from the user issue, though.

updated with this change

mrbuds:
BW Packager can't find toc file

`Could not find an addon TOC file. In another directory? Make sure it matches the 'package-as' in .pkgmeta`

Stanzilla:
the proper multitoc stuff is not in the main branch yet either

mrbuds:
for reference https://github.com/BigWigsMods/packager/tree/multitoc#whats-new-with-v2

Stanzilla:
Is there any mode you prefer? I would be fine with what we currently have but you mentioned you want split tocs because of easier testing in classic?

mrbuds:
Using multiple toc files indeed make it easier to dev weakauras on multiple clients.
Also if we choose to use a single package for all 3 versions it will definitely solve the long time issue we had with curseforge distributing wrong version for a client, but that imply increase a bit the size of the package for all versions as retail version would include tbc & classic specific files & libs.

Stanzilla:
Hrm, to make testing easier, you could create the toc files locally and add them to .gitignore, no?

mrbuds:
Mhmm true indeed i didn't think about it, but i'd have to copy them again when i checkout an other branch/repo

Stanzilla:
I thought that would leave ignored files alone but I haven't checked in a while

InfusOnWoW:
I'd like to have multi-toc in, it would make my life easier too.

Stanzilla:
`WeakAurasModelPaths-Mainline.toc` needs to be updated to remove the packager keywords as well

Stanzilla:
Also looks like we might need an empty `WeakAuras.toc` because CurseForge requires that this file exists. Not sure if it will accept the empty file or if we have to fill it but let's test.

mrbuds:
After renaming Mainline files it produce 2 artifacts instead of 3

mrbuds:
Looks like -Classic.toc files are not handled

Stanzilla:
Looks good to me?

mrbuds:
Right, we'll see what are the best practice to use when multitoc branch is merged and is stable

mrbuds:
Moved to #3722

PR #3139, opened 2021-06-08
WeakAuras: Add a stage trigger for BigWigs

Justw8:
# Description

This change adds a bigwigs stage trigger which can be used for weak auras.

## Type of change

## How Has This Been Tested

- [x] Changed the stages in bigwigs using the callback, and it then shows/hides the WA depending on the stage settings set in the trigger.

## Checklist


WeakAuras/Prototypes.lua
```diff
@@ -4768,6 +4768,31 @@ Private.event_prototypes = {
...
+    init = function(trigger)
+      WeakAuras.RegisterBigWigsCallback("BigWigs_SetStage");
+      return [[
+         local stage = WeakAuras.GetBigWigsStage()
+      ]]
+    end,
+    args = {
+      {
+        name = "stage",
+        init = "arg",
```
InfusOnWoW, line 4787: It's "better" to use init = "WeakAuras.GetBigWigsStage()" here.

WeakAuras/Prototypes.lua
```diff
@@ -4768,6 +4768,31 @@ Private.event_prototypes = {
...
+        display = L["Stage"],
+        type = "number",
+        conditionType = "number",
```
InfusOnWoW, line 4790: Has this been tested? I think a store = true is missing to make conditions work.

WeakAuras/GenericTrigger.lua
```diff
@@ -2964,6 +2965,10 @@ do
           WeakAuras.ScanEvents("BigWigs_StopBar", id)
         end
       end
+    elseif event == "BigWigs_SetStage" then
+      local addon, stage = ...
+      currentStage = stage
+      WeakAuras.ScanEvents("BigWigs_SetStage", ...)
```
InfusOnWoW, line 2971: Is this event also send on encounter end/reset? (I would assume not, and then it probably requires adjusting the code in the BigWigs_OnBossDisable above?

Justw8: Yea, I discussed this with Funkeh and we probablly will reset the stage number on boss resets/wins. We don't do this rightnow.

WeakAuras/Prototypes.lua
Justw8: Yea, right now it seems to work fine. see: https://streamable.com/ui1mwu
I can try it with and without store = true and see what happens.

WeakAuras/GenericTrigger.lua
Justw8: to add: we do always set it to 1 on loading a boss module/engaging right now.

WeakAuras/Prototypes.lua
Justw8: I didn't see a difference in store with or without.

Justw8:
You can test it with this: 
https://gist.github.com/Justw8/730270f176791d32f44b5f6144d85f84

InfusOnWoW:
Thanks

PR #3151, opened 2021-06-13
Smallpatches

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -7729,12 +7729,20 @@ Private.event_prototypes = {
...
         name = "movespeedpercent",
-        display = L["Movement Speed (%)"],
+        display = L["Current Movement Speed (%)"],
         type = "number",
         init = "GetUnitSpeed('player') / 7 * 100",
         store = true,
         conditionType = "number"
       },
+      {
+        name = "runpeedpercent",
```
nullKomplex, line 7739: Unfortunate typo.

nullKomplex (changes requested):

WeakAuras/Prototypes.lua
InfusOnWoW: Ah fixed, thanks.

WeakAuras/BuffTrigger2.lua
```diff
@@ -967,6 +967,14 @@ local function TriggerInfoApplies(triggerInfo, unit)
     return false
   end
 
+  if triggerInfo.arenaSpec and unit:sub(1, 5) == "arena" then
+    -- GetArenaOpponentSpec doesn't use unit ids!
+    local i = tonumber(unit:sub(6))
+    if not triggerInfo.arenaSpec[GetArenaOpponentSpec(i)] then
```
mrbuds, line 973: GetArenaOpponentSpec is nil on bcc & classic

InfusOnWoW: That code path can never be encountered on classic/bcc

mrbuds: Are you sure it can't be encounter if you import an aura with that option enable from Retail to BCC ?

InfusOnWoW: Yes it checks if the unit id matches arena two lines above, and that unit id comes from the game.

PR #3175, opened 2021-06-30
Add DBM Stage trigger

mrbuds:
# Description

Support new DBM_SetStage callback

`fireEvent("DBM_SetStage", self, self.id, self.vb.phase, encounterId)--Mod, modId, Stage, Encounter Id (if available).`

## Type of change

## How Has This Been Tested

I didn't do any test, @MysticalOS can you check if it's working ?

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -2624,6 +2624,7 @@ do
   local bars = {}
   local nextExpire -- time of next expiring timer
   local recheckTimer -- handle of timer
+  local currentStage = 0
 
   local function dbmRecheckTimers()
     local now = GetTime()
@@ -2743,6 +2744,10 @@ do
         end
       end
       WeakAuras.ScanEvents("DBM_TimerUpdate", id)
+    elseif event == "DBM_SetStage" then
+      local mod, modId, stage, encounterId = ...
+      currentStage = stage
+      WeakAuras.ScanEvents("DBM_SetStage", ...)
     else -- DBM_Announce
       WeakAuras.ScanEvents(event, ...)
     end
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
Is there some easy way to test it?

MysticalOS:
Don't forget you can add a fallback to the WeakAuras.GetDBMStage(), in event lets say a user reloaded mid combat or something and your local cache of it is gone, niche case I know but one dbm considers since within 3 seconds of a reconnect it resyncs all fight variables to make mod self aware again of current timers and stage etc.

Then weak auras can get that back to via
DBM:GetStage(modId) which will return stage number of a specific mod or use DBM:GetStage() which acts as a catchall and just returns stage of first mod from the incombat table if any bosses are engaged. if no fight in progress or modid is invalid it'd just be nil

mrbuds:
> Don't forget you can add a fallback to the WeakAuras.GetDBMStage(), in event lets say a user reloaded mid combat or something and your local cache of it is gone, niche case I know but one dbm considers since within 3 seconds of a reconnect it resyncs all fight variables to make mod self aware again of current timers and stage etc.


In that case doesn't DBM fire DBM_SetStage event ?

MysticalOS:
It fires it only when event occurs, but on disconnect/reconnect it wouldn't fire callback until next time stage changes, since DBM just recovers those variables from another user and sets them locally only. I suppose locally I could maybe do something hacky in future to fire callback again after this happens, but at moment it doesn't. It's also a super niche case though since not many authors have even designed their mods around disconnect recovery.

InfusOnWoW:
Then obviously WeakAuras can't detect a stage after reconnect, since we can't guess how long that would take. It probably doesn't matter anyway.

InfusOnWoW (approved):

PR #3177, opened 2021-07-01
Validate Companion Data before showing it to the user as a potential update.

nullKomplex:
# Description

In rare situations it's possible for some of the data provided by the companion update to be invalid. In the usual scenario, this is a result of an old aura missing modern data used to display the thumbnail. This change validates and modernizes the data before attempting to use it.

Fixes an issue from Discord by Dradux.

## Type of change

## How Has This Been Tested

First I imported an older version of an aura, where the most recent version is missing modern data:

https://wago.io/AggramarTechnique/0.0.4

Note the missing `barColor` and `orientation` fields necessary for aurabars to display their thumbnail.

Second I closed the game and loaded up the companion for updates.

Third I opened up the game again, logged in, and then opened WA Options. (You may need to Expand the Ready for Updates section if it is not already)

Prior to the update, an error was thrown and WA Options was broken preventing use. After fix, it works as intended.

## Checklist

I did not traverse the children, as the only aura used by the update is the "base" aura.

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasPendingInstallButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasPendingInstallButton.lua
@@ -35,6 +35,7 @@ local methods = {
     self:SetTitle(self.companionData.name)
     self.update:SetScript("OnClick", self.callbacks.OnUpdateClick)
     local data = OptionsPrivate.Private.StringToTable(self.companionData.encoded, true)
+    WeakAuras.PreAdd(data.d)
     self.data = data.d
     self.frame:EnableKeyboard(false)
     self:Enable()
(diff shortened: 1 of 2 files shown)


mrbuds (approved):

nullKomplex:
It prevents proper use of the WA Options. It's not as fatal as WA Options errors typically are, but the user reported it was preventing them from working on auras. Notably, if the Ready for Update section is already expanded when WA Options are loaded, not all of the buttons are created, including the ones at the top for import and new aura.

The lines above and below my change decode the data from the Companion and store it for use later. This is different than the data used when it actually imports (seen [here](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasPendingUpdateButton.lua#L61)). This data is then used to form the thumbnail through [this when creating the button in the options](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAurasOptions/WeakAurasOptions.lua#L956). Acquiring a thumnail is defined [here](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAuras/WeakAuras.lua#L434) invoking the region's specific modifyThumbnail, in this case failing in [the aurabar's modifyThumbnail](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAurasOptions/RegionOptions/AuraBar.lua#L473) when it attempts to `find` a non-existent value, as the data did not contain it and wasn't validated beforehand.

I hope that explanation wasn't too over the top.

emptyrivers:
Hmm, my instinct then is to say that the companion addon should change how it's injecting data into WeakAuras. Massaging data multiple times is not ideal. But that is something that could be improved in the future.

emptyrivers (approved):

PR #3189, opened 2021-07-05
change party/raid/group units to include pets support

### LINKED ISSUE #3111: Smart group + Pets option
I created a WA that applies to a Smart Group, which works excellent. I wanted to apply that to track HoT's on raid members Pet's as well, but the Smart Group does not apply to pet frames.

I would like to have the option of applying a WA to both Smart Group + Pets, to track things such as HoT's on raid member's pets.

Reaching out to the WA Discord, the response I received was that it would require a custom trigger. While I am not a developer myself, I will see what I can do to achieve this result. 

**Additional context**
N/A

mrbuds:
# Description

This PR add an options to generic triggers and aura trigger with unit = party / raid / smart group to for including pets, or show only pets

Fixes #3111

## Done
- [x] Add "Include Pets" option to Health Trigger
- [x] Filter pets for generic triggers at event level
- [x] Add UNIT_*:grouppets:grouppetsonly meta filter for custom events
- [x] Add "Include Pets" option to Bufftrigger2

## To Do
- [ ] Add includePets option to all unit generic trigger (will do after pre-review)
- [ ] Make some pics for blog

## Type of change

## How Has This Been Tested
- [x] aura anchored to unit frames
- [x] dynamic group of health / power bars
- [x] test party/raid/group by switching hunter to own subgroup or an other subgroup

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -1665,9 +1665,12 @@ local function EventHandler(frame, event, arg1, arg2, ...)
       tinsert(unitsToRemove, "focus")
     end
   elseif event == "UNIT_PET" then
-    ScanGroupUnit(time, matchDataChanged, nil, arg1)
-    if not UnitExistsFixed(arg1) then
-      tinsert(unitsToRemove, arg1) -- not sure if his is working, needs tests
+    local pet = WeakAuras.unitToPetUnit[arg1]
+    if pet then
+      ScanGroupUnit(time, matchDataChanged, nil, pet)
+      if not UnitExistsFixed(pet) then
+        tinsert(unitsToRemove, pet)
+      end
     end
   elseif event == "NAME_PLATE_UNIT_ADDED" then
     nameplateExists[arg1] = true
@@ -1747,7 +1750,7 @@ frame:RegisterEvent("UNIT_FACTION")
 frame:RegisterEvent("UNIT_NAME_UPDATE")
 frame:RegisterEvent("UNIT_FLAGS")
 frame:RegisterEvent("PLAYER_FLAGS_CHANGED")
-frame:RegisterUnitEvent("UNIT_PET", "player")
+frame:RegisterEvent("UNIT_PET")
 if not WeakAuras.IsClassic() then
   frame:RegisterEvent("PLAYER_FOCUS_CHANGED")
   if WeakAuras.IsRetail() then
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -128,8 +128,14 @@ local function UnitIsVisibleFixed(unit)
   return unitVisible[unit]
 end
 
-local function UnitInSubgroupOrPlayer(unit)
-  return UnitInSubgroup(WeakAuras.petUnitToUnit[unit] or unit) or UnitIsUnit("player", unit)
+local function UnitInSubgroupOrPlayer(unit, includePets)
+  if includePets == nil then
+    return UnitInSubgroup(unit) or UnitIsUnit("player", unit)
+  elseif includePets == true then
+    return UnitInSubgroup(WeakAuras.petUnitToUnit[unit] or unit) or UnitIsUnit("player", unit) or UnitIsUnit("pet", unit)
+  elseif includePets == false then
+    return UnitInSubgroup(WeakAuras.petUnitToUnit[unit]) or UnitIsUnit("pet", unit)
+  end
 end
 
 local function GetOrCreateSubTable(base, next, ...)
@@ -1050,11 +1056,11 @@ local function TriggerInfoApplies(triggerInfo, unit)
   if triggerInfo.unit == "group" and triggerInfo.groupSubType == "party" then
     if IsInRaid() then
       -- Filter our player/party# while in raid and keep only raid units that are correct
-      if not Private.multiUnitUnits.raid[unit] or not UnitInSubgroupOrPlayer(unit) then
+      if not Private.multiUnitUnits.raid[unit] or not UnitInSubgroupOrPlayer(unit, triggerInfo.includePets) then
         return false
       end
     else
-      if not UnitInSubgroupOrPlayer(unit) then
+      if not UnitInSubgroupOrPlayer(unit, triggerInfo.includePets) then
         return false
       end
     end
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -129,7 +129,7 @@ local function UnitIsVisibleFixed(unit)
 end
 
 local function UnitInSubgroupOrPlayer(unit)
-  return UnitInSubgroup(unit) or UnitIsUnit("player", unit)
+  return UnitInSubgroup(WeakAuras.petUnitToUnit[unit] or unit) or UnitIsUnit("player", unit)
 end
 
 local function GetOrCreateSubTable(base, next, ...)
@@ -837,17 +837,34 @@ local function RemoveState(triggerStates, cloneId)
   end
 end
 
-local function GetAllUnits(unit, allUnits)
+local function GetAllUnits(unit, allUnits, includePets)
   if unit == "group" then
     if allUnits then
       local i = 1
       local raid = true
+      local pets = true
       return function()
         if raid then
           if i <= 40 then
-            local ret = WeakAuras.raidUnits[i]
-            i = i + 1
+            local ret
+            if includePets == true then -- pets + raid
+              ret = pets and WeakAuras.raidpetUnits[i] or WeakAuras.raidUnits[i]
+              pets = not pets
+              if pets then
+                i = i + 1
+              end
+            elseif includePets == false then -- pets only
+              ret = WeakAuras.raidpetUnits[i]
+              i = i + 1
+            else -- raid
+              ret = WeakAuras.raidUnits[i]
+              i = i + 1
+            end
             return ret
+          end
+          if includePets and pets then
+            pets = not pets
+            return "pet"
           end
           raid = false
           i = 1
@@ -855,8 +872,20 @@ local function GetAllUnits(unit, allUnits)
         end
 
         if i <= 4 then
-          local ret =  WeakAuras.partyUnits[i]
-          i = i + 1
+          local ret
+          if includePets == true then -- pets + group
+            ret = pets and WeakAuras.partypetUnits[i] or WeakAuras.partyUnits[i]
+            pets = not pets
+            if pets then
+              i = i + 1
+            end
+          elseif includePets == false then -- pets only
+            ret = WeakAuras.partypetUnits[i]
+            i = i + 1
+          else -- group
+            ret = WeakAuras.partyUnits[i]
+            i = i + 1
+          end
           return ret
         end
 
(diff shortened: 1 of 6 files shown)


emptyrivers:
Hmm, smart groups are common afaik. I feel like automagically enabling this for everyone might risk a flurry of "bug" reports.

Causese:
Don't like this, there are bunch of auras that check debuff missing and list the player's names. 
Now all of those will show pet names?
Also raidbuff WA's will always remain active since pet never gets the buff (assumingly)
I can also see lots of people hating themselves for forgetting/having to enable the pet filter every time

Nightwarden24:
> A button in the filters that makes the iteration includes pets might be a better idea

I agree with that.

InfusOnWoW:
Yep, by default it should not be enabled.

mrbuds:
Would having an "Exclude Pets" options + migration to enable it for old auras be ok ?

for generic triggers it would easy
```
      {
        name = "ignorePets",
        display = L["Ignore Pets"],
        type = "toggle",
        init = "not WeakAuras.UnitIsPet(unit)",
        enable = function(trigger)
          return trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party"
        end
      },
```

but i don't get how to reverse the logic into an "includePets" option that exclude them by default

InfusOnWoW:
I've looked through this, and I think fundamentally the functions need to be explicit on whether they should include pets or not. That is both MultiUnitLoop and GetAllUnits need a parameter on whether they should include pets. (Or maybe a tristate? Exclude Pets, Include Pets, Only Pets.)

For BT2, the right change is in TriggerInfoApply to check the pet's inclusivity against the unit. That's run a lot less than other things. I'm not 100% sure what needs to happen on UNIT_PET.

For generic trigger, the easiest check is probably to register the trigger only for the unit pet ids if it includes pets. I think that could work.

One good question is how the syntax for custom triggers registering for group units should be work, e.g. how should UNIT_HEALTH:group work. My first guess is that it shouldn't include pets, and that there should be UNIT_HEALTH:grouppets.

mrbuds:
Did some work on generic trigger, still only Health trigger support it for now
> Pets are filtered at event registration level
> excludePets is now includePets, need to make it support an other label when state is "false" (it currently show "Not include pets" when what it's doing is "Pets only (exclude players)"

Did some changes to bufftrigger2 but it's still wip

mrbuds:
Bufftrigger2 "include pets" option looks good now but need more tests

TODO:
- fix option in BT2 to be a tristate
- fix option in generic trigger to show proper text when it's false
- add the option to all unit generic triggers (after pre-review)

WeakAuras/BuffTrigger2.lua
```diff
@@ -837,26 +837,55 @@ local function RemoveState(triggerStates, cloneId)
   end
 end
 
-local function GetAllUnits(unit, allUnits)
+local function GetAllUnits(unit, allUnits, includePets)
```
InfusOnWoW, line 840: true, false, nil are 3 values, but we should avoid using it as a tristate. Even though we do use it in some places, those usages lead to rather poor code.

WeakAuras/BuffTrigger2.lua
```diff
@@ -1463,7 +1529,7 @@ end
 local function ScanAllGroup(time, matchDataChanged)
   -- We iterate over all raid/player unit ids here because ScanGroupUnit also
   -- handles the cases where a unit existence changes.
-  for unit in GetAllUnits("group") do
+  for unit in GetAllUnits("group", nil, true) do
```
InfusOnWoW, line 1532: So the comment does not match what the function actually does.

And it turns out that's becasue the function is no longer used, so it can just be deleted.

WeakAuras/GenericTrigger.lua
```diff
@@ -1062,12 +1097,15 @@ function GenericTrigger.LoadDisplays(toLoad, loadEvent, ...)
...
-                function (unit)
-                  if not (genericTriggerRegisteredUnitEvents[unit] and genericTriggerRegisteredUnitEvents[unit][event]) then
-                    unitEventsToRegister[unit] = unitEventsToRegister[unit] or {}
-                    unitEventsToRegister[unit][event] = true
+                function (u)
+                  if not (genericTriggerRegisteredUnitEvents[u] and genericTriggerRegisteredUnitEvents[u][event]) then
+                    unitEventsToRegister[u] = unitEventsToRegister[u] or {}
+                    unitEventsToRegister[u][event] = true
+                    if WeakAuras.UnitIsPet(u) then
+                      unitEventsToRegister[u].UNIT_PET = true
```
InfusOnWoW, line 1105: And here too.

WeakAuras/GenericTrigger.lua
```diff
@@ -1292,12 +1330,23 @@ function GenericTrigger.Add(data, region)
...
                   end
                 elseif trueEvent:match("^UNIT_") then
+                  local includePets
+                  if i == "grouppets" then
+                    includePets = true
+                    i = "group"
+                  elseif i == "grouppetsonly" then
+                    includePets = false
+                    i = "group"
+                  end
```
InfusOnWoW, line 1340: Right, that seems good.

WeakAuras/GenericTrigger.lua
```diff
@@ -1013,12 +1043,17 @@ function LoadEvent(id, triggernum, data)
...
-            loaded_unit_events[unit][event] = loaded_unit_events[unit][event] or {};
-            loaded_unit_events[unit][event][id] = loaded_unit_events[unit][event][id] or {}
-            loaded_unit_events[unit][event][id][triggernum] = data;
-          end, unit
+          function(u)
+            loaded_unit_events[u] = loaded_unit_events[u] or {};
+            loaded_unit_events[u][event] = loaded_unit_events[u][event] or {};
+            loaded_unit_events[u][event][id] = loaded_unit_events[u][event][id] or {}
+            loaded_unit_events[u][event][id][triggernum] = data;
+            if WeakAuras.UnitIsPet(u) then
```
InfusOnWoW, line 1051: Is this correct?

Don't we need to add UNIT_PET for the pet owner units?

That is, I think the check here should be reversed?

Also is inside the loop the right place for the logic? Shouldn't adding UNIT_PET be the responsible of the caller?

WeakAuras/Prototypes.lua
```diff
@@ -1703,7 +1703,7 @@ Private.load_prototype = {
   }
 };
 
-local function AddUnitChangeInternalEvents(triggerUnit, t)
+local function AddUnitChangeInternalEvents(triggerUnit, t, includePets)
```
InfusOnWoW, line 1706: Ah, what was the purpose of this code... I don't quite remember.

WeakAuras/BuffTrigger2.lua
InfusOnWoW: Actually this is due to the options code. Argh! Then it's probably okayish.

InfusOnWoW:
I think code wise this looks pretty good. I haven't actually checked how it looks in game. Maybe the "Include Pets" shouldn't be a tristate but a drop down? As it behaves different than all the other tristates?

WeakAuras/Prototypes.lua
mrbuds: It's an helper function to set which units WeakAuras.WatchUnitChange should look for and launch scanevents when unit changes faction, role, reaction, or guid.
So for pets, this would track if a pet is mind control or changed guid.

WeakAuras/GenericTrigger.lua
```diff
@@ -1009,17 +1039,28 @@ function LoadEvent(id, triggernum, data)
     end
   end
   if data.unit_events then
+    local includePets = data.trigger.use_includePets == true and data.trigger.includePets or nil
```
InfusOnWoW, line 1042: There are two issues here, both a bit subtle but:

1) Since  you only converted the Health trigger, someone could configure a Health trigger for "PetsOnly", and then switch to the Alternate Power trigger. The Alternate Power trigger doesn't have a includePets setting, but the PetsOnly option is still active.

Now, I think the "Alternate Power" trigger might not need the pets only setting.

2) This code is also run for custom triggers, but custom triggers don't have a data.trigger.use_includePets setting. The way this code works is that in GenericTrigger.Add the various settings are extracted and put into a general format. E.g. unit_events are parsed there too. And then LoadEvent + LoadDisplay work on that general format and don't have to care about which trigger it was.

I've fixed both issues in a commit, which I now need to test.

mrbuds: I was going to add the option to more generic trigger, was just waiting for review on the health one.

WeakAuras/GenericTrigger.lua
```diff
@@ -1292,13 +1337,29 @@ function GenericTrigger.Add(data, region)
...
+                    i = "group"
+                  end
                   MultiUnitLoop(
                     function(unit)
                       trigger_unit_events[unit] = trigger_unit_events[unit] or {}
                       tinsert(trigger_unit_events[unit], trueEvent)
                       isUnitEvent = true
-                    end, i
+                    end, i, includePets
                   )
```
InfusOnWoW: I have made several changes here.

a) The handling of multi-units is a bit confussing, and it turns out that GenericTrigger.Add should not expand it here. Instead it's expanded in LoadDisplays + LoadEvent do that. So the loop here was always unnecessary. (At least that's what I think)

b) I removed the special casing of UNIT_PET. Instead custom coders will need to add UNIT_PET:group to their event list, if they want to listen to that event. For internal triggers this works without UNIT_PET due to the UNIT_CHANGED internal events.

c) I added partypets, partypetsonly

d) I made the matching for grouppets etc, case insensitive, as that's what is done for matching group

e) I removed the warning that showed up for grouppets, grouppetsonly

WeakAuras/Types.lua
```diff
@@ -3058,7 +3063,11 @@ Private.multiUnitId = {
   ["boss"] = true,
   ["arena"] = true,
   ["group"] = true,
+  ["grouppets"] = true,
+  ["grouppetsonly"] = true,
   ["party"] = true,
+  ["partypets"] = true,
+  ["partypetsonly"] = true,
```
InfusOnWoW, line 3070: This prevents a warning showing that grouppets is not a valid unit

WeakAuras/GenericTrigger.lua
```diff
@@ -1337,29 +1330,18 @@ function GenericTrigger.Add(data, region)
                     hasParam = true
                   end
                 elseif trueEvent:match("^UNIT_") then
-                  local includePets
-                  if i == "grouppets" then
+                  isUnitEvent = true
+
+                  if string.lower(strsub(i, #i - 3)) == "pets" then
```
InfusOnWoW, line 1335: I added partypets, partypetsonly. And made the matching case insensitive.

WeakAuras/GenericTrigger.lua
```diff
@@ -1337,29 +1330,18 @@ function GenericTrigger.Add(data, region)
...
+                    i = strsub(i, 1, #i-4)
                     includePets = "PlayersAndPets"
-                    i = "group"
-                  elseif i == "grouppetsonly" then
+                  elseif string.lower(strsub(i, #i - 7)) == "petsonly" then
                     includePets = "PetsOnly"
-                    i = "group"
-                  end
-                  MultiUnitLoop(
```
InfusOnWoW, line 1295: It turns out that this expanding of multi units is not necessary at all. Since it will be done later by LoadDisplay/LoadEvent

WeakAuras/GenericTrigger.lua
```diff
@@ -1337,29 +1330,18 @@ function GenericTrigger.Add(data, region)
...
-                      trigger_unit_events[unit] = trigger_unit_events[unit] or {}
-                      tinsert(trigger_unit_events[unit], trueEvent)
-                      isUnitEvent = true
-                    end, i, includePets
-                  )
-                  if includePets ~= nil then
-                    MultiUnitLoop(
-                      function(unit)
-                        trigger_unit_events[unit] = trigger_unit_events[unit] or {}
-                        tinsert(trigger_unit_events[unit], "UNIT_PET")
```
InfusOnWoW: I did remove the special handling of UNIT_PET here. Custom coders will need to add UNIT_PET:group to their event list if they want that event. That sounds fine to me.

For internal triggers, we listen to our fake UNIT_CHANGED_* events, so that work without any problem

InfusOnWoW (approved):

WeakAuras/Prototypes.lua
```diff
@@ -2282,6 +2296,7 @@ Private.event_prototypes = {
   },
   ["Health"] = {
     type = "unit",
+    includePets = "true",
```
InfusOnWoW, line 2299: A few changes here, and below so that triggers that don't support pets are handled the same as before, even if there might be strange settings.

mrbuds:
Added the option to "Cast" and "Power" triggers, wasn't sure about the Unit Characteristics one.

Agree it's ready for merge

PR #3222, opened 2021-07-20
Swing Timer & Character stats triggers fixes

mrbuds:
# Description

Swing Timer & Character stats triggers fixes

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -1488,6 +1488,10 @@ do
     return 0, math.huge;
   end
 
+  local function swingTriggerUpdate()
+    WeakAuras.ScanEvents("SWING_TIMER_UPDATE")
+  end
+
   local function swingEnd(hand)
     if(hand == "main") then
       lastSwingMain, swingDurationMain, mainSwingOffset = nil, nil, nil;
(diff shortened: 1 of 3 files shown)


emptyrivers:
Hi I found a bizzare edge case that you need to immediately fix.

emptyrivers:
No, I will not be giving details!

WeakAuras/GenericTrigger.lua
```diff
@@ -1647,7 +1647,9 @@ do
       swingTimerFrame:RegisterEvent("COMBAT_LOG_EVENT_UNFILTERED");
       swingTimerFrame:RegisterEvent("PLAYER_ENTER_COMBAT");
       swingTimerFrame:RegisterEvent("PLAYER_LEAVE_COMBAT");
-      swingTimerFrame:RegisterUnitEvent("UNIT_ATTACK_SPEED", "player");
+      if not WeakAuras.IsBCC() then
```
nullKomplex, line 1683: But it's not in Classic _or_ Retail?

InfusOnWoW: No one complained while classic was current, so I doubt it. For retail we tested it.

mrbuds: It may apply to classic too, but i'd rather not make a change i won't test

WeakAuras/GenericTrigger.lua
```diff
@@ -1582,6 +1582,38 @@ do
...
+        swingDurationMain = mainSpeed
+        mainSwingOffset = 0
+        timer:CancelTimer(mainTimer)
+        mainTimer = timer:ScheduleTimerFixed(swingEnd, mainSpeed, "main")
+        startEvent = true
+      end
+      if lastSwingOff then
+        lastSwingOff = currentTime
+        swingDurationOff = offSpeed
+        mainSwingOffset = 0
```
InfusOnWoW, line 1601: Is setting mainSwingOffset to zero correct here?

WeakAuras/GenericTrigger.lua
```diff
@@ -1582,6 +1582,38 @@ do
...
+      local startEvent = false
+      if lastSwingMain then
+        lastSwingMain = currentTime
+        swingDurationMain = mainSpeed
+        mainSwingOffset = 0
+        timer:CancelTimer(mainTimer)
+        mainTimer = timer:ScheduleTimerFixed(swingEnd, mainSpeed, "main")
+        startEvent = true
+      end
+      if lastSwingOff then
```
InfusOnWoW, line 1598: I'm not sure whether you consider that a weapon swap could be from e.g. a two-hander to two weapons. Or from a normal weapon to a ranged weapon.

Don't you need to check for what kind of weapons are equipped?

mrbuds: From my tests switching from any weapon to another reset swing timer, e.g. having no offhand and equipping a shield reset main hand

WeakAuras/GenericTrigger.lua
mrbuds: oops

WeakAuras/GenericTrigger.lua
InfusOnWoW: Yeah, but e.g. if you switch from a 2 weapons to one weapons, you still seem to schedule a timer for the offhand?

mrbuds: if you had a 2H weapon lastSwingOff should be nil

InfusOnWoW: previously 2 weapons,
now one weapon

Stanzilla: What if I disconnect in a fight and log back in, will the swing timer still be accurate for the next swing or only for the one after it?

PR #3246, opened 2021-07-29
Add Raid Marker format to Formats %unit

Hekili:
# Description

Adds Raid Marker format to Formats %unit
Also fixes minor issue that color defaults to false when options are "class" vs. "none".

## Type of change

## How Has This Been Tested

- [x] Test A: Apply raid marker to a unit.  Create WeakAura for that unit (i.e., target).  Create text field that includes %unit; click Format Options, select Format: Formats %unit, check Raid Marker.  

image

- [x] Test B: Apply raid marker to a unit.  Create WeakAura for that unit (i.e., target).  Create text field that includes %unit; click Format Options, select Format: Formats %unit, check Raid Marker and check Abbreviate, set Max Char to preferred value (8), confirm text is cut to 8 characters.

image

- [x] Test C: Apply raid marker to a unit.  Create WeakAura for that unit (i.e., target).  Create text field that includes %unit; click Format Options, select Format: Formats %unit, check Raid Marker, select Color = Class, and check Abbreviate, set Max Char to preferred value (8), confirm text is cut to 8 characters and class color is preserved.

image

- [x] Test D: Realm Name

image

## Checklist

Defects:
- Does not auto-update mark if a target receives a mark after trigger occurs.  Updates next trigger update.  Infus is aware.
- Could be enhanced to select between appending raid mark or prepending raid mark.  Not critical.

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -425,7 +425,6 @@ Private.format_types = {
       local nameFunc
       local colorFunc
       local abbreviateFunc
-      local raidTargetFunc
 
       if color == "class" then
         colorFunc = function(unit, text)
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -408,16 +408,25 @@ Private.format_types = {
           return not get(symbol .. "_abbreviate")
         end
       })
+      addOption(symbol .. "_raid_marker", {
+        type = "toggle",
+        name = L["Raid Marker"],
+        width = WeakAuras.normalWidth,
+        hidden = hidden,
+      })
     end,
     CreateFormatter = function(symbol, get)
-      local color = get(symbol .. "_color", true)
+      local color = get(symbol .. "_color", "none")
       local realm = get(symbol .. "_realm_name", "never")
       local abbreviate = get(symbol .. "_abbreviate", false)
       local abbreviateMax = get(symbol .. "_abbreviate_max", 8)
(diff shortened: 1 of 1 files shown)


Hekili:
WeakAura for testing purposes:  https://wago.io/VdSQXUgTl

asaka-wa:
Does this update when the unit's mark changes? I don't see the event being tracked.

InfusOnWoW:
No it doesn't, I promised to implement that. And then realized deeper problems, that might change how we approach it.

asaka-wa:
When we help people achieve this with a text function I usually recommend just running it every frame to avoid the run around of trying to get it to run on the event. It'd be a nice feature. Would be used a lot.

InfusOnWoW:
That's what I suspect.

The deeper problem is that it's in a text function and not part of the state and makes it unavailable for sorting. And I kinda feel if people want to see the buffs with marks, they probably want them sorted.

InfusOnWoW:
After pondering this for a bit, I have decided that the best option atm is to add support for role/roleIcon/raid mark directly to the bufftrigger 2.

I'm not too happy about that, since it's only one of the many unit related triggers we have, but making it general is hard, especially if it should be in the state. And it really should be.

So https://github.com/WeakAuras/WeakAuras2/pull/3253 it is then. 

Thanks anyway! Without the PR, I'd probably not made that a priority.

PR #3251, opened 2021-08-01
Smallpatches

InfusOnWoW:


InfusOnWoW:
The fix for the enchant trigger needs a bit more work. According to https://wow.tools/dbc/?dbc=globalstrings&build=9.1.0.39617&locale=zhCN#page=1&search=ITEM_ENCHANT_TIME_LEFT, the chinese locale uses double width ().

InfusOnWoW:
Also according to https://wow.tools/dbc/?dbc=globalstrings&build=9.1.0.39617&locale=zhTW#page=1&search=ITEM_ENCHANT_TIME_LEFT the sapce between name and "(" needs to be optional.

PR #3253, opened 2021-08-08
Smallpatches

InfusOnWoW:


mrbuds (approved):
Tested on retail / bcc / classic_era, work as intended

PR #3255, opened 2021-08-10
Swingtimer

mrbuds:
# Description

Fix 2 issues with swing timer, probably add more issues

## Type of change

## How Has This Been Tested
- [x] hit a few level 36 crocodile with my level 35 paladin

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -3199,6 +3199,8 @@ Private.noreset_swing_spells = {
   --35474 Drums of Panic DO reset the swing timer, do not add
 }
 
+Private.swingtimer_snapshotted_aura = {}
+
 Private.item_weapon_types = {}
 
 local skippedWeaponTypes = {}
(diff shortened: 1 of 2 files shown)


WeakAuras/Types.lua
```diff
@@ -3199,6 +3199,8 @@ Private.noreset_swing_spells = {
   --35474 Drums of Panic DO reset the swing timer, do not add
 }
 
+Private.swingtimer_snapshoted_aura = {}
```
Stanzilla, line 3202: snapshotted with two t

InfusOnWoW:
The first commit is fine. The second one scares me, that's very ugly. My feeling is that it's too brittle, and Blizzard really ought to fix their api.

Also you can use GetPlayerAuraBySpellID to effectively check if a given buff is there. I think even with checking against all paladin buffs, that's faster than UnitAura.

mrbuds:
I reverted the 2nd commit then, kept it in history of the branch

PR #3262, opened 2021-08-15
Smallfix

mrbuds:


mrbuds:
I'm not sure if the alpha fix is the best way to do it

WeakAuras/GenericTrigger.lua
```diff
@@ -4028,19 +4028,13 @@ WeakAuras.GetItemSubClassInfo = function(i)
...
-    local spellCrit = GetSpellCritChance(2)
-    for i = 3, MAX_SPELL_SCHOOLS do
-      spellCrit = min(spellCrit, GetSpellCritChance(i))
-    end
-    return max(spellCrit, GetRangedCritChance(), GetCritChance())
+WeakAuras.GetCritChance = function()
+  -- Based on what the wow paper doll does
+  local spellCrit = 0
+  for i = 2, MAX_SPELL_SCHOOLS or 7 do -- MAX_SPELL_SCHOOLS is nil on classic_era
+    spellCrit = max(spellCrit, GetSpellCritChance(i))
```
InfusOnWoW, line 4035: The paerpdoll uses min here:

See:
https://github.com/Gethe/wow-ui-source/blob/fb934dc6558497b2a494898a33e0587a0e6bfe3d/FrameXML/PaperDollFrame.lua#L1014

I think the reasoning is that if there's some kind of "bonus crit chance" on a specific spell school, that gets ignored.

I have no idea whether that is even a thing, but I suggest keeping it that way.

InfusOnWoW: Also curse the reporter of that bug, because they did not state which version of the game they are using.

mrbuds: > I think the reasoning is that if there's some kind of "bonus crit chance" on a specific spell school, that gets ignored.

image

This is a screen of paperdoll for my mage with fire spec, i have 9.35% "base" spell crit, that's what the `min` give, and tooltip show detailed crit per school, i think since we do some magic returning the max crit of melee/ranged/spell it make sense to also return the highest crit of all mage school bceause it is the school you use in your spell rotation.

PR #3264, opened 2021-08-15
Change the transmission version for nested exports

InfusOnWoW:


WeakAuras/Transmission.lua
```diff
@@ -772,22 +771,24 @@ function Private.DisplayToString(id, forChat)
...
+            uids[child.uid] = true
           end
-          transmit.c[index] = CompressDisplay(childData);
+        else
+          child.uid = GenerateUniqueID()
+        end
+        transmit.c[index] = CompressDisplay(child);
+        index = index + 1
+        if child.controlledChildren then
+          transmit.v = 2000
```
emptyrivers, line 791: A "sensible" option might be 3000 or 4000, to mirror the WeakAuras major version. But it genuinely doesn't matter

emptyrivers (approved):

PR #3266, opened 2021-08-17
GetHitChance : fix potential nil error

mrbuds:


InfusOnWoW:
That can return nil?

mrbuds:
Yeah it's weird and i couldn't reproduce by running those lines manually, but after a dungeon on my mage which have an aura using this trigger, i had a Lua error for math operation with nil on bugsack.

Stanzilla:
> Yeah it's weird and i couldn't reproduce by running those lines manually, but after a dungeon on my mage which have an aura using this trigger, i had a Lua error for math operation with nil on bugsack.

On PTR or live?

mrbuds:
live

PR #3274, opened 2021-08-22
Deprecate some old sound files

InfusOnWoW:


InfusOnWoW:
Turns out pretty much every sound that I thought is almost unusable is actually used by many people.

PR #3305, opened 2021-09-21
Accept e.g. 30% for Time Remaining triggers.

jdashg:
# Description

Triggers now accept percentages (e.g. 50%) for the Time Remaining option. For a 10.0s duration aura, 30% would mean 3.0s.
In particular, this makes [pandemic window](https://blog.askmrrobot.com/how-wow-works-periodic-damage-and-healing-dots-and-hots/) tracking for dots much easier.

## Type of change

## How Has This Been Tested

- [x] I made a new Aura Icon, with a Trigger: Aura, Unit=Target, Aura Type=Debuff, Name(s)=Rip, Remaining Time=(Operator:<=, Remaining Time:50%)
- [x] As a regression test, I made sure Remaining Time:3 still worked.

## Checklist

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -408,12 +408,19 @@ local function UpdateMatchData(time, matchDataChanged, unit, index, filter, name
   return changed or data.lastChanged == time
 end
 
-local function calculateNextCheck(triggerInfoRemaining, auraDataRemaining, auraDataExpirationTime,  nextCheck)
+local function calculateNextCheck(triggerInfo, time, auraData, nextCheck)
+  local triggerInfoRemaining = triggerInfo.remainingCheck
+  local auraDataRemaining = auraData.expirationTime - time
+
+  if triggerInfo.remainingPercent then
+    triggerInfoRemaining = triggerInfoRemaining * auraData.duration
+  end
+
   if auraDataRemaining > 0 and auraDataRemaining >= triggerInfoRemaining then
     if not nextCheck then
-      return auraDataExpirationTime - triggerInfoRemaining
+      return auraData.expirationTime - triggerInfoRemaining
     else
-      return min(auraDataExpirationTime - triggerInfoRemaining, nextCheck)
+      return min(auraData.expirationTime - triggerInfoRemaining, nextCheck)
     end
   end
   return nextCheck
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
I'm afraid, that's a not something we should add. People will be tempted to use it for pandemid windows, which is just wrong.

jdashg:
Do you mean that this is not the best way to track pandemic windows, or that you think we should not encourage pandemic window tracking at all?

InfusOnWoW:
It's simply wrong. 30% of the duration is not the right threshold to use.

jdashg:
Can you describe in more detail why?

asaka-wa:
First application of a 24second debuff, duration is 24 secs
Refreshed early adds 30% of 24 (7.2) so new duration is 31.2secs
Refreshing this doesn't add 30% of the current duration (31.2 * 0.3 = 9.36) is still adds a max of 7.2. 

Using `0.3 * duration` is a common method for getting a pandemic value but it's only correct on the initial application. Lots of people are doing it wrong.

jdashg:
In that case, the pandemic threshold should always be between 30% of the duration (initial) and 0.3/1.3=23% (longest pandemic-extended duration), is that right? I can pandemic-extend every time so long as I reapply with 23% or less remaining then?

asaka-wa:
No. 
The pandemic value of any buff is 30% of the initial value. The current duration is never relevant, but the game doesn't provide an API that gives the initial duration.

nullKomplex:
It's 30% of the length of the aura you're applying, which is especially pertinent for something like a rogue or cat where the debuff or buff duration scale with the amount of combo points you have. That value will never be accurate when looking only at the currently applied aura.

InfusOnWoW:
And that's why this wouldn't be a good feature.

jdashg:
Ok! Thanks for the explanations! It sounds like while the 23% method would work for e.g. Rake (all applications have same duration) it cannot work effectively for e.g. Rip (many possible durations).

asaka-wa:
23% doesn't help. Why would you assume that a given buff was previously refreshed to a max possible total duration?

jdashg:
A duration is at-most 130% of the base duration, and at-least 100% of the base duration. 23% means that in both the 100% case, and the 130% case, you're definitely within the pandemic window. In the 100% case, yeah, you'd miss out on some of the window in that case, but maybe that's an acceptable approximation. (It's better than naively using 30% at least!)

PR #3312, opened 2021-09-26
Smallpatches

InfusOnWoW:


WeakAuras/GenericTrigger.lua
```diff
@@ -3665,6 +3665,19 @@ function GenericTrigger.GetAdditionalProperties(data, triggernum)
         ret = ret .. additional;
       end
     end
+  else
+    if (trigger.custom_type == "stateupdate") then
+      local variables = events[data.id][triggernum].tsuConditionVariables();
+      if (type(variables) == "table") then
+        for var, varData in pairs(variables) do
+          if (type(varData) == "table") then
+            if varData.description then
```
mrbuds, line 3674: Are you sure you want to add a new field ? I though it was going to use the previously supported  `display` field

WeakAuras/Prototypes.lua
```diff
@@ -5471,6 +5471,17 @@ Private.event_prototypes = {
       );
     end,
     args = {
+      {
+        name = "note",
+        type = "description",
+        display = "",
+        text = function()
+          if not WeakAuras.IsRetail() then
+            return L["Note: Due to how complicated the swing timer behaviour is and missing apis from Blizzard, the swing timer is inaccurate in edge cases."]
```
Stanzilla, line 5480: maybe "lack of APIs"

WeakAuras/Prototypes.lua
```diff
@@ -5382,6 +5471,17 @@ Private.event_prototypes = {
       );
     end,
     args = {
+      {
+        name = "note",
+        type = "description",
+        display = "",
+        text = function()
+          if not WeakAuras.IsRetail() then
+            return L["Note: Due to how complicated the swing timer behaviour is and lack of apis from Blizzard, the swing timer is inaccurate in edge cases."]
```
Stanzilla, line 5480:
```suggestion
            return L["Note: Due to how complicated the swing timer behaviour is and the lack of APIs from Blizzard, results are inaccurate in edge cases."]
```

Stanzilla:
Do you think it would be worth trying to repair the external addons feature if we advertise it more?

PR #3372, opened 2021-11-30
Smallpatches

InfusOnWoW:


WeakAuras/Transmission.lua
```diff
@@ -667,7 +667,7 @@ local receivedData;
 hooksecurefunc("SetItemRef", function(link, text)
   buttonAnchor:Hide()
   if(link == "garrmission:weakauras") then
-    local _, _, characterName, displayName = text:find("|Hgarrmission:weakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- ([^%]]+)%]|h");
+    local _, _, characterName, displayName = text:find("|Hgarrmission:weakauras|h|cFF8800FF%[([^%s]+) |r|cFF8800FF%- (.*)%]|h");
```
emptyrivers, line 670: Shouldn't this be a lazy match? .* would likely match all the way up through a second linked aura in the same chat message

PR #3381, opened 2021-12-06
add cutoff times

emptyrivers:
dunno why i never added this to the savedvariables but whatever


InfusOnWoW:
As discussed in the discord, this is a bit problematic because one of the unintended use cases for the archive is the restoration of auras after a data loss.

PR #3382, opened 2021-12-07
Smallpatches

InfusOnWoW:


InfusOnWoW:
@Stanzilla how can I tell luacheck that I want the code to stay as is?

Stanzilla:
> @Stanzilla how can I tell luacheck that I want the code to stay as is?

You can use inline ignores https://luacheck.readthedocs.io/en/stable/inline.html

PR #3385, opened 2021-12-10
UnitCastingInfo & UnitChannelInfo update

mrbuds:
temporary fix for 2.5.3 UnitCastingInfo & UnitChannelInfo changes, spellid was moved to 9th arg (same as mainline/classic_era)
use them also for classic_era

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -91,7 +91,9 @@ if WeakAuras.IsClassic() then
   LibClassicCasterino = LibStub("LibClassicCasterino")
 end
 
-if WeakAuras.IsRetail() then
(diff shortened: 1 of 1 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -101,7 +103,7 @@ elseif WeakAuras.IsBCC() then
 else
   WeakAuras.UnitCastingInfo = function(unit)
     if UnitIsUnit(unit, "player") then
-      return CastingInfo()
+      return UnitCastingInfo("player")
```
InfusOnWoW, line 106: Does this work on both BCC 2.5.3 and livw?

InfusOnWoW (approved):

PR #3394, opened 2021-12-21
subregions tweaks

mrbuds:
Let model subregion be bigger than it's region and have framelevel under foreground
Moved foreground subregion to it's own files, it's not working for icon/textures regions yet


WeakAuras/SubRegionTypes/Foreground.lua
```diff
@@ -0,0 +1,42 @@
...
+    return regionType == "texture"
+    or regionType == "progresstexture"
+    or regionType == "icon"
+    or regionType == "aurabar"
+end
+
+local function noop()
+end
+
+local function SetFrameLevel(self, level)
```
InfusOnWoW, line 15: I think t the cleanest would be if the regions would have a function "SetForegroundFrameLevel", and that the foreground subregion simply calls that function.

InfusOnWoW:
This looks pretty complete, except for the missing Modernize code for progress bars.

WeakAuras/SubRegionTypes/Foreground.lua
mrbuds: Tried that but couldn't make it work for Icon region, finally moved it to `Private.ApplyFrameLevel` and i think it's better

WeakAuras/WeakAuras.lua
```diff
@@ -3501,12 +3501,16 @@ end
...
   if region.subRegions then
     for index, subRegion in pairs(region.subRegions) do
-      subRegion:SetFrameLevel(frameLevel + index + 1)
+      if subRegion.type == "subforeground" then
+        subforegroundIndex = index
+      end
+      subRegion:SetFrameLevel(frameLevel + index)
     end
   end
+  region:SetFrameLevel(frameLevel + subforegroundIndex)
```
InfusOnWoW, line 3513: That's not correct though, since it affects the frame level of all region parts. For aurabars it did previously only affect the foreground.

Stepping back a bit:
The use case I had in mind when I introduced the dummy sub element, was to enable the user to add a model subelement attached to the background, with the foreground in front, and the background behind. For that reason the sub element only affected the foreground, and the background was set via ApplyFrameLevel=>region:SetFrameLevel

You are going further in wanting to allow the user to put the background on top of a sub element. And then for all region types.

I think the right solution is this:
* All regions must have at least one dummy sub element that controls the region framelevel
=> I think that ApplyFrameLevel should never call region:SetFrameLevel. Instead the dummy sub element should call SetFrameLevel (or maybe another function, see next)

* The aura bar has two dummy sub elements, one for the background, and one for the foreground. 
=> We need to distinguish these, probably with some kind of naming store in the element, which is then passed to the region as a parameter.

mrbuds: > => We need to distinguish these, probably with some kind of naming store in the element, which is then passed to the region as a parameter.

I'm not sure i exactly understand what you mean, i made some changes in last commit without squishing. 

AuraBar has now a subelement for Background + an other one for Foreground.
It works but it's clunky ...
- "foreground" is a bad name for "Progress" maybe?
- foreground should not be movable behind background

Also i'm thinking of remove oversize options for submodel when clipped by progress is checked.

InfusOnWoW: > > => We need to distinguish these, probably with some kind of naming store in the element, which is then passed to the region as a parameter.
> 
> I'm not sure i exactly understand what you mean, i made some changes in last commit without squishing.

Ah this is good enough. I was maybe overengineering it a bit. 
 
> AuraBar has now a subelement for Background + an other one for Foreground. It works but it's clunky ...
 >     * "foreground" is a bad name for "Progress" maybe?
It's currently icon+progress foreground, so I guess foreground is fine.
 
>     * foreground should not be movable behind background
I'm not at all concerned with that. If the user wants a non-sensical layout, we don't need to prevent. Especially when considering my vision for sub elements. 

The introduction of sub elements wasn't just to add new features in a better way than before. But also to eventually split up existing region types, so that you can mix your auras in the way you want. That is, add a small "progress bar" on the side of your icon! Add a texture popup to your progress bar! A stop model overlay that spicies up your icon. Etc. 

So eventually I want to have the foreground not be a dummy sub element, but a reusable element that can be added to any region type.

And in that vision people then can obviously set up auras in non-sensical ways, but that's just the price for flexibility.


> Also i'm thinking of remove oversize options for submodel when clipped by progress is checked.
Right, that's good.

PR #3405, opened 2022-01-01
Fix Modernize of barmodels

InfusOnWoW:


mrbuds:
internalVersion is still 47 in WeakAuras.lua

mrbuds (approved):

PR #3408, opened 2022-01-02
subbackground should support everything but groups

### LINKED ISSUE #3406: Subregion Error on StopMotion Auras.
### Description

Every StopMotion aura is giving an error when loading in.

WeakAuras:  ERROR in '[AuraName]' unknown or incompatible sub element type 'subbackground'

### WeakAuras Version

3.7.3-43-g2716714

### World of Warcraft Flavor

Retail (Default)

- [ ] Yes
- [X] No

### Lua Error

No Lua error.  Just an error printed in the chat box.

WeakAuras:  ERROR in '[AuraName]' unknown or incompatible sub element type 'subbackground'

### Reproduction Steps

Make a Stop Motion Aura, then ReloadUI.

### Last Good Version

The version that I downloaded off of GIT at the end of my last ticket.

### Screenshots

_No response_

### Export String

_No response_

mrbuds:
fixes #3406

### DIFF
--- a/WeakAuras/SubRegionTypes/Background.lua
+++ b/WeakAuras/SubRegionTypes/Background.lua
@@ -4,11 +4,7 @@ local L = WeakAuras.L;
 
 do
     local function subSupports(regionType)
-        return regionType == "texture"
-        or regionType == "progresstexture"
-        or regionType == "icon"
-        or regionType == "aurabar"
-        or regionType == "text"
+        return regionType ~= "group" and regionType ~= "dynamicgroup"
     end
 
     local function noop()


WeakAuras/SubRegionTypes/Background.lua
```diff
@@ -4,11 +4,7 @@ local L = WeakAuras.L;
 
 do
     local function subSupports(regionType)
-        return regionType == "texture"
-        or regionType == "progresstexture"
-        or regionType == "icon"
-        or regionType == "aurabar"
-        or regionType == "text"
+        return not regionType:match("group")
```
InfusOnWoW, line 7: I kinda perfer being explicit, that is checking against "group" and "dynamicgroup". That makes it easier to find the code if e.g. searching for everything dynamic group related.

PR #3451, opened 2022-01-25
additional WA logo textures

Translitdk:
# Description

Additional WA logo textures. Default .blp is low res

## Type of change

## How Has This Been Tested

Just some textures. Nothing special to test

## Checklist

no new code used


InfusOnWoW:
@mrbuds, @Stanzilla any opinion on this? I don't care either way.

Stanzilla:
Fine by me tbh

Stanzilla:
Don't think we need the 512 tbh? Nobody is going to use it that large and the file is also rather big

mrbuds:
they need to be added to the texture selector

Translitdk:
Added 512 one because i tried to use default one for background texture of EditBox and it was awful.

mrbuds:
Added 64px & 256px to texture picker in a new "WeakAuras" entry and removed 512px version

PR #3470, opened 2022-02-04
make condition events use RegisterUnitEvent when possible

mrbuds:
this should fix performance issues with the cooldown trigger with "Insufficient Resources" and "Spell Usable" conditions
add support for filtering in custom checks too
does not support multi units (group, nameplate, ...)

**This will break existing custom checks conditions using ":" as separator**

### DIFF
--- a/WeakAuras/Conditions.lua
+++ b/WeakAuras/Conditions.lua
@@ -698,6 +698,22 @@ local function handleDynamicConditions(self, event)
   Private.StopProfileSystem("dynamic conditions")
 end
 
+local function handleDynamicConditionsPerUnit(self, event, unit)
+  Private.StartProfileSystem("dynamic conditions")
+  if unit then
+    local unitEvent = event..":"..unit
+    if globalDynamicConditionFuncs[unitEvent] then
+      for i, func in ipairs(globalDynamicConditionFuncs[unitEvent]) do
+        func(globalConditionState);
+      end
+    end
+    if (dynamicConditions[unitEvent]) then
+      runDynamicConditionFunctions(dynamicConditions[unitEvent]);
+    end
+  end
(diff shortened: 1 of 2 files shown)


WeakAuras/Conditions.lua
```diff
@@ -781,14 +797,29 @@ function Private.RegisterForGlobalConditions(uid)
...
+        pcall(dynamicConditionsFrame.RegisterEvent, dynamicConditionsFrame, event);
+      end
     end
   end
 end
 
 function Private.UnregisterForGlobalConditions(uid)
   for event, condFuncs in pairs(dynamicConditions) do
     condFuncs[uid] = nil;
+    if next(dynamicConditions) == nil then
```
InfusOnWoW, line 818: I this should be next(dynamicConditions)

WeakAuras/Conditions.lua
```diff
@@ -781,14 +797,29 @@ function Private.RegisterForGlobalConditions(uid)
...
+      if not event:match("([^:]+):([^:]+)") then
+        dynamicConditionsFrame:UnregisterEvent(event)
+      end
```
InfusOnWoW, line 821: Was there a reason you didn't add unregistering for the unit events?

mrbuds:
@InfusOnWoW https://github.com/WeakAuras/WeakAuras2/pull/3470/commits/36529523d81438378be2e188a0497e8b02b3517f should fix your comments

InfusOnWoW:
We should actually also unregister in UnloadAllConditions

mrbuds:
> We should actually also unregister in UnloadAllConditions

fixed, and also fixed a bug in UnregisterForGlobalConditions not cleaning dynamicConditions that prevented events to be registered again

PR #3474, opened 2022-02-06
Fix swing timer resets and pauses

Crablicious:
# Description

Update lists of spells resetting and not resetting swing timer.

Previous behavior:
- Drinking a swing-resetting potion would not cause swing timer to reset.
- Starting a cast of slam would pause the swing timer.

New behavior:
- Drinking a swing-resetting potion causes swing timer to reset.
- Starting a cast of slam does not pause swing timer.

## Type of change

## How Has This Been Tested

Testing with a basic swing timer (https://wago.io/VF0rkqXwQ) and two handed sword in TBC Classic.

All basic non-zone specific potions/drinks have been tested. Leaves out Nethergon Vapor/Cenarion Healing Salve, but they should work the same.

Slam change is easily tested by starting a cast of slam mid-swing and then cancelling the slam-cast before it lands. The swing timer should be unaffected by this.

## Checklist

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -3448,6 +3448,10 @@ if WeakAuras.IsBCC() then
     2764, 3018, -- Shoots,
     19434, 20900, 20901, 20902, 20903, 20904, 27065, -- Aimed Shot
     20066, -- Repentance
+    11350, -- Fire Shield (Oil of Immolation)
+    50986, -- Sulfuron Slammer
+    439, 440, 441, 2024, 4042, 17534, 28495, -- Minor/Lesser/Greater/Superior/Major/Super Healing Potion
+    41619, 41620 -- Cenarion Healing Salve/Bottled Nethergon Vapor
   }
   for _, spellid in ipairs(reset_swing_spell_list) do
     Private.reset_swing_spells[spellid] = true
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -3229,6 +3229,12 @@ Private.noreset_swing_spells = {
   [35478] = true, -- Drums of Restoration
   [34120] = true, -- Steady Shot (rank 1)
   [19434] = true, -- Aimed Shot (rank 1)
+  [1464] = true, -- Slam (rank 1)
+  [8820] = true, -- Slam (rank 2)
+  [11604] = true, -- Slam (rank 3)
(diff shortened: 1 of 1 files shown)


WeakAuras/Types.lua
```diff
@@ -3448,6 +3448,10 @@ if WeakAuras.IsBCC() then
     2764, 3018, -- Shoots,
     19434, 20900, 20901, 20902, 20903, 20904, 27065, -- Aimed Shot
     20066, -- Repentance
+    11350, -- Fire Shield (Oil of Immolation)
+    50986, -- Sulfuron Slammer
+    439, 440, 441, 2024, 4042, 17534, 28495, -- Minor/Lesser/Greater/Superior/Major/Super Healing Potion
+    41619, 41620 -- Cenarion Healing Salve/Bottled Nethergon Vapor
```
mrbuds, line 3454: These spells are added in the BCC section, these are classic spells

WeakAuras/Types.lua
```diff
@@ -3229,6 +3229,12 @@ Private.noreset_swing_spells = {
   [35478] = true, -- Drums of Restoration
   [34120] = true, -- Steady Shot (rank 1)
   [19434] = true, -- Aimed Shot (rank 1)
+  [1464] = true, -- Slam (rank 1)
+  [8820] = true, -- Slam (rank 2)
+  [11604] = true, -- Slam (rank 3)
+  [11605] = true, -- Slam (rank 4)
+  [25241] = true, -- Slam (rank 5)
+  [25242] = true, -- Slam (rank 6)
```
mrbuds, line 3237: Did you check if the behavior is correct on classic_era?

WeakAuras/Types.lua
Crablicious: Yes, they are indeed classic spells, but they are still used in BCC and the classic reset_swing_spell_list is not used in BCC, because it's reliant on WeakAuras.IsClassic() being true.

The BCC table with reset_swing_spell_list already has duplication of most classic spells.

WeakAuras/Types.lua
Crablicious: I have not. I sadly don't have a character available to test with there either without paying Blizzard for my classic era character.

So the safe options would be to either only include the new BCC version here (this is what the majority of people would care about), or generate it similarly to reset_swing_spells.

mrbuds (approved):

PR #3481, opened 2022-02-12
[WIP] StopMotion: support more flexible frame size

mrbuds:
# Description

With current options you need to have frames (width * cols) and (height * rows) to be a power of 2, this change let you use whatever individual frame size you want, only image size have to be a power of 2.

example image:  [monkeys.zip](https://github.com/WeakAuras/WeakAuras2/files/8053168/monkeys.zip)
this work with these settings:
```
  cols: 5,
  rows: 4,
  frameCount: 19,
  fileWidth: 1024,
  fileHeight: 1024,
  frameWidth: 186,
  frameHeight: 249
```

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## TODO

- add a toggle?
- add background support
- tests

### DIFF
--- a/WeakAuras/RegionTypes/StopMotion.lua
+++ b/WeakAuras/RegionTypes/StopMotion.lua
@@ -126,7 +126,7 @@ local function setTile(texture, frame, rows, columns, frameScaleW, frameScaleH)
 
   local top = deltaY * row;
   local bottom = top + deltaY;
-  texture:SetTexCoord(left, right, top, bottom);
+  pcall(function() texture:SetTexCoord(left, right, top, bottom) end)
 end
 
 WeakAuras.setTile = setTile;
--- a/WeakAuras/RegionTypes/StopMotion.lua
+++ b/WeakAuras/RegionTypes/StopMotion.lua
@@ -154,8 +154,9 @@ end
 
 local function modify(parent, region, data)
     WeakAuras.regionPrototype.modify(parent, region, data);
-
+    region.foreground = region.foreground or {}
     local pattern = "%.x(%d+)y(%d+)f(%d+)%.[tb][gl][ap]"
+    local pattern2 = "%.x(%d+)y(%d+)f(%d+)w(%d+)h(%d+)W(%d+)H(%d+)%.[tb][gl][ap]"
     local tdata = texture_data[data.foregroundTexture];
     if (tdata) then
       local lastFrame = tdata.count - 1;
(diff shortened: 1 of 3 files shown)


WeakAurasOptions/RegionOptions/StopMotion.lua
```diff
@@ -384,6 +419,10 @@ local function modifyThumbnail(parent, region, data, fullModify, size)
         region.endFrame = floor( (data.endPercent or 1) * lastFrame) + 1;
         region.foregroundRows = data.customForegroundRows;
         region.foregroundColumns = data.customForegroundColumns;
+        region.foregroundFileWidth = data.customForegroundFileWidth
```
InfusOnWoW, line 422: There's special code a few lines above this which auto-detects the custom settings from the fileName, which sounds like something that would help your workflow.

WeakAuras/RegionTypes/StopMotion.lua
```diff
@@ -143,8 +154,9 @@ end
 
 local function modify(parent, region, data)
     WeakAuras.regionPrototype.modify(parent, region, data);
-
+    region.foreground = region.foreground or {}
     local pattern = "%.x(%d+)y(%d+)f(%d+)%.[tb][gl][ap]"
+    local pattern2 = "%.x(%d+)y(%d+)f(%d+)w(%d+)h(%d+)W(%d+)H(%d+)%.[tb][gl][ap]"
     local tdata = texture_data[data.foregroundTexture];
     if (tdata) then
```
InfusOnWoW, line 161: You need to ensure that all 3 differnt configuration options set all properties.

That is consider the case where someone first selects a pattern2 file, and then switches to one of the other configuration methods.

WeakAurasOptions/RegionOptions/StopMotion.lua
```diff
@@ -361,29 +403,47 @@ local function modifyThumbnail(parent, region, data, fullModify, size)
...
+      local pattern2 = "%.x(%d+)y(%d+)f(%d+)w(%d+)h(%d+)W(%d+)H(%d+)%.[tb][gl][ap]"
       local rows, columns, frames = data.foregroundTexture:lower():match(pattern)
-      if rows and columns and frames then
+      if rows then
         local lastFrame = frames - 1;
         region.startFrame = floor( (data.startPercent or 0) * lastFrame) + 1;
         region.endFrame = floor( (data.endPercent or 1) * lastFrame) + 1;
-        region.foregroundRows = rows;
-        region.foregroundColumns = columns;
+        region.foreground.rows = tonumber(rows);
```
InfusOnWoW, line 422: Same here everythung needs to be set

InfusOnWoW:
Looks pretty good already.

Issue #3488, opened 2022-02-18
Lua Error since Version 3.7.9


Cheveyo10:
### Description

I updated my Weakauras Addon to 3.7.9 and since that Update I recieving lua errors and its everytime the same Error. 

### WeakAuras Version

3.7.9

### World of Warcraft Flavor

Retail (Default)

### Tested with only WeakAuras

- [ ] Yes
- [X] No

### Lua Error

1x WeakAuras\Conditions.lua:831: Attempt to unregister unknown event "ITEM_COOLDOWN_CHANGED"
[string "=[C]"]: in function UnregisterEvent'
[string "@WeakAuras\Conditions.lua"]:831: in function UnregisterForGlobalConditions'
[string "@WeakAuras\Conditions.lua"]:862: in function UnloadConditions'
[string "@WeakAuras\WeakAuras-3.7.9.lua"]:1735: in function UnloadDisplays'
[string "@WeakAuras\WeakAuras-3.7.9.lua"]:1521: in function <WeakAuras\WeakAuras.lua:1393>
[string "@WeakAuras\WeakAuras-3.7.9.lua"]:1557: in function `ScanForLoads'
[string "@WeakAuras\WeakAuras-3.7.9.lua"]:1618: in function <WeakAuras\WeakAuras.lua:1614>
[string "@SharedXML\C_TimerAugment.lua"]:16: in function <SharedXML\C_TimerAugment.lua:14>

Locals:
(temporary) = <unnamed> {
 0 = <userdata>
 onUpdate = true
 units = <table> {
 }
}
(temporary) = "ITEM_COOLDOWN_CHANGED"

### Reproduction Steps

Updated to Version 3.7.9 

### Last Good Version

3.7.8

Cheveyo10:
[Weakauras SavedVariables.txt](https://github.com/WeakAuras/WeakAuras2/files/8100130/Weakauras.SavedVariables.txt)

InfusOnWoW:
Should be fixed for the next release.

PR #3489, opened 2022-02-18
fix "unregister unknown event" error in conditions

mrbuds:
# Description

potential fix for  #3488

Reporter has not submitted his SV to test the fix yet.
There may be an other fix to do, as an internal event should not be registered/unregistered like that


WeakAuras/Conditions.lua
```diff
@@ -819,7 +819,7 @@ function Private.UnregisterForGlobalConditions(uid)
     if next(condFuncs) == nil then
       local unitEvent, unit = event:match("([^:]+):([^:]+)")
       if unitEvent and unit then
-        dynamicConditionsFrame.units[unit]:UnregisterEvent(unitEvent)
+        pcall(dynamicConditionsFrame.UnregisterEvent.units[unit], dynamicConditionsFrame.units[unit], unitEvent, unit);
```
InfusOnWoW, line 822: That doesn't look correct:
dynamicConditionsFrame.UnregisterEvent.units[unit]

Also the unit parameter should not be necessary.

mrbuds:
updated

InfusOnWoW:
@mrbuds can you check with the settings on the ticket?

mrbuds:
I searched for "Cooldown Progress (Item)" triggers, made a copy of "Phial of Serenity (Pala)" and adapted it to work with a mana potion, load condition in combat, used a potion in combat then got out of combat, and it didn't trigger any error with this change.

PR #3490, opened 2022-02-18
bcc: fix Character Stats > hastepercent

mrbuds:
# Description

tested on SL/BCC/SOM


Stanzilla:
@mrbuds needs the luacheckrc addition

PR #3495, opened 2022-02-19
Smallfix

### LINKED ISSUE #3494: Incorrect information for some factions (the Faction Reputation trigger)
### Description

There are several special single-NPC factions on Retail. Examples:
- The Archivists' Codex
- Ve'nari
- Chromie
- Bizmo's Brawlpub
- Brawl'gar Arena
- Tormmok
- Leorajh
- Talonpriest Ishaal
- etc.

WeakAuras doesn't take into account these cases, provides and contains incorrect information about such factions. [GetFriendshipReputation()](https://wowpedia.fandom.com/wiki/API_GetFriendshipReputation) and [GetFriendshipReputationRanks()](https://wowpedia.fandom.com/wiki/API_GetFriendshipReputationRanks) functions should be used to get details. Last one, fortunately for us, again returns values since an unknown Shadowlands patch (I guess it's 9.1.5). 

Notes:
- There are only problems with standing and standing ID (ID may be greater than 8 for some factions) at the present time, but I would get all data from the above functions and not from GetFactionInfoByID()
- No simple way to get text of all player's standing ranks with a friend (only current). Will have to make a list with common values (like rank #i) for the Standing option. But I would like %standing to still return a localized text.

P.S. Blizzard uses the following code to get standing text for normal factions, because in some languages (ES, FR, IT etc) it depends on gender of the character:
```lua
GetText("FACTION_STANDING_LABEL"..standingID, UnitSex("player"))
```
I don't know how important this is, but we could do the same [here](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAuras/Prototypes.lua#L2156) and [here](
https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAuras/Prototypes.lua#L2228)

### WeakAuras Version

3.7.8-13-gdb3bd18

### World of Warcraft Flavor

Retail (Default)

- [X] Yes
- [ ] No

### Lua Error

No erros

### Reproduction Steps

1. Create aura with Faction Reputation trigger
2. Select any single-NPC faction (in [that](https://wowpedia.fandom.com/wiki/FactionID) table they are marked with a special icon)
3. Сompare data from aura and in the player's reputation pane

### Last Good Version

_No response_

### Screenshots

_No response_

### Export String

_No response_

mrbuds:
## Faction Reputation

add 2 friendship filters in trigger, condition and text replacement
image
image
image

text replacement for %standing will use friendship when it's not nil

updated the way we get standing text for gendered labels

Fixes #3494

## Character Stats

Revert hastepercent change for retail

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -2153,9 +2153,18 @@ Private.event_prototypes = {
         end
         local standing
         if tonumber(standingId) then
-           standing = getglobal("FACTION_STANDING_LABEL"..standingId)
+           standing = GetText("FACTION_STANDING_LABEL"..standingId, UnitSex("player"))
         end
       ]=]
+      if WeakAuras.IsRetail() then
+        ret = ret .. [=[
+          local _, _, _, _, _, _, friendTextLevel = GetFriendshipReputation(factionID)
+          if friendTextLevel then
+            standing = friendTextLevel
+          end
+          local friendshipRank, friendshipMaxRank = GetFriendshipReputationRanks(factionID)
+        ]=]
+      end
       return ret:format(trigger.factionID or 0, trigger.use_watched and "true" or "false")
     end,
     statesParameter = "one",
@@ -2225,7 +2234,7 @@ Private.event_prototypes = {
         values = function()
           local ret = {}
           for i = 0, 8 do
-            ret[i] = getglobal("FACTION_STANDING_LABEL"..i)
+            ret[i] = GetText("FACTION_STANDING_LABEL"..i, UnitSex("player"))
           end
           return ret
         end,
@@ -2241,6 +2250,28 @@ Private.event_prototypes = {
         store = "true",
         hidden = "true",
         test = "true"
+      },
+      {
+        name = "friendshipRank",
+        display = L["Friendship Rank"],
+        type = "number",
+        init = "friendshipRank",
+        store = "true",
+        test = "true",
+        conditionType = "number",
+        enable = WeakAuras.IsRetail(),
+        hidden = not WeakAuras.IsRetail()
+      },
+      {
+        name = "friendshipMaxRank",
+        display = L["Friendship Max Rank"],
+        type = "number",
+        init = "friendshipMaxRank",
+        store = "true",
+        test = "true",
+        conditionType = "number",
+        enable = WeakAuras.IsRetail(),
+        hidden = not WeakAuras.IsRetail()
       },
       {
         name = "progressType",
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -8145,7 +8145,7 @@ Private.event_prototypes = {
         name = "hastepercent",
         display = L["Haste (%)"],
         type = "number",
-        init = WeakAuras.IsClassic() and "GetHaste()" or "GetCombatRatingBonus(CR_HASTE_SPELL)",
+        init = WeakAuras.IsBCC() and "GetCombatRatingBonus(CR_HASTE_SPELL)" or "GetHaste()",
         store = true,
         conditionType = "number"
       },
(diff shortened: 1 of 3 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -2153,7 +2153,6 @@ Private.event_prototypes = {
         end
         local standing
         if tonumber(standingId) then
-           standing = getglobal("FACTION_STANDING_LABEL"..standingId)
```
InfusOnWoW, line 2156: you accidentally removed this line here and readded it in the next commit.

You can fix that with a bit of git-fu.

Start with both git commits applied.
Run "git reset -p HEAD~1"
This will show you each chunk, answer "y" for this chunk, "n" for every other
You can examine what this does via. git diff --staged vs git diff
Next is
"git commit --amend"
"git commit -m squash"

Now git rebase -i HEAD~3
move the squash commit to the second line and replace "pick" with fixup.

Nightwarden24:
I'm not sure if that's what I meant. You just added Friendship Rank and Friendship Max Rank as new options. I imagined it a little differently.

- Check each faction: if friend ID exists, get ALL data from GetFriendshipReputation() and GetFriendshipReputationRanks() and not only friendTextLevel. StandingId may already be greater than 8 for some fractions. I'm not sure if the rest of values match with information from GetFactionInfo(). For example, Blizzard doesn't use this function to get details in frienship faction case.
- Create a list for [the Standing option](https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAuras/Prototypes.lua#L2225) and condition with common values like rank#i. Their number is equal maxRank. In this case don't use FACTION_STANDING_LABEL at all
- %standing  returns text of the player's current standing with the friend (you already did that, standing = friendTextLevel)

Your fix leaves the Standing option with incorrect values for friendship faction and displays Friendship Rank for normal faction at the same time. Will the user be able to figure out what should be used? I don't know

mrbuds:
> * Check each faction: if friend ID exists, get ALL data from GetFriendshipReputation() and GetFriendshipReputationRanks() and not only friendTextLevel. StandingId may already be greater than 8 for some factions. I'm not sure if the rest of values match with information from GetFactionInfo(). For example, Blizzard doesn't use this function to get details in frienship faction case

I don't like mixing standing & friendship together, it make sense for %standing but not for identifiers.

> Your fix leaves the Standing option with incorrect values for friendship faction and displays Friendship Rank for normal faction at the same time. Will the user be able to figure out what should be used? I don't know

friendshipRank & friendshipMaxRank conditions were added so user can show or hide a text only when it's relevant

Nightwarden24:
> > Check each faction: if friend ID exists, get ALL data from GetFriendshipReputation() and GetFriendshipReputationRanks() and not only friendTextLevel. StandingId may already be greater than 8 for some factions. I'm not sure if the rest of values match with information from GetFactionInfo(). For example, Blizzard doesn't use this function to get details in frienship faction case
> 
> I don't like mixing standing & friendship together, it make sense for %standing but not for identifiers.

Then create two arguments (standingID and friendshipID), name them the same (display = L["Standing"]). Depending on which faction is selected (normal or friendship) display the appropriate argument in options and conditions

> >Your fix leaves the Standing option with incorrect values for friendship faction and displays Friendship Rank for normal faction at the same time. Will the user be able to figure out what should be used? I don't know
>
>friendshipRank & friendshipMaxRank conditions were added so user can show or hide a text only when it's relevant

I meant options and their values on the Trigger tab

mrbuds:
> Then create two arguments (standingID and friendshipID), name them the same (display = L["Standing"]). Depending on which faction is selected (normal or friendship) display the appropriate argument in options and conditions

No, a standingID is a standingID  and a friendshipID is a friendshipID, making a field that randomly show one or the other makes no sense, but i will a friendshipID in data

Nightwarden24:
Do what you think is right or you want.
\*disappointment\*

mrbuds:
added fix for https://github.com/WeakAuras/WeakAuras2/issues/3496
and enable autoshine glow on bcc/som thanks to https://github.com/Stanzilla/LibCustomGlow/commit/f15e6de03f59acff163c5d5f7d66fed5dbcbc897

Stanzilla (approved):

PR #3499, opened 2022-02-21
Handle Dynamic Info section in the Activation Function

asaka-wa:
# Description

If the returned value is a number and it's a valid trigger number, then use it to set the Dynamic Info. Otherwise revert to data value. 

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -3843,6 +3843,7 @@ local function evaluateTriggerStateTriggers(id)
     if (triggerState[id].disjunctive == "custom" and triggerState[id].triggerLogicFunc) then
       local ok, returnValue = xpcall(triggerState[id].triggerLogicFunc, geterrorhandler(), triggerState[id].triggers);
       result = ok and returnValue;
+        triggerState[id].activeTriggerMode = type(returnValue) == "number" and returnValue <= triggerState[id].numTriggers and returnValue or db.displays[id].triggers.activeTriggerMode
     end
   end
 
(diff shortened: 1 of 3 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -3843,6 +3843,7 @@ local function evaluateTriggerStateTriggers(id)
     if (triggerState[id].disjunctive == "custom" and triggerState[id].triggerLogicFunc) then
       local ok, returnValue = xpcall(triggerState[id].triggerLogicFunc, geterrorhandler(), triggerState[id].triggers);
       result = ok and returnValue;
+        triggerState[id].activeTriggerMode = type(returnValue) == "number" and returnValue <= triggerState[id].numTriggers and returnValue or db.displays[id].triggers.activeTriggerMode
```
mrbuds, line 3847: I see a few issues with that implementation
- using `returnValue` will break auras already returning a number.
- evaluateTriggerStateTriggers returns `result` but doesn't edit any table, it should return another value and use it later when making `newActiveTrigger`
- UI still show a dropdown for `Dynamic Information`, that's why i think ability to custom code which trigger is the dynamic info belongs as a new option in that dropdown and then show a new custom code widget

asaka-wa: This was kind of a proof of concept to see if it's a PR that would be considered. 
https://discord.com/channels/172440238717665280/273144328447197186/945357029290164234

> using returnValue will break auras already returning a number.

We could ask Ora to check if it's being done at all but my guess is it isn't, and if it is it won't break anything, just display things unexpectedly. 

> evaluateTriggerStateTriggers returns result but doesn't edit any table, it should return another value and use it later when making newActiveTrigger

> UI still show a dropdown for Dynamic Information, that's why i think ability to custom code which trigger is the dynamic info belongs as a new option in that dropdown and then show a new custom code widget

Yup, I think I get all that. If this is something that would be considered, conceptually, then I can work on tidying it up.

PR #3535, opened 2022-03-13
Smallfix

mrbuds:
# Description

Revert hastepercent change for bcc:
GetHaste() value is bugged with real haste you have between gear's haste, drums and bloodlust, but no one complained and now we get one for not matching character sheet.

Combat Log trigger: add spellSchool filter:
used <https://wowpedia.fandom.com/wiki/COMBAT_LOG_EVENT#Spell_School> and global strings
add %spellSchool (mask value) and %spellSchoolString (localised string)
dropdown is a mess, order is bad, i'm not happy with it, but it'll do the job.
I choosed to use bit.band for matching, so "frostfire" will match "fire", i'm not 100% sold if that's best.

Spell Known trigger: add inverse option:
tested with known spell & unknown spell with both exact spellid and not


WeakAuras/Prototypes.lua
```diff
@@ -3739,10 +3743,22 @@ Private.event_prototypes = {
         conditionType = "string"
       },
       {
+        name = "spellSchool",
+        display = L["Spell School"],
+        type = "select",
+        values = "combatlog_spell_school_types_for_ui",
+        test = "spellSchool and bit.band(spellSchool, %q) ~= 0",
```
InfusOnWoW, line 3750: That doesn't work, in two ways:

a) E.g. a check for holy, also will match a "holyfire" event. (Or is that what the user expects?). That would make e.g. Cosmic events match "every" filter.
b) A check for "holyfire", also will match a "holy" event. That's clearly not right. 

So I think we actually might want to have state.spellSchool == %q ?

WeakAuras/Prototypes.lua
```diff
@@ -3994,7 +4010,19 @@ Private.event_prototypes = {
...
+        display = L["Spell School String"],
+        type = "string",
+        init = "spellSchool and WeakAuras.SpellSchool(spellSchool)",
+        store = true,
+        test = "true",
+        hidden = true,
+        enable = function(trigger)
+          return trigger.subeventPrefix and (trigger.subeventPrefix:find("SPELL") or trigger.subeventPrefix == "RANGE" or trigger.subeventPrefix:find("DAMAGE"))
+        end
+      },
```
InfusOnWoW, line 4025: I don't know about that. Do you think it's useful enough?

WeakAuras/Prototypes.lua
mrbuds: let's go for spellSchool == %d

WeakAuras/Prototypes.lua
mrbuds: It can be useful for the combat text use case, i don't mind removing it

PR #3548, opened 2022-03-22
handle CDR with modRate

### LINKED ISSUE #3544: CDR (cooldown reduction) isn't handled by WA
### Description

While the `modRate` value on `GetSpellCooldown` and `etSpellCharges` has been around for a while there really haven't been many notable examples of it being used until the recently added Encrypted Affix in M+

While buffed player CDs run faster. Currently WA continues to just use `start` and `duration` values which make the displayed CD jump dramatically rather than just smoothly run faster. The default Blizz UI sends `modRate` into `SetCooldown` on the CD widget but presumably it just makes the widget run faster

### WeakAuras Version

3.7.13

### World of Warcraft Flavor

Retail (Default)

- [ ] Yes
- [X] No

### Lua Error

n/a

### Reproduction Steps

1. Have a M+10 level character
2. Run a M+ of +10 or higher
3. have the group kill the [Urh Relic](https://ptr.wowhead.com/npc=185685/urh-relic) first, then the [Urh Dismantler](https://ptr.wowhead.com/npc=184911/urh-dismantler) that spawns
4. The group will get the [Decrypted Urh Cypher](https://ptr.wowhead.com/spell=368239/decrypted-urh-cypher) buff
5. See the difference between the way WA displays the CD and the way the default UI displays it

### Last Good Version

n/a

### Screenshots

_No response_

### Export String

_No response_

mrbuds:
# Description

Use modRate modifier for spell cooldown
Fixes #3544

## Type of change

## TESTS

- [x] Compare weakauras vs actionbar on paladin with avenging wrath

## DONE & TODO

- [x] Icon swipe
- [x] %p
- [ ] AuraBar
- [ ] ProgressTexture
- [x] Spell with charge not working yet

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -4109,7 +4109,7 @@ Private.event_prototypes = {
         local ignoreSpellKnown = %s;
         local track = %q
         local startTime, duration, gcdCooldown, readyTime, modRate = WeakAuras.GetSpellCooldown(spellname, ignoreRuneCD, showgcd, ignoreSpellKnown, track);
-        local charges, maxCharges, spellCount, chargeGainTime, chargeLostTime, modRateCharges = WeakAuras.GetSpellCharges(spellname, ignoreSpellKnown);
+        local charges, maxCharges, spellCount, chargeGainTime, chargeLostTime = WeakAuras.GetSpellCharges(spellname, ignoreSpellKnown);
         local stacks = maxCharges and maxCharges ~= 1 and charges or (spellCount and spellCount > 0 and spellCount) or nil;
         if (charges == nil) then
           -- Use fake charges for spells that use GetSpellCooldown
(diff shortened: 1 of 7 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -4118,6 +4118,8 @@ Private.event_prototypes = {
         local genericShowOn = %s
         local expirationTime = startTime and duration and startTime + duration
         state.spellname = spellname;
+        state.modRate = modRate;
+        state.modRateCharges = modRateCharges;
```
InfusOnWoW, line 4122: This should only set state.modRate, to either modRate/modRteCharges, depending on what expirationTime was used.

mrbuds:
Did same thing as for Icons with AuraBar & ProgressTextures, but reverted it as (p/x) / (t/x) = p/t calculated progress is the same.
It may still be important for progressTexture's tick subregion positioning ...

mrbuds:
Added modRate support with bufftrigger2, i didn't get any error but:
- i don't know how to test it, which auras or buffs make it not equal to 1
- sort by expirationTime doesn't handle modRate

mrbuds:
Tested your change on bcc, all good

PR #3557, opened 2022-03-25
new global range check condtion

mrbuds:
# Description

Add a new global condition for testing range with group & enemies

image
image
image

## Type of change

## How Has This Been Tested

- [x] Tested with enemy nameplates
- [ ] Tested with group (i don't have a group :-( )

### DIFF
--- a/WeakAuras/Conditions.lua
+++ b/WeakAuras/Conditions.lua
@@ -271,6 +271,52 @@ local function CreateTestForCondition(uid, input, allConditionsTemplate, usedSta
       else
         check = stateCheck .. stateVariableCheck .. "state[" .. trigger .. "]".. string.format("[%q]", variable) .. op .. "'" .. value .. "'";
       end
+    elseif (cType == "range" and value and op and input.type and input.op_range and input.range) then
+      local fn
+      if input.type == "group" then
+        fn = [[
+          return function()
+            local found = 0
+            local op = %q
+            local range = %s
+            for unit in WA_IterateGroupMembers() do
+              if not UnitIsUnit(unit, "player") and WeakAuras.CheckRange(unit, range, op) then
+                found = found + 1
+              end
+            end
+            return found %s %d
+          end
+        ]]
+        fn = fn:format(input.op_range, input.range, op, value)
+      elseif input.type == "enemies" then
+        fn = [[
+          return function()
+            local found = 0
+            local op = %q
+            local range = %s
+            for i = 1, 40 do
+              local unit = "nameplate" .. i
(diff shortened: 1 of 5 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -8014,6 +8020,17 @@ Private.event_prototypes = {
...
+      {
+        name = "onUpdateUnitTarget",
+        display = L["Advanced Caster's Target Check"],
+        desc = L["Check nameplate's target every 0.2s"],
+        type = "toggle",
+        test = "true",
+        enable = function(trigger)
+          return trigger.unit == "nameplate"
+        end,
+        reloadOptions = true
```
InfusOnWoW, line 8049: I don't think that's necessary.

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -1548,48 +1548,33 @@ local function addControlsForIfLine(args, order, data, conditionVariable, totalA
...
-    if (data.controlledChildren) then
-      setOp = function(info, v)
-        check = getOrCreateSubCheck(conditions[i].check, path);
-        for id, reference in pairs(conditions[i].check.references) do
-          local auraData = WeakAuras.GetData(id);
-          local childCheck = getOrCreateSubCheck(auraData[conditionVariable][reference.conditionIndex].check, path);
-          childCheck.op = v;
-          WeakAuras.Add(auraData);
-          OptionsPrivate.ClearOptions(auraData.id)
+    local function makeSetter(field)
```
InfusOnWoW, line 1551: I renamed this from "makeSet" to "makeSetter", as that is the proper name for a set function.

WeakAuras/Conditions.lua
```diff
@@ -271,6 +271,52 @@ local function CreateTestForCondition(uid, input, allConditionsTemplate, usedSta
...
       end
+    elseif (cType == "range" and value and op and input.type and input.op_range and input.range) then
+      local fn
+      if input.type == "group" then
+        fn = [[
+          return function()
+            local found = 0
+            local op = %q
+            local range = %s
+            for unit in WA_IterateGroupMembers() do
```
InfusOnWoW, line 282: The reason I took so long was that I'm not entirely happy with this. Each condition iterates all members or enemies to check its status. If there are multiple auras using the same range check, then all of them will use they separate check.

I think there are improvements possible, but they do require a bigger rework of our internal events.

mrbuds: Ideally i think it belongs to LibRangeCheck scope.
It should have functions to handle group & nameplate checks, with a cache clean each new frame

PR #3606, opened 2022-05-01
stopmotion picker: slow down animation frame rate

mrbuds:
# Description

stopmotion picker too fast too furious


WeakAurasOptions/RegionOptions/StopMotion.lua
```diff
@@ -49,25 +49,33 @@ local function setTextureFunc(textureWidget, texturePath, textureName)
         setTile(textureWidget, data.count, data.rows, data.columns, data.frameScaleW or 1, data.frameScaleH or 1);
 
         textureWidget:SetOnUpdate(function()
```
InfusOnWoW, line 51: missing parameter elapsed

mrbuds: bad copy/pasta, pushed an update

Issue #236, opened 2017-07-09
"Pixel Perfect" Sizing, Alignment


Hekili:
**What steps will reproduce the problem?**  

If you are using a UI that is scaled (i.e., Borderless Windowed), your WeakAuras may not be sized or placed in a way that lines up with actual pixels on the screen.

**What is the expected output? What do you see instead?**

https://cdn.discordapp.com/attachments/172440238717665280/333680409495863308/unknown.png

In the graphic above, you can see that some borders are slightly larger than they ought to be -- all are set to 1px, but the placement and sizing of 1px would actually overlap two pixels, so the bottom of the "Stormbringer 6s" bar fills two pixels.  You can also see that the spacing between bars is somewhat irregular, despite being a "Dynamic Group" whose spacing is managed by WA.

**Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?**

n/a

**What version of WeakAuras are you using?**

```
Interface: 70200
Title: WeakAuras 2
Author: Mirrored and the WeakAuras Team
Version: 2.4.9.1-2-gfad9a86
```

**Was it working in a previous version? If yes, which was the last good one?**

No, "pixel perfect" has been discussed before but not implemented.

**Do you have an error log of what happened? If you don't see any errors, make sure that error reporting is enabled (`/console scriptErrors 1`) or install [BugSack](https://mods.curse.com/addons/wow/BugSack). **

n/a

Please provide any additional information below.

After obtaining my UI scale as follows:

```
local monitorIndex = ( tonumber(GetCVar('gxMonitor')) or 0 ) + 1
local resolutions = { GetScreenResolutions() }
local resolution = resolutions[ GetCurrentResolution() ] or GetCVar("gxWindowedResolution")

local scale = GetCVar( "UseUIScale" ) == "1" and ( GetScreenHeight() / resolution:match("%d+x(%d+)") ) or 1
```

... I found my scale value to be 0.833333 (5/6).  Updating the "Space" value on my Dynamic Group ( 4 x 5 / 6 ) = 3.33 and on the sizes of my bars and what, ultimately results in this:

https://cdn.discordapp.com/attachments/172440238717665280/333687505356455936/unknown.png

Example Group WA, tweaked for a 5/6 uiScale:

[...]

PR #3611, opened 2022-05-04
pixel perfect improvement

mrbuds:
# Description

Attempt at fixing some pixel perfect issues
Greatly inspired by OmniCD

## Type of change

## Working

- dynamic group (with & without border) down/right/grid, with icons, and aurabars (with & without border) tested with offsets & size set to 1

## Not working

- group scaling
- fractional size/position

need more tests

### DIFF
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -403,8 +403,8 @@ local barPrototype = {
           end
 
           extraTexture:ClearAllPoints();
-          extraTexture:SetPoint(self.align1, self, self.align1, xOffset * Private.PixelMult, yOffset * Private.PixelMult);
-          extraTexture:SetPoint(self.align2, self, self.align2, xOffset * Private.PixelMult, yOffset * Private.PixelMult);
+          extraTexture:SetPoint(self.align1, self, self.align1, xOffset, yOffset);
+          extraTexture:SetPoint(self.align2, self, self.align2, xOffset, yOffset);
         end
       end
 
--- a/WeakAuras/RegionTypes/AuraBar.lua
+++ b/WeakAuras/RegionTypes/AuraBar.lua
@@ -253,7 +253,7 @@ local barPrototype = {
     self.fgFrame:SetPoint(self.align1);
     self.fgFrame:SetPoint(self.align2);
 
-    self.spark:SetPoint("CENTER", self.fg, self.alignSpark, self.spark.sparkOffsetX or 0, self.spark.sparkOffsetY or 0);
+    self.spark:SetPoint("CENTER", self.fg, self.alignSpark, (self.spark.sparkOffsetX or 0) * Private.PixelMult, (self.spark.sparkOffsetY or 0) * Private.PixelMult);
 
     local sparkMirror = self.spark.sparkMirror;
     local sparkRotationMode = self.spark.sparkRotationMode;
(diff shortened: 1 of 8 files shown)


WeakAuras/RegionTypes/AuraBar.lua
```diff
@@ -404,8 +403,8 @@ local barPrototype = {
           end
 
           extraTexture:ClearAllPoints();
-          extraTexture:SetPoint(self.align1, self, self.align1, xOffset, yOffset);
-          extraTexture:SetPoint(self.align2, self, self.align2, xOffset, yOffset);
+          extraTexture:SetPoint(self.align1, self, self.align1, xOffset * Private.PixelMult, yOffset * Private.PixelMult);
```
InfusOnWoW, line 406: This can't be correct. Let's use these terms:
"uiscale-pixel" => a pixel in wow's coordinate system, unmodifed
"perfect pixel" => a pixel that is adjusted to be pixel-perfect

xOffset is derived from width (and progress)
* width is from GetRealSize()
* which is a uiscale pixel
Thus xOffset is in uiscale pixels.

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -293,8 +292,8 @@ local function modify(parent, region, data)
   end
 
   function region:UpdateSize()
-    local width = region.width * math.abs(region.scalex);
-    local height = region.height * math.abs(region.scaley);
+    local width = region.width * region.scalex;
```
InfusOnWoW, line 295: The math.abs is there for a "reason". I think the flip animation uses negative scales. (Yes it's rather insane, can't change it.)

WeakAuras/RegionTypes/Icon.lua
```diff
@@ -314,8 +313,8 @@ local function modify(parent, region, data)
     if not region.keepAspectRatio then
       aspectRatio = 1;
     else
-      local width = region.width * math.abs(region.scalex);
-      local height = region.height * math.abs(region.scaley);
+      local width = region.width * (Private.PixelMult / region.scalex);
```
InfusOnWoW, line 316: That doesn't look correct. Previously that multipled by scaleX, now it divides?

WeakAuras/WeakAuras.lua
```diff
@@ -1163,6 +1165,8 @@ loadedFrame:SetScript("OnEvent", function(self, event, addon)
     in_loading_screen = true;
   elseif(event == "LOADING_SCREEN_DISABLED") then
     in_loading_screen = false;
+  elseif(event == "UI_SCALE_CHANGED") then
+    Private:SetPixelMult()
```
InfusOnWoW, line 1169: Well this needs to do a lot more, if we want to dynamically adjust to the uiscale. I think that's a bit too hard, and annoying.

I suggest we don't handle that. That is we react to the event, by printing that WAs will adjust on next ui reload. Otherwise you get very strange mix of some regions having updated to the new scale, and others not.

WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -416,8 +416,8 @@ end
 
 local growers = {
   LEFT = function(data)
-    local stagger = -(data.stagger or 0)
-    local space = data.space or 0
+    local stagger = -(data.stagger or 0) * Private.PixelMult
```
InfusOnWoW, line 419: This can't be backwards compatible.

There are a few options:
a) Make a global WeakAuras wide option that enables it. So basically a user has to opt into it.
b) Make this a per aura setting (Or a per group?)
c) Don't multiply but align

Things to consider are:
* Pixel-Perfect auras (whether that's global or per aura setting) are harder to share. Because of the way uiscale uses the same ui scale, the auras "appear" to be similar size regardless of actual screen resolution. (Which is only true for users that don't set the uiscale)
* There's a ticket for it: https://github.com/WeakAuras/WeakAuras2/issues/236
* And Blizzard has some utility functions, see https://github.com/tomrus88/BlizzardInterfaceCode/blob/master/Interface/SharedXML/PixelUtil.lua

mrbuds:
This PR is not going anywhere

PR #3623, opened 2022-05-13
Smallpatches

InfusOnWoW:


WeakAurasOptions/OptionsFrames/Update.lua
```diff
@@ -1213,6 +1213,50 @@ local function ListAuras(auras, uidMap)
...
+
+    if collapsed then
+      auraLabelContainer:ReleaseChildren()
+    else
+      for _, name in ipairs(sortedNames) do
+        local auraLabel = AceGUI:Create("Label")
+        auraLabel:SetText("   • " .. name)
+        auraLabel:SetFullWidth(true)
+        auraLabelContainer:AddChild(auraLabel)
+      end
```
mrbuds, line 1253:
```suggestion
      local text
      for _, name in ipairs(sortedNames) do
        text = (text or "") .. "   • " .. name .. "\n"
      end
      if text then
        local auraLabel = AceGUI:Create("Label")
        auraLabel:SetText(text)
        auraLabel:SetFullWidth(true)
        auraLabelContainer:AddChild(auraLabel)
      end
```
This have same result but is much faster with huge list like 
image

mrbuds (approved):

PR #3640, opened 2022-05-22
"Range Check" extended to handle multiple units

Irame:
# Description

This change adds the option to select a smart group, raid, party, boss, arena or nameplates in the "Range Check" trigger.

My motivation for this change was, that I wanted to create an aura which tells me how many party or raid members are within a certain range so i don't have to guess.
Also I found it odd that the new condition for range checks allows for multi unit checks but the trigger does not.

## Type of change

## How Has This Been Tested

I did some smoke testing with nameplates where I tested all options and the glow for external elements (nameplates).
I didn't have time for testing with other unit types yet.

## Checklist

- [ ] Should one change the way the "Distance" condition-option for this trigger works? This could lead to breaking changes for existing auras.

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -102,6 +102,78 @@ function WeakAuras.UnitExistsFixed(unit, smart)
   return UnitExists(unit) or UnitGUID(unit)
 end
 
+-- similiar to GetAllUnits in BuffTrigger2
+-- this one does not support pets
+-- and has no flag to also return non existing units
+-- maybe this can be combined in the future
+function WeakAuras.GetAllUnits(unit)
+  if unit == "raid" or (unit == "group" and IsInRaid()) then
+    local i = 1
+    local max = GetNumGroupMembers()
+    return function()
+      if i <= max then
+        local ret = WeakAuras.raidUnits[i]
+        i = i + 1
+        return ret
+      end
+      i = 1
+    end
+  elseif unit == "party" or unit == "group" then
+    local i = 0
+    local max = GetNumSubgroupMembers()
+    return function()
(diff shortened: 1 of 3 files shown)


mrbuds:
> Also I found it odd that the new condition for range checks allows for multi unit checks but the trigger does not.

The reasoning behind that is you can have as much global range check condition on loaded auras at once as you want, they will use resource only when triggers combination of their respective aura is active.

If you do that with a generic trigger they will use resource even if aura is not active. It does not scale well.

Irame:
Ok, if I understand this correctly then the problem is, that the range check is very resource intensive and doing this for multiple auras on every frame could be slow, right?

Could a solution to this be to cache the results for the range check for every frame, so that wie only determine it once per frame per unit?

mrbuds:
> Could a solution to this be to cache the results for the range check for every frame, so that wie only determine it once per frame per unit?

It's a solution but i think it's a job for LibRangeCheck, it could use additional functions to handle group/nameplate/etc..

Irame:
Yes, I thought the same, that it's the job of LibRangeCheck to do that. Especially because other addons would profit from this as well.

Do you know if there is a way to contribute to this library directly, because as far as I'm aware that's a svn repository at curseforge and I don't know how to contribute there (except with issues).

mrbuds:
WeakAuras and ElvUI use this version https://github.com/WeakAuras/LibRangeCheck-2.0

Irame:
Oh sry, forgot to look into the .pkgmeta file, thanks. I'll look into that.

mrbuds:
This PR needs a major patch in LibRangeCheck prior getting further.

@Irame have not heard news from you here or on LibRangeCheck repository, i'm going to close it for now.

PR #3651, opened 2022-05-25
Custom options; Toggle, Dropdown, so forth

Bjarnebobbel:
It would be lovely if you added a new feature whenever a user has started a "custom options" with toggle or dropdown so it shows in load or shows under condition or trigger.

it is a jungle to learn how to write lua just to use a "user-friendly" custom options. 

My case:
I've created a checklist for consumables whenever im in oribos.
It lights red if i don't have any or if it's not enough for 1 dungeon

it show how many i have if i have enough for 1-2 dungeons

it does not show if i have "more than enough".

I've created different custom options toggles for food, feast and kettle.

so the user can choose which food they want to track and if they want to track kettle or feast.

as of now it tracks all the different food variables like haste, crit, mastery and vers.


InfusOnWoW:
This looks like you accidentally created a pull request instead.

PR #3660, opened 2022-05-28
Smallpatches

InfusOnWoW:


InfusOnWoW:
> could use a variable number of arguments
Did implement that

> could auto-serialize tables
Hacked until it could use the /dump infrastructure

> add a premade snippets DebugPrint (the one i use the most myself is ViragDevTool_AddData)
Didn't do this

> add weakauras version to debug console output and add wow build version to debug console output
Is now in there

> timestamp? not sure about that
> add player class/spec/location/in encounter/in group/raid or any information a user could forget to tell aura's author that could be the missing element of a bug report
That probably depens on the aura, did not do anything for that

> add buggrabber session data (if it is installed or.. ship it with weakauras?)
We should replace all our error handlers, so that they do take note of which auras created a lua error. So that we can show the lua error both on the display button and on the information tab. But I didn't do that

> show warning in chat if debug data size is getting to a gigantic level, i have no idea how to quantify that, but i can imagine people forgetting it's enable and leave it enable forever
Did that for 1000 lines

>  show debug enable warning on aurabutton
The AuraWarnings icon now shows up if debug logging is enabled, and a error icon is shown for more than 1000 lines.
Also the AuraWarnings Icon is pretty overloaded now, at some future point it needs some work.

Nightwarden24:
Could you localize aura warning messages, please?

Stanzilla:
Dump is in a load on demand part so I think you need do LoadAddon it before you can use it?

WeakAuras/DebugLog.lua
```diff
@@ -0,0 +1,97 @@
...
+function Private.DebugLog.Print(uid, text, ...)
+  if enabled[uid] then
+    local log = debugLogs[uid]
+    tinsert(log, "")
+    if select('#', ...) == 0 then
+      serialize(log, text)
+    else
+      serialize(log, text)
+      local texts = {...}
+      for i, v in ipairs(texts) do
```
mrbuds, line 47: If one of the arguments is nil furthers arguments are ignored
`DebugPrint("a", "b", nil, "c", "d")`
will add in log  `a b`

WeakAuras/DebugLog.lua
```diff
@@ -0,0 +1,97 @@
...
+        serialize(log, v)
+      end
```
mrbuds, line 49:
```suggestion
      for i = 1, select('#', ...) do
        local v = select(i, ...)
        serialize(log, v)
      end
```

Nightwarden24:
You missed fixing one error message [here](https://github.com/InfusOnWoW/WeakAuras2/blob/smallpatches/WeakAuras/RegionTypes/RegionPrototype.lua#L933)

Causese:
- opening options / closing editor seems to cause all DebugPrint to fire regardless of trigger type
- "Debug logging enabled" tooltip is not persistent in the following scenario:
1. enable debug logging
2. notice how debug logging icon appears in aura display
3. export the aura
4. delete it
5. import deleted aura
6. notice how debug logging icon is gone

InfusOnWoW:
The DebugPrints probably run because the actual code did run!

Ah, interesing that the icon does not appear in that case.

InfusOnWoW:
@Causese the question is also, do you think this is useful for you and your users?

InfusOnWoW:
Fixed the icon not appearing, some data wasn't correctly cleared on deleting a aura.

mrbuds (approved):

InfusOnWoW:
@Stanzilla it works even if the DebugTools addon is not loaded.

PR #3661, opened 2022-05-29
merge loaded/unload, add loaded filter button and deep data search

mrbuds:
# Description

Experiment changes on aura list

- Merged Loaded & Unloaded
- Added button to filter out unloaded auras
- Changed what an "unloaded aura" means for UI, it is now aura with "never", or wrong character class, race, name,  or realm
- Search can explore data, with an helper menu with premade filters

Search in data use this format:
`path:value`

Working examples:
`load.class:shaman`
`load.encounter:123`
`load.use_never:true` (need to prefix with "use_" for pure toggles
`regionType:text`
`actions.start.do_sound:true`

Idea for improvement:
- ~~support lists, something like `triggers.*.type:aura2` maybe~~ done => support triggers.*.trigger or triggers.1.trigger (works also with conditions)
- ~~handle multiple filters at once~~ done
- partial matching like `:~contains`
- ~~autocompletion?~~ added menu with premade filters

Known Issues
- with change made to what is considered loaded, opening weakauras options may show too much auras => important
- show all displays button (top green eye) doesn't account for what is filtered => important
- nested conditions are not supported
- search for a "Cast" trigger require to check both "type" and "event", when searching in all triggers with "*" can return false positive with a type=aura2 trigger with unused "event" field not cleaned up (can't do much about it)

## Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

PR #3677, opened 2022-06-12
Improve search

mrbuds:
# Description

Simplified version of https://github.com/WeakAuras/WeakAuras2/pull/3661/commits without changes made to loaded/unloaded

## Known issues / TODO:
- ~~does not handle path or values with a space character, like `trigger.*.trigger.event:Threat Situation`~~ `trigger.*.trigger.event:"Threat Situation"` works now
- load condition with namerealm check not working perfect
- no hint in UI about search operators `-` `+`

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -51,6 +51,8 @@ globals = {
 	"WeakAurasSaved",
 	"WeakAurasTimers",
 	"WeakAurasArchive",
+	"WeakAurasSearchMenuFrame",
+	"WeakAurasFilterInput",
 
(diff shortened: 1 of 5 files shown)


Stanzilla:
Do you want to handle spaces with double quotes? I think some values might have single quotes so double should be fine

mrbuds:
While it's pretty cool, it suffer a few issues
- too much dependent on internal data structure
- ask too much user _hidden_ knowledge for an effective use
- maintenance burden


It may do a come back but needs more thinking

PR #3689, opened 2022-06-22
fix array options again

emptyrivers:
i also added noMerge as a valid field in group types


emptyrivers:
there's a small performance surprise when you enable the no merging tick box on something that has a lot of things merged together, they immediately explode out into hundreds of controls and Ace chokes a bit on that. I could do some finangling to collapse the extra stuff but i figured that was too much effort.

PR #3699, opened 2022-07-16
Add a framework for built in trigger updates to fire into custom triggers

asaka-wa:
# Description

When triggers are added catch the format "TRIGGER:#" where `#` is a trigger number and flag those triggers. On trigger updates, check for the flag and send a ScanEvents into the custom trigger that requested them, along with that trigger's allstates

This needs thorough testing but I wanted to get opinions (and some more people trying to find ways to break it)

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -71,7 +71,7 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
-local TTC_events = Private.TTC_events -- trigger to custom
+local trigger_to_custom_events = Private.trigger_to_custom_events -- trigger to custom
 
 local UnitGroupRolesAssigned = WeakAuras.IsWrathOrRetail() and UnitGroupRolesAssigned or function() return "DAMAGER" end
 
@@ -1470,8 +1470,8 @@ local function UpdateTriggerState(time, id, triggernum)
 
   -- if the trigger has updated then check to see if it is flagged for trigger_to_custom and send to queue if it is
   if updated then
-    if TTC_events[id] and TTC_events[id][triggernum] then
-      Private.AddTriggerToCustomQueue(id, triggernum)
+    if trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] then
+      Private.AddToTriggerToCustomDelay(id, triggernum)
     end
   end
   return updated
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -71,7 +71,7 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
-local trigger_to_custom_events = Private.trigger_to_custom_events
+local watched_trigger_events = Private.watched_trigger_events
 
 local UnitGroupRolesAssigned = WeakAuras.IsWrathOrRetail() and UnitGroupRolesAssigned or function() return "DAMAGER" end
 
@@ -1468,10 +1468,10 @@ local function UpdateTriggerState(time, id, triggernum)
     triggerInfo.nextScheduledCheck = nil
   end
 
-  -- if the trigger has updated then check to see if it is flagged for trigger_to_custom and send to queue if it is
+  -- if the trigger has updated then check to see if it is flagged for WatchedTrigger and send to queue if it is
   if updated then
-    if trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] then
-      Private.AddToTriggerToCustomDelay(id, triggernum)
+    if watched_trigger_events[id] and watched_trigger_events[id][triggernum] then
+      Private.AddToWatchedTriggerDelay(id, triggernum)
     end
   end
   return updated
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -71,7 +71,7 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
-local trigger_to_custom_events = Private.trigger_to_custom_events -- trigger to custom
+local trigger_to_custom_events = Private.trigger_to_custom_events
 
 local UnitGroupRolesAssigned = WeakAuras.IsWrathOrRetail() and UnitGroupRolesAssigned or function() return "DAMAGER" end
 
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -1466,6 +1466,13 @@ local function UpdateTriggerState(time, id, triggernum)
     triggerInfo.nextScheduledCheck = nil
   end
 
+  -- if the trigger has updated then check to see if it is flagged for trigger_to_custom and send to queue if it is
+  if updated then
+    local data = WeakAuras.GetData(id)
+    if data.triggers[triggernum].trigger.trigger_to_custom then
+      Private.AddTriggerToCustomQueue(id, triggernum)
+    end
+  end
   return updated
 end
 
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -71,6 +71,8 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
+local TTC_events = Private.TTC_events -- trigger to custom
+
 local UnitGroupRolesAssigned = WeakAuras.IsWrathOrRetail() and UnitGroupRolesAssigned or function() return "DAMAGER" end
 
 -- keyed on unit, debuffType, spellname, with a scan object value
@@ -1468,8 +1470,7 @@ local function UpdateTriggerState(time, id, triggernum)
 
   -- if the trigger has updated then check to see if it is flagged for trigger_to_custom and send to queue if it is
   if updated then
-    local data = WeakAuras.GetData(id)
-    if data.triggers[triggernum].trigger.trigger_to_custom then
+    if TTC_events[id] and TTC_events[id][triggernum] then
       Private.AddTriggerToCustomQueue(id, triggernum)
     end
   end
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -761,15 +761,15 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
   end
 end
 
-function WeakAuras.ScanEventsInternalToCustom(id, triggernum)
+function Private.ScanEventsInternalToCustom(id, triggernum)
   Private.StartProfileAura(id);
   Private.ActivateAuraEnvironment(id);
   local updateTriggerState = false
   if trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] then
+    local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
     for requestingTrigger, bool in pairs(trigger_to_custom_events[id][triggernum]) do
       if bool then
         local data = events and events[id] and events[id][requestingTrigger]
-        local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
         local allstates = WeakAuras.GetTriggerStateForTrigger(id, requestingTrigger)
         if data and allstates and updatedTriggerStates then
           if RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER", triggernum, updatedTriggerStates) then
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -80,7 +80,7 @@ local timer = WeakAuras.timer;
 local events = {}
 local loaded_events = {}
 local loaded_unit_events = {};
-local TTC_events = Private.TTC_events
+local trigger_to_custom_events = Private.trigger_to_custom_events
 local loaded_auras = {}; -- id to bool map
 local timers = WeakAuras.timers;
 
@@ -678,9 +678,9 @@ local function RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2
       end
     end
   end
-  if updateTriggerState and TTC_events[id] and TTC_events[id][triggernum] then
+  if updateTriggerState and trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] then
     -- if this trigger's udpates are requested to be sent into one of the Aura's custom triggers
-    Private.AddTriggerToCustomQueue(id, triggernum)
+    Private.AddToTriggerToCustomDelay(id, triggernum)
   end
   return updateTriggerState;
 end
@@ -760,9 +760,35 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
     if (updateTriggerState) then
       Private.UpdatedTriggerState(id);
     end
-    Private.StopProfileAura(id);
-    Private.ActivateAuraEnvironment(nil);
+    Private.StartProfileAura(id);
+    Private.ActivateAuraEnvironment(id);
   end
+end
+
+function WeakAuras.ScanEventsInternalToCustom(id, triggernum)
+  Private.StartProfileAura(id);
+  Private.ActivateAuraEnvironment(id);
+  local data = events and events[id] and events[id][triggernum]
+  local updateTriggerState = false
+  if trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] then
+    for requestingTrigger, bool in pairs(trigger_to_custom_events[id][triggernum]) do
+      if bool then
+        local data = events and events[id] and events[id][requestingTrigger]
+        local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
+        local allstates = WeakAuras.GetTriggerStateForTrigger(id, requestingTrigger)
+        if data and allstates and updatedTriggerStates then
+          if RunTriggerFunc(allStates, data, id, requestingTrigger, "TRIGGER:"..triggernum, updatedTriggerStates) then
+            updateTriggerState = true
+          end
+        end
+      end
+    end
+  end
+  if (updateTriggerState) then
+    Private.UpdatedTriggerState(id)
+  end
+  Private.StartProfileAura(id)
+  Private.ActivateAuraEnvironment(id)
 end
 
 local function AddFakeTime(state)
@@ -1173,7 +1199,7 @@ end
 function GenericTrigger.Add(data, region)
   local id = data.id;
   events[id] = nil;
-  TTC_events[id] = nil
+  trigger_to_custom_events[id] = nil
 
   for triggernum, triggerData in ipairs(data.triggers) do
     local trigger, untrigger = triggerData.trigger, triggerData.untrigger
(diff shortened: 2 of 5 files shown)


asaka-wa:
If 2 custom triggers tracked each other, and updated their own trigger when they caught the update, then I think it would be an infinite loop so I prevented it, stopping either from tracking each other. It would be nice to show the red error message that already exists for that Event(s) box, but I don't think there's a way to see across triggers in the Options code.

The tooltip for that Event(s) box actually needs updating to tell users about CLEU:subevents and UNIT:units, so if this PR is accepted then I'll do an update to the tooltip to add TRIGGER:number at the same time.

WeakAuras/GenericTrigger.lua
```diff
@@ -1408,7 +1435,16 @@ function GenericTrigger.Add(data, region)
       end
     end
   end
-
+  -- if any of the trigger updates are requested to be sent into custom a trigger then set a flag on that trigger
+  if #trigger_to_custom > 0 then
+    for requestingTrigger, triggernums in pairs(trigger_to_custom) do
+      for triggernum, _ in pairs(triggernums) do
+        if data.triggers[triggernum] then
+          data.triggers[triggernum].trigger.trigger_to_custom = true
```
mrbuds, line 1443: aura data should not be edited

WeakAuras/WeakAuras.lua
```diff
@@ -4108,6 +4121,8 @@ function Private.UpdatedTriggerState(id)
       state.changed = false;
     end
   end
+  -- once updatedTriggerStates is complete, and empty states removed, etc., then check for queued trigger_to_custom updates
+  Private.SendQueuedTriggers()
```
mrbuds, line 4125: I don't think it is the correct place to call it, not after ApplyStatesToRegions
In Bufftrigger2 i'd want run it after `UpdateTriggerState` (before Private.UpdatedTriggerState)
In GenericTrigger, after `RunTriggerFunc`, maybe in RunTriggerFunc function at the end?

mrbuds (changes requested):
First thoughts, not sure what is best solution

WeakAuras/WeakAuras.lua
```diff
@@ -4006,6 +4006,19 @@ local function ApplyStatesToRegions(id, activeTrigger, states)
...
+-- we need the id and triggernum that's changing, but can't send the ScanEvents to the custom trigger until after UpdatedTriggerState has fired
+local trigger_to_custom_Queue = {}
+function Private.AddTriggerToCustomQueue(id, triggernum)
+  trigger_to_custom_Queue[id] = triggernum
+end
+function Private.SendQueuedTriggers()
+  for id,triggernum in pairs(trigger_to_custom_Queue) do
+    WeakAuras.ScanEvents("WA_TRIGGER_"..id.."_"..triggernum, triggerState[id][triggernum])
+  end
+  trigger_to_custom_Queue = {}
```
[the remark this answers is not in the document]
asaka-wa: 👍

WeakAuras/WeakAuras.lua
asaka-wa: I started out exactly as you say, but that would mean I'm sending the events before WA has cleared up dead states. This would require checking `state.show` to see if it's a real state or not at that point. 

What's the problem with this location?

mrbuds: The problem i see with this location is that some code will run twice or more

asaka-wa:
@mrbuds The trouble with what you describe is that it's a GenericTrigger only solution. I need to pull bufftrigger2 in also, which is quite firmly separated. I need to have the flag somewhere that can be seen in both places.

mrbuds:
> @mrbuds The trouble with what you describe is that it's a GenericTrigger only solution. I need to pull bufftrigger2 in also, which is quite firmly separated. I need to have the flag somewhere that can be seen in both places.

Having trigger_to_custom in Private namespace solve it

asaka-wa:
👍

WeakAuras/WeakAuras.lua
asaka-wa: I haven't been able to get the code to run more than required in this location. Can you think of a specific setup that would make this happen so I can test? 
This location feels right to me. I queue things when the triggers update (RunTriggerFunc and UpdateTriggerState) but wait until WA cleans up states table before sending the events. Things could be queued more than once but that won't make anything fire anyway.

WeakAuras/GenericTrigger.lua
asaka-wa: 👍

WeakAuras/GenericTrigger.lua
```diff
@@ -1346,6 +1357,20 @@ function GenericTrigger.Add(data, region)
...
+                  -- be processed once all the Aura's triggers are added
+                  local requestedTriggernum = tonumber(i)
+                  if requestedTriggernum then
+                    if trigger_to_custom[requestedTriggernum] and trigger_to_custom[requestedTriggernum][triggernum] then
+                      -- if the request is reciprocal (2 custom triggers request each other which would cause a stack overflow) then neither get added.
+                      trigger_to_custom[requestedTriggernum][triggernum] = nil
+                    elseif requestedTriggernum and requestedTriggernum ~= triggernum then
+                      tinsert(trigger_events, "WA_TRIGGER_"..id.."_"..i)
+                      trigger_to_custom[triggernum] = trigger_to_custom[triggernum] or {}
+                      trigger_to_custom[triggernum][requestedTriggernum] = true
```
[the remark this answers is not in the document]
asaka-wa: This note has taken some time to get my heard around - I've really had to get to grips with the code path for ScanEvents and ScanEventsInternal, and probably am still not entirely there. 

I see that going in to `trigger_events` with everything else is sneaking into that system which wasn't entirely designed for the purpose, but making a system for the purpose would require rebuilding so much of the same stuff, from what I can tell. The data in `events[id][triggernum]` is needed for `ScanInternalEvents`/`RunTriggerFunc` later.

Am I misunderstanding something?

WeakAuras/BuffTrigger2.lua
```diff
@@ -71,6 +71,8 @@ local timer = WeakAuras.timer
 local BuffTrigger = {}
 local triggerInfos = {}
 
+local TTC_events = Private.TTC_events -- trigger to custom
```
mrbuds, line 74: Replace local TTC* with camelCase `triggerToCustomEvents` `triggerToCustomQueue` and PascalCase for subvariables like `Private.TTC_events` => `Private.TriggerToCustomEvents`
Weakauras code is not very consistent `trigger_to_custom_events` works too, at least avoid non-descriptive variable that require to find comment explaining what it is

WeakAuras/GenericTrigger.lua
```diff
@@ -1358,17 +1359,15 @@ function GenericTrigger.Add(data, region)
...
-                  -- if the event is requesting a trigger's updated be sent in to the custo trigger then check it isn't requesting itself then add to a table to
-                  -- be processed once all the Aura's triggers are added
                   local requestedTriggernum = tonumber(i)
                   if requestedTriggernum then
-                    if trigger_to_custom[requestedTriggernum] and trigger_to_custom[requestedTriggernum][triggernum] then
+                    if TTC_events[id] and TTC_events[id][triggernum] and TTC_events[id][triggernum][requestedTriggernum] then
                       -- if the request is reciprocal (2 custom triggers request each other which would cause a stack overflow) then neither get added.
-                      trigger_to_custom[requestedTriggernum][triggernum] = nil
                     elseif requestedTriggernum and requestedTriggernum ~= triggernum then
                       tinsert(trigger_events, "WA_TRIGGER_"..id.."_"..i)
```
mrbuds, line 1367: you can remove that line

WeakAuras/WeakAuras.lua
```diff
@@ -4008,15 +4010,15 @@ end
 
 -- handle trigger updates that have been requested to be sent into custom
 -- we need the id and triggernum that's changing, but can't send the ScanEvents to the custom trigger until after UpdatedTriggerState has fired
-local trigger_to_custom_Queue = {}
+local TTC_Queue = {}
 function Private.AddTriggerToCustomQueue(id, triggernum)
-  trigger_to_custom_Queue[id] = triggernum
+  TTC_Queue[id] = triggernum
```
mrbuds, line 4015: This does not support multiple triggers in queue for same aura

WeakAuras/WeakAuras.lua
```diff
@@ -4008,15 +4010,15 @@ end
...
 end
 function Private.SendQueuedTriggers()
-  for id,triggernum in pairs(trigger_to_custom_Queue) do
+  for id,triggernum in pairs(TTC_Queue) do
     WeakAuras.ScanEvents("WA_TRIGGER_"..id.."_"..triggernum, triggerState[id][triggernum])
```
mrbuds, line 4019: I'd replace this with something like
```Lua
for id, triggers in pairs(TTC_Queue)
  for triggerDone in pairs(triggers) do
    for triggerTodo in pairs(Private.TTC_events[id][triggerDone]) do
      -- RunTriggerFunc on triggerTodo with allStates of triggerDone as parameter and TRIGGER:triggerDone as event
      WeakAuras.ScanEventsInternalFromOtherTrigger(id, triggerDone, triggerTodo)
    end
  end
end
```

WeakAuras/WeakAuras.lua
```diff
@@ -4008,15 +4010,15 @@ end
 
 -- handle trigger updates that have been requested to be sent into custom
 -- we need the id and triggernum that's changing, but can't send the ScanEvents to the custom trigger until after UpdatedTriggerState has fired
-local trigger_to_custom_Queue = {}
+local TTC_Queue = {}
```
mrbuds, line 4013: :(

mrbuds (changes requested):

WeakAuras/BuffTrigger2.lua
asaka-wa: Yeah, seemed overly long, but you're right about TTC being non-descriptive of course

WeakAuras/GenericTrigger.lua
asaka-wa: I don't think that's there any more already, not sure why it was showing. Gone now in any case.

WeakAuras/WeakAuras.lua
asaka-wa: It would never need to. Perhaps "queue" isn't quite the right word. The only intention is to delay sending to custom until the previous state is fully updated.

WeakAuras/GenericTrigger.lua
```diff
@@ -756,6 +761,31 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
...
+        local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
+        local allstates = WeakAuras.GetTriggerStateForTrigger(id, requestingTrigger)
+        if data and allstates and updatedTriggerStates then
+          if RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER:"..triggernum, updatedTriggerStates) then
+            updateTriggerState = true
+          end
+        end
+      end
+    end
+  end
```
mrbuds, line 781:
```suggestion
    for requestingTrigger in pairs(trigger_to_custom_events[id][triggernum]) do
      local data = events and events[id] and events[id][requestingTrigger]
      local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
      local allstates = WeakAuras.GetTriggerStateForTrigger(id, requestingTrigger)
      if data and allstates and updatedTriggerStates then
        if RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER:"..triggernum, updatedTriggerStates) then
          updateTriggerState = true
        end
      end
    end
  end
```

mrbuds: looks good

mrbuds (changes requested):
Good progress!

Next step is to update `trigger_to_custom_events` table in 
- GenericTrigger.UnloadDisplays
- GenericTrigger.UnloadAll
- GenericTrigger.Rename

And make sure nothing broke after an aura was unloaded then loaded

WeakAuras/GenericTrigger.lua
mrbuds: I wonder if it wouldn't be better to change trigger function signature to
```
function(allstates, event, triggerNum, statesFromTrigger)
```
aka
```
RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER", triggernum, updatedTriggerStates) 
```

asaka-wa: That info is carried in the event name. The user would make a trigger like
```LUA
function(event, triggerStates)
  if event == "TRIGGER:3" then
```
Adding the number in also certainly can't hurt, and could save people a step sometimes if they needed the value as a number, but doesn't feel vital. 

Happy to go with whatever people think will feel intuitive though.

mrbuds: IMO sending "TRIGGER" & trigger number as an extra argument is consistent with what is done with "CLEU:SPELL_CAST_SUCCESS" or "UNIT_AURA:player", 1 information per argument

asaka-wa: Oh I see now. Sure, okay.

asaka-wa:
> Good progress!
> 
> Next step is to update `trigger_to_custom_events` table in
> 
> * GenericTrigger.UnloadDisplays
> * GenericTrigger.UnloadAll
> * GenericTrigger.Rename
> 
> And make sure nothing broke after an aura was unloaded then loaded

UnloadDisplays = all good

UnloadAll - this doesn't wipe `events` only `loaded_events` and my `trigger_to_custom_events` is the equivalent of `events` not `loaded_events`

Rename = all good. 

Tested lots of renaming, adding new with the old name, deleting, recreating. Can't make anything odd happen.

mrbuds (approved):

WeakAuras/WeakAuras.lua
```diff
@@ -4006,6 +4008,19 @@ local function ApplyStatesToRegions(id, activeTrigger, states)
   end
 end
 
+-- handle trigger updates that have been requested to be sent into custom
+-- we need the id and triggernum that's changing, but can't send the ScanEvents to the custom trigger until after UpdatedTriggerState has fired
+local trigger_to_custom_delayed = {}
+function Private.AddToTriggerToCustomDelay(id, triggernum)
+  trigger_to_custom_delayed[id] = triggernum
```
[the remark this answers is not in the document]
asaka-wa: The delay is just to let WA tidy up any removed state tables. Triggers being queued would be handled and removed before any other Triggers/Auras update.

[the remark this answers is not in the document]
InfusOnWoW: No, if you have two bufftriggers, then both call AddToTriggerToCustomDelay.

But the custom trigger will only be called once, because you overwrite it

WeakAuras/WeakAuras.lua
```diff
@@ -4006,6 +4008,19 @@ local function ApplyStatesToRegions(id, activeTrigger, states)
...
+end
+function Private.SendDelayedTriggerToCustom()
```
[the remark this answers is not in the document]
mrbuds: MetaTrigger maybe?

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
asaka-wa: I see. This is specific to the way bufftrigger process updates and/or where I queued things from in bufftrigger2?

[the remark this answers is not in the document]
InfusOnWoW: No, if you look at UpdateStates in BT2, that iterates over all states before calling UpdatedTriggerState
But that's also the case for e.g. WeakAuras.ScanUnitEvents/WeakAuras.ScanEventsInternal

They all handle all triggers for a given id, before calling UpdatedTriggerState

WeakAuras/GenericTrigger.lua
```diff
@@ -756,6 +761,31 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
...
+        local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum)
+        local allstates = WeakAuras.GetTriggerStateForTrigger(id, requestingTrigger)
+        if data and allstates and updatedTriggerStates then
+          if RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER", triggernum, updatedTriggerStates) then
+            updateTriggerState = true
+          end
+        end
+      end
+    end
+  end
```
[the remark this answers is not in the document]
asaka-wa: The check for whether a trigger's updates is requested is in RunTrigerFunc which is called in ScanEventsInternalToCustom. So to my understanding of the code path, and testing in game, I think it does.

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
asaka-wa: Happy to go and update everywhere if people are happy with "MetaTrigger"

WeakAuras/GenericTrigger.lua
```diff
@@ -1346,6 +1385,23 @@ function GenericTrigger.Add(data, region)
 
                   trigger_unit_events[i] = trigger_unit_events[i] or {}
                   tinsert(trigger_unit_events[i], trueEvent)
+                elseif isTrigger then
+                  local requestedTriggernum = tonumber(i)
+                  if requestedTriggernum then
+                    if trigger_to_custom_events[id] and trigger_to_custom_events[id][triggernum] and trigger_to_custom_events[id][triggernum][requestedTriggernum] then
```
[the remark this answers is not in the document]
asaka-wa: Yes, I struggled with this. I can't see a way to process this issue inside only WeakAurasOptions, which AFAIK would be necessary. 
My solution was going to be to mention it in the tooltip, that reciprocal requests will be ignored.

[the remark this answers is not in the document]
InfusOnWoW: There's already a warning for certain unknown unit events (iirc), so it should be the same.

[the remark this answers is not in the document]
InfusOnWoW: That is the options code already parses the event list (iirc).

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
InfusOnWoW: MetaTrigger doesn't sound right either, but as long as there's some cohesion in naming, it's probably fine.

WeakAuras/GenericTrigger.lua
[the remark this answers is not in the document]
InfusOnWoW: Ah indee.

WeakAuras/GenericTrigger.lua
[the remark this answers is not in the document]
asaka-wa: My difficulty would be having to look across different triggers. Seems like the triggerinfo available in the options is only the current one?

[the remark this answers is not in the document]
InfusOnWoW: trigger_to_custom_events should be up to date, since any change to the event field calls WeakAuras.Add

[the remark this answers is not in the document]
asaka-wa: 👍

WeakAuras/WeakAuras.lua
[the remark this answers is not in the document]
asaka-wa: WatchedTriggers?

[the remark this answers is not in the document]
asaka-wa: ("Watched" is only used for the Faction trigger currently and "WatchedTrigger" appears nowhere)

asaka-wa:
Updated based on feedback.

InfusOnWoW:
@asaka-wa congratulations

WeakAuras/GenericTrigger.lua
```diff
@@ -761,18 +761,19 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
...
+
+  for _, wathcedTrigger in ipairs(watchedTriggernums) do
+    if watched_trigger_events[id] and watched_trigger_events[id][wathcedTrigger] then
+      local updatedTriggerStates = WeakAuras.GetTriggerStateForTrigger(id, wathcedTrigger)
+      for observerTrigger in pairs(watched_trigger_events[id][wathcedTrigger]) do
+        local data = events and events[id] and events[id][observerTrigger]
+        local allstates = WeakAuras.GetTriggerStateForTrigger(id, observerTrigger)
         if data and allstates and updatedTriggerStates then
-          if RunTriggerFunc(allstates, data, id, requestingTrigger, "TRIGGER", triggernum, updatedTriggerStates) then
+          if RunTriggerFunc(allstates, data, id, observerTrigger, "TRIGGER", wathcedTrigger, updatedTriggerStates) then
```
mrbuds, line 776: wathcedTrigger

InfusOnWoW:
Two random thoughts that I wanted to write down.

a) I did consider whether making triggers "hierarchial" would be a better approach. And while that feels more clean, its much more complex for e.g. conditions the planned visibility overhaul, text replacements. And it's actually less powerful, since you often want to combine the information from the original trigger and the custom trigger.

b) We can eventually maybe use this for e.g. the "range checking" trigger. That is instead of the unit combobox, you set the trigger up to "watch another trigger's state", and the range checker automatically creates states that contain the range for the corresponding units of the main trigger. To be truely useful, the visibility would require the rewrite that I hinted at several times already. But that does sound like a good approach for something like the range trigger.

PR #3703, opened 2022-07-17
support ace event messages

charxene:
# Description

allow AceEvent message types to be specified with `AceEvent:` prefix any place a regular event is valid 

## Type of change

## How Has This Been Tested

- [x] made triggers and conditions for events from grid2 

## Checklist

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -84,7 +84,7 @@ local loaded_auras = {}; -- id to bool map
 local timers = WeakAuras.timers;
 
 -- Local functions
-local LoadEvent, HandleEvent, HandleUnitEvent, TestForTriState, TestForToggle, TestForLongString, TestForMultiSelect
+local LoadEvent, HandleEvent, HandleMessage, HandleUnitEvent, TestForTriState, TestForToggle, TestForLongString, TestForMultiSelect
 local ConstructTest, ConstructFunction
 
 
@@ -836,6 +836,10 @@ function GenericTrigger.ScanWithFakeEvent(id, fake)
   Private.ActivateAuraEnvironment(nil);
 end
 
+function HandleMessage(frame, event, ...) 
+  HandleEvent(frame, "AceEvent:" .. event, ...)
+end
+
 function HandleEvent(frame, event, arg1, arg2, ...)
   Private.StartProfileSystem("generictrigger " .. event);
   if event == "NAME_PLATE_UNIT_ADDED" then
@@ -1068,6 +1072,19 @@ local function trueFunction()
   return true;
 end
 
+local function parseAceEventMessageType(event) 
+  local eventType
+  for i in string.gmatch(event, "[^:]+") do
+    if not eventType then
+      eventType = i
+    elseif eventType == "AceEvent" then
+      return i
+    else
+      return
+    end
+  end
+end
+
 local eventsToRegister = {};
 local unitEventsToRegister = {};
 function GenericTrigger.LoadDisplays(toLoad, loadEvent, ...)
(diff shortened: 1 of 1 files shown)


WeakAuras/GenericTrigger.lua
```diff
@@ -1121,8 +1138,14 @@ function GenericTrigger.LoadDisplays(toLoad, loadEvent, ...)
   end
 
   for event in pairs(eventsToRegister) do
-    xpcall(frame.RegisterEvent, trueFunction, frame, event)
-    genericTriggerRegisteredEvents[event] = true;
+    local aceEventMessage = parseAceEventMessageType(event)
+    if aceEventMessage then
+      aceEvents:RegisterMessage(aceEventMessage, HandleMessage, frame)
+      genericTriggerRegisteredEvents[event] = true;
```
charxene, line 1144: should I try to unregister messages somewhere? doesn't seem like that ever happens for the regular events.

InfusOnWoW:
What is the use case for this?

WeakAuras/GenericTrigger.lua
InfusOnWoW: We don't track that in any place yet.

WeakAuras/GenericTrigger.lua
```diff
@@ -1068,6 +1072,19 @@ local function trueFunction()
   return true;
 end
 
+local function parseAceEventMessageType(event) 
+  local eventType
+  for i in string.gmatch(event, "[^:]+") do
```
InfusOnWoW, line 1077: You get passed in a single event if I understood that code correctly.

InfusOnWoW (changes requested):

WeakAuras/GenericTrigger.lua
npgm: the event could be: `AceEvent:TestEvent`
split = ["AceEvent", "TestEvent"]
if split[0] == AceEvent -> return split[1]
otherwise return null

charxene: its one event but need to test if it has the aceevent prefix

charxene:
use case is to allow auras to listen to events from other addons. its essentially the same use case as listening to blizzard events, but now extended to any event generated by any addon.

WeakAuras/GenericTrigger.lua
InfusOnWoW: gmatch is the wrong function for it, you want match then

charxene: I want to get the event without the prefix though - its what i'm returning

InfusOnWoW:
What kind of events of other addons?

Note, that this is easily done in custom code, yet I haven't seen any custom coded auras that have needed.

WeakAuras/GenericTrigger.lua
InfusOnWoW: match can do pattern matching and do that for you

charxene:
how would this be done in custom code?

mrbuds:
> use case is to allow auras to listen to events from other addons. its essentially the same use case as listening to blizzard events, but now extended to any event generated by any addon.

Common way to push events to other addons is thru CallbackHandler-1.0, then it's easy to make it work in custom code
https://github.com/WeakAuras/WeakAuras2/pull/3684#issuecomment-1159488285 for reference

charxene:
@mrbuds i guess thats possible, but I think a lot of things that are done in triggers and actions could be implemented that way. why would I want to write my own event ingestion like that instead of just having support for any messages as events?

WeakAuras/GenericTrigger.lua
```diff
@@ -1068,6 +1072,19 @@ local function trueFunction()
...
+    if not eventType then
+      eventType = i
+    elseif eventType == "AceEvent" then
+      return i
+    else
+      return
+    end
+  end
```
charxene, line 1085:
```suggestion
  return string.match("AceEvent:(.*)", event)
```

InfusOnWoW:
That's not a good enough reason for us to add it. It needs to add value to WA, because we have to maintain, document and support it after wards.

You still haven't really told us why you think it's worth it.

charxene:
ok - i'll just use custom

InfusOnWoW:
So you can't tell us what addon it is?

charxene:
@InfusOnWoW grid2 - i wanted to listen to enter / exit mouse events to approximate tooltips with a mouse follow aura that only applies to grid2 (so the mouseover unit update isn't sufficient)

PR #3704, opened 2022-07-17
"Faction Reputation" trigger: handle paragons

### LINKED ISSUE #3698: Add Paragon Reputation to Reputation Trigger
The headline basically sums it all up. 
If i'm not completely blind there is currently no way to track paragon reputation within the default reputation trigger, so i'm requesting it as a new feature :)

mrbuds:
# Description

Fixes #3698


WeakAuras/Prototypes.lua
```diff
@@ -2191,11 +2191,20 @@ Private.event_prototypes = {
...
-            standing = friendTextLevel
+          if factionID then
+            local _, _, _, _, _, _, friendTextLevel = GetFriendshipReputation(factionID)
+            if friendTextLevel then
+              standing = friendTextLevel
+            end
+            local friendshipRank, friendshipMaxRank = GetFriendshipReputationRanks(factionID)
+            local currentValue, threshold, rewardQuestID, hasRewardPending, tooLowLevelForParagon  = C_Reputation.GetFactionParagonInfo(factionID)
+            if not tooLowLevelForParagon and currentValue and threshold then
+              local level = math.floor(currentValue/threshold)-(hasRewardPending and 1 or 0)
```
InfusOnWoW, line 2202: Is that based on how blizzard does it?

mrbuds: I used example from wiki https://wowpedia.fandom.com/wiki/API_C_Reputation.GetFactionParagonInfo but indeed blizzard code is better, changed
Also changed friendship reputation values like blizzard does

PR #3706, opened 2022-07-21
Fixes loss of mouse control in code editbox

Grimonja:
## Description
Fixes a dreaded problem where script editbox stops behaving.
Resulting in inability to select any text with the mouse cursor.

## Type of change

## How Has This Been Tested
To replicate this bug you do the following:

1. Open a WA code field that has a longish editbox
2. Scroll down that editbox
3. Close the editbox
4. Reopen that same code field but don't scroll
5. Open any code field that DOESENT have the scrollbar enabled
6. ???
7. Profit

Root cause of the problem is that in some cases HitRectInsets is not reset.

### DIFF
--- a/WeakAurasOptions/OptionsFrames/TextEditor.lua
+++ b/WeakAurasOptions/OptionsFrames/TextEditor.lua
@@ -188,9 +188,8 @@ local function ConstructTextEditor(frame)
   -- Fixes a ACEGUI issue where HitRectInsets are not reset in some cases
   -- Causing inability to use the mouse to select any text.
   editor.editBox:HookScript("OnHide",function(self)
-    self:SetHitRectInsets(0,0,0,0);
-  end);
-
+    self:SetHitRectInsets(0,0,0,0)
+  end)
 
   -- The indention lib overrides GetText, but for the line number
   -- display we ned the original, so save it here.
--- a/WeakAurasOptions/OptionsFrames/TextEditor.lua
+++ b/WeakAurasOptions/OptionsFrames/TextEditor.lua
@@ -185,6 +185,13 @@ local function ConstructTextEditor(frame)
     originalOnCursorChanged(self, ...)
   end)
 
+  -- Fixes a ACEGUI issue where HitRectInsets are not reset in some cases
+  -- Causing inability to use the mouse to select any text.
+  editor.editBox:HookScript("OnHide",function(self)
+    self:SetHitRectInsets(0,0,0,0);
+  end);
+
+
   -- The indention lib overrides GetText, but for the line number
   -- display we ned the original, so save it here.
   local originalGetText = editor.editBox.GetText


Stanzilla:
Thanks! Can you remove the semicolons and one of the new lines at the end of the function?

Grimonja:
As requested

InfusOnWoW:
I don't quite understand why that is necessary, but from the comment that sounds like a upstream issue?

mrbuds:
We need better reproduction steps, i can't reproduce the issue https://images-ext-1.discordapp.net/external/RHwp8ho7QoxnLL-5j7bnYI0gUw2YIWiPnVNirz5lfjE/https/i.imgur.com/WLa0shf.mp4

Grimonja:
Technically this is an issue with ACEGui MultiLineEditBox widget.
Fixing this in ACE is much more involved than this.

I would argue that this is similar case to this: 
https://github.com/WeakAuras/WeakAuras2/blob/main/WeakAurasOptions/OptionsFrames/TextEditor.lua#L176-L186

Grimonja:
> We need better reproduction steps, i can't reproduce the issue https://images-ext-1.discordapp.net/external/RHwp8ho7QoxnLL-5j7bnYI0gUw2YIWiPnVNirz5lfjE/https/i.imgur.com/WLa0shf.mp4

https://dl.dropboxusercontent.com/s/5bh1wg2tej7ak6l/Wow_eYnzImGn0x.mp4

Stanzilla:
@Nevcairiel thoughts?

WeakAurasOptions/OptionsFrames/TextEditor.lua
```diff
@@ -185,6 +185,12 @@ local function ConstructTextEditor(frame)
     originalOnCursorChanged(self, ...)
   end)
 
+  -- Fixes a ACEGUI issue where HitRectInsets are not reset in some cases
+  -- Causing inability to use the mouse to select any text.
+  editor.editBox:HookScript("OnHide",function(self)
+    self:SetHitRectInsets(0,0,0,0)
+  end)
+
```
StanzillaManticore, line 193:
```suggestion
  -- Fixes a AceGUI issue where HitRectInsets are not reset in some cases
  -- Causing inability to use the mouse to select any text.
  editor.editBox:HookScript("OnHide", function(self)
    self:SetHitRectInsets(0, 0, 0, 0)
  end)

```

StanzillaManticore (changes requested):

Nevcairiel:
This should be resolved in Ace3 r1266, please me know if the issue persists in some circumstances.

InfusOnWoW:
[acegui.txt](https://github.com/WeakAuras/WeakAuras2/files/9175478/acegui.txt)

Thanks nev. This is the change in AceGui if anyone wants to try it out.

Grimonja:
This sadly wont really do anything, as WAs use the same frame over the whole addon lifetime.
Meaning "OnAcquire" will not fire

Nevcairiel:
I've applied a different change to properly reset the HitRect when the scroll frame changes, hopefully that also works with persistent widgets

Grimonja:
This should fix the issue although it does introduce some behavior that I don't really understand.

During scrolling/moving of the frame ``OnScrollRangeChanged`` fires sometimes for unknown reasons. In some cases it is firing every frame for some reason, but only on certain widget sizes.

https://dl.dropboxusercontent.com/s/m4y135kbo4zm996/Wow_3JdxAWxmc1.mp4

I'm not really sure if this is a problem but there it is.

Nevcairiel:
There is probably some inaccuracy with word wrapping that sometimes oscillates the size of the text box, but that should not be a new issue, as the HitRect should not influence that.

PR #3710, opened 2022-07-25
Wrath

mrbuds:
# Description

- Vehicule API of Retail works on Wrath
- Added helper functions for version check to get rid of boolean logic nightmare everywhere, and make IsBCC not include Wrath (easier change if WOW_PROJECT_ID get changed for Wrath)


WeakAuras/Init.lua
```diff
@@ -56,6 +56,14 @@ function WeakAuras.IsRetail()
   return WOW_PROJECT_ID == WOW_PROJECT_MAINLINE
 end
 
+function WeakAuras.IsBCCAndNotWrath()
+  return WeakAuras.IsBCC() and not WeakAuras.IsWrathClassic()
```
InfusOnWoW: I would expect IsBCC to already not include Wrath, why doesn't that work?

mrbuds: left over code i'll delete this function (not use anywhere)

mrbuds: ha no it's changed next commit

mrbuds: WOW_PROJECT_ID == WOW_PROJECT_BURNING_CRUSADE_CLASSIC on Wrath

Adding this function on vehicle commit made me realize the next commit that change all version checks had to be done

Stanzilla:
I'm not sure if I'm the biggest fan of that logic spaghetti but if it makes the most sense right now, I'm okay with it :D

mrbuds:
> I'm not sure if I'm the biggest fan of that logic spaghetti but if it makes the most sense right now, I'm okay with it :D

There is a bit of spaghetti in Init.lua, but it make all other version test in other files more clean

PR #3715, opened 2022-07-26
Add a "Never Loaded" state to the Loaded Indicator

asaka-wa:
# Description

Adds a red indicator to the current Yellow (loaded) and Grey (not currently loaded) for when an Aura is set to Never. 

The main use case for this would likely be people not realising why something never works. Mostly it's just extending the information carried by that Loaded icon which is currently a bit overlooked. 
https://i.imgur.com/RKBtDBv.png

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist


Stanzilla:
I think that is a great small addition

InfusOnWoW:
We did discuss that in the discord and I don't think this is worth it:

* Everyone agrees that the loaded indicator is pretty bad, and this builts on top of it. Even emphasizes it.
* Never load is not a error, so using a red color is pretty bad.
* Never load is not that important, that it needs to be highlighted there. 

I'd rather see a more comprehensive try at improving the loaded indicator than this

asaka-wa:
For me, easily seeing if an Aura is set to Never is quite useful. 

Consider a large group of Auras that you've imported. Some are experimental so you're using "Never" to swap them in or out in between pulls. Without being able to see this easily I've resorted to things like adding "-never" to the name - which inevitably gets forgotten at some point leading to confusion. 

I understand that this is fairly niche. The less niche use case is just people who are newer to the WA config not understanding why something isn't working and not chasing it down to the Load tab. 

So neither is a major issue facing the WA userbase, but nor is this a major change. Personally I think that it makes something that's not currently useful to me, have some small usefulness. I am aware that the current iteration of the icon has some usefulness to others though, it gets discussed occasionally on the Discord.

InfusOnWoW:
Then the problem is:
You designed a solution for your niche, and now say it's useful for a related problem.

Users are indeed sometimes confused on why something doesn't show up, but "never load" is just one variant.
There's also, the user expected something to show, but the aura was set to only load in combat/dungeon/m+/affix/encounter etc.

That is - without question - much harder problem! Currently the indicator has a very low information value, that's why it is so ignorable. As I said on discord, it probably should be hidden by default and only show "exceptions". Obviously one such exception could be "never loaded". But others are probably worth highlighting too. And it needs something different than a red colored icon, that looks like some kind of warning.

asaka-wa:
I think you're letting the "perfect be the enemy of the good". If the red icon is the main argument against - aside from a deeper look into the concept of the icon being broadly needed at some point - then can that not be discussed and iterated upon?

InfusOnWoW:
Well you did indicate that you didn't want to iterate on it.

In this state, it's a step in the wrong direction. It adds visual complexity to badly solve a niche problem.

InfusOnWoW:
The visual isn't the only problem. I thought I already explained that this is simply the wrong approach at development. Solving one niche problem at a time is the wrong approach. It can't lead to a good result. I'd rather have the current state, than a state where "never loaded" is highlighted.

PR #3720, opened 2022-07-30
Smallpatches

InfusOnWoW:


WeakAuras/WeakAuras.toc
```diff
@@ -50,6 +54,7 @@ Animations.lua
 Conditions.lua
 
 # Trigger systems
+LibSpecializationWrapper.lua
```
mrbuds, line 57: LibSpecializationWrapper.lua should only be loaded on retail

otherwise `1x WeakAuras\LibSpecializationWrapper.lua:4: Cannot find a library instance of "LibSpecialization".`

WeakAuras/BuffTrigger2.lua
```diff
@@ -1927,6 +1933,17 @@ local function EventHandler(frame, event, arg1, arg2, ...)
   Private.StopProfileSystem("bufftrigger2")
 end
 
+Private.LibSpecWrapper.Register(function(unit)
```
mrbuds, line 1935: need to check WeakAuras.IsRetail()

WeakAuras/GenericTrigger.lua
```diff
@@ -3526,6 +3526,13 @@ function WeakAuras.RegisterItemCountWatch()
   end
 end
 
+-- LibSpecWrapper
+-- We always register, because it's probably not that often called, and ScanEvents checks
+-- early if anyone wants the event
+Private.LibSpecWrapper.Register(function(unit)
```
mrbuds, line 3532: need to check WeakAuras.IsRetail()

WeakAuras/WeakAuras.toc
```diff
@@ -33,6 +33,10 @@ embeds.xml
 # #@end-version-classic@
 #@end-non-version-retail@
 
+#@version-retail@
+# Libs\LibSpecialization\LibSpecialization.lua
+#@end-version-retail@
```
mrbuds, line 38: This break using non-packaged version on retail, and if un-commented it break non-packaged version on other versions. That would be annoying for dev (currently work on bcc/wrath).

I think the multitoc PR has to be resurrected first.

WeakAuras/WeakAuras.toc
mrbuds, line 57: Need to check for retail, in toc file or in lua

mrbuds (changes requested):

WeakAuras/Prototypes.lua
```diff
@@ -3572,6 +3633,19 @@ Private.event_prototypes = {
...
+        display = L["Specialization"],
+        type = "multiselect",
+        init = "WeakAuras.SpecForUnit(unit)",
+        values = "spec_types_all",
+        store = true,
+        conditionType = "select",
+        enable = function(trigger)
+          return WeakAuras.IsRetail() and (trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party")
+        end,
+        desc = L["Requires syncíng the specialization via LibSpecialization."],
```
mrbuds, line 3647: minor typo in all 3 desc fields: syncíng => syncing

WeakAuras/LibSpecializationWrapper.lua
```diff
@@ -0,0 +1,85 @@
...
+
+Private.LibSpecWrapper = {}
+if LibSpec then
+  local frame = CreateFrame("Frame")
+  frame:RegisterEvent("PLAYER_LOGIN")
+  frame:RegisterEvent("GROUP_ROSTER_UPDATE")
+  frame:SetScript("OnEvent", function()
+    local ownName = GetUnitName("player", true)
+
+    nameToUnitMap = {}
```
mrbuds, line 22: shouldn't it use wipe(nameToUnitMap) for less garbage?

InfusOnWoW: Doesn't really matter, because this happens so infrequently.

Using a new table each time does show up in the memory consumption, but is likely actually faster than wiping the table.

mrbuds (approved):

PR #3730, opened 2022-08-10
Wrath Talent load condition rework

mrbuds:
# Description
It work only on Wrath, (no classic, tbc, shadowlands, and a version of this widget will be done for dragonflight

This PR add an ace3 widget to use as control for "multiselect" and add a few options to multiselect for load conditions
- arg.multiNoSingle => skip single mode, only disable or multi
- arg.multiTristate => values of "multi" can be false, value for test parsed is 4 for true, 5 for false (i think i can simplify that now that there won't be any backward compability)
- arg.multiAll => all tests on values of the multiselect are grouped with "and" instead of "or"
- arg.orConjonctionGroup => each load condition with same value will be tested together with "or" instead of "and"

img

preview https://i.imgur.com/clkuuLV.gif


InfusOnWoW:
So the feedback on that was a bit mixed in discord, but I personally like these kind of ideas by you. But yeah it takes up too much space. It maybe needs to be a flyout, something where you select the talents and once you selected, it doesn't take more than one row.

And as I mentioned in Discord;
* we have to consider backwards compatibility, although we can break it in SL->DF. 
* And we have to think a bit about how talent selection works in DF. In SL you often had cases where you either selected active ability X or Y, and depending on which one was selected a different aura was loaded. I kinda guess, that doesn't work to well in DF. 
* And the current load options aren't that good. They are essentially a AND combination of OR talent selections. 

I think if people want more complex talent load options, then they think more in OR combination of AND selections. Each AND selection is a build, and you would want to setup a WA to be loaded with several different builds.

Anyway, it's an interesting experiment.

The commit to split up the Types.lua file per expansion is a good idea, if you want make a separate PR for it.

mrbuds:
> we have to consider backwards compatibility, although we can break it in SL->DF.

My plan for each version
- Classic/TBC => no change
- Wrath => get it merged before pre-patch (date not announced yet), no backward compatibility
- Shadowland => no change, scrap what has been done, it was nice for training
- DF => no backward compatibility

> depending on which one was selected a different aura was loaded. I kinda guess, that doesn't work to well in DF.

I don't see how it's different from classic/tbc/wrath, i fail to see what wouldn't work

> And the current load options aren't that good. They are essentially a AND combination of OR talent selections.

100% agree

mrbuds:
Shadowlands changes are reverted now, only Wrath version is affected by this PR

> I think if people want more complex talent load options, then they think more in OR combination of AND selections. Each AND selection is a build, and you would want to setup a WA to be loaded with several different builds.

This is implemented for Wrath, talents selected in a tree are "AND" and the "And Talent" is now "Or Talent"
Load conditions that should be grouped with "or" instead of "and" have to use `arg.orConjonctionGroup` set with same value

If you have a better name than `orConjonctionGroup` i take it!

What's left TODO is to "roll back" to 1st version of the widget where selected talents were shown in an InlineGroup, and show talent frame when a button is clicked, or group mouseover, or something else, not sure yet what is best.

mrbuds:
There is currently (that i know of) one regression concerning load conditions with arg.control = "WeakAurasSortedDropdown" (`Instance Size Type` and `Instance Type`) because arg.control was not considered when use_[name] == false with "multiselect"

Before
image

After
image

mrbuds:
> There is currently (that i know of) one regression concerning load conditions with arg.control = "WeakAurasSortedDropdown" (`Instance Size Type` and `Instance Type`) because arg.control was not considered when use_[name] == false with "multiselect" and now it is

fixed

Stanzilla:
My problem with this is that I'm not really convinced it's going to be a good UX for DF. Because if it isn't then we only have this setup for Wrath, which would be a bit odd

mrbuds:
DF talents UI is pretty close to Wrath one, differences are :

- view is split in 2 large panels, can do:
  -  vertical layout instead of horizontal
  -  use tabs, or button to switch between them
  -  open a modal window for larger scaling
- dual talent nodes, i don't see it as a problem

This PR change logic between talents selected (and/or/not, talent1, talent2, ..) for the better, and does it only for Wrath for now (it's planned for DF too).
This change in logic is too complex to migrate, that's why i'd like if it can be merge before Wrath pre-patch.

DF version will need work, if for example it use a modal window it can be ported to Wrath version later without breaking anything.

InfusOnWoW:
Select Talent is cut of for me.
WowClassicB_fw9dwudZYx

InfusOnWoW:
The hover and selected don't overlap, which looks wrong. The hover is too wide.


WowClassicB_OPjAWHzzgZ

mrbuds:
Tuned talent buttons to be more clean when hover

For "Select Talent" button, i can reproduce issue if i change set big uiscale, /reload, then open weakauras options and lower uiscale :alot:

image
image

but after a /reload text is fixed

image

InfusOnWoW:
Nothing major, I think this looks pretty good.

I did indeed change the ui scale a bunch, because it was the first time on woltk. So don't bother with figuring that out for now, it's not critical.

mrbuds:
Fixed comments, removed unused Retail widget, and applied StyLua on wrath widget

InfusOnWoW:
@Stanzilla I think this looks good now and is a great improvement compared to the current behaviour.

We'll figure out how it'll work for DF, once there's DF.

Stanzilla:
Alright, let's roll with it then

PR #3760, opened 2022-08-21
Switch escape over to needing a modifier down to close without saving

asaka-wa:
# Description

Prevent losing work

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/WeakAurasOptions/OptionsFrames/TextEditor.lua
+++ b/WeakAurasOptions/OptionsFrames/TextEditor.lua
@@ -522,14 +522,14 @@ local function ConstructTextEditor(frame)
       end
   )
 
-  -- CTRL + S saves and closes, ESC cancels and closes
+  -- CTRL + S saves and closes, ANYMOD + ESC cancels and closes
   editor.editBox:HookScript(
     "OnKeyDown",
     function(_, key)
       if IsControlKeyDown() and key == "S" then
         group:Close()
       end
-      if key == "ESCAPE" then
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/OptionsFrames/TextEditor.lua
```diff
@@ -522,14 +522,14 @@ local function ConstructTextEditor(frame)
...
-  -- CTRL + S saves and closes, ESC cancels and closes
+  -- CTRL + S saves and closes, ANYMOD + ESC cancels and closes
   editor.editBox:HookScript(
     "OnKeyDown",
     function(_, key)
       if IsControlKeyDown() and key == "S" then
         group:Close()
       end
-      if key == "ESCAPE" then
+      if IsModifierKeyDown() and key == "ESCAPE" then
```
InfusOnWoW, line 532: Just remove it, having a modifier is non intuitive and wouldn't make the ratio between actually intended and unintended usages any better.

asaka-wa: Fair enough. I thought it would be awkward enough to use a modifier with escape that it wouldn't happen by accident while giving people the option if they need it. But I'm certainly not going to put up a fight because I hate it :D

InfusOnWoW:
Yes the accidental usage would be 0. But the intended usage wouldn't be much higher, at that point it's not worth keeping.

PR #3785, opened 2022-08-29
Smallfix

mrbuds:


WeakAuras/BuffTrigger2.lua
```diff
@@ -1129,8 +1129,10 @@ local function TriggerInfoApplies(triggerInfo, unit)
     return false
   end
 
-  if triggerInfo.raidRole and not triggerInfo.raidRole[WeakAuras.UnitRaidRole(controllingUnit) or ""] then
-    return false
+  if WeakAuras.IsClassicOrBCCOrWrath() then
+    if triggerInfo.raidRole and not triggerInfo.raidRole[WeakAuras.UnitRaidRole(controllingUnit) or ""] then
```
InfusOnWoW, line 1133: I think it's better to make effectiveRaidRole nil on Retail in BuffTrigger.Add line 2595

mrbuds: fixed in https://github.com/WeakAuras/WeakAuras2/pull/3785/commits/db9189b38104e0ccf6257456c8a5150f3d9367e5 where more fields are tested

InfusOnWoW:
I renamed initTest to "enableTest", because that felt like a better name for what the function does.

PR #3809, opened 2022-09-11
Switch to StyLua code style

Stanzilla:
# Description

Second try, this time forcing the max width to a number we never reach so the hanging of argument is the same. Also sprinkled in some `-- stylua: ignore` comments for places where we (probably) want manual formatting.

## Type of change

## How Has This Been Tested

- [ ] Test A
- [ ] Test B

## Checklist


WeakAuras/Conditions.lua
```diff
@@ -384,110 +380,116 @@ local function ParseProperty(property)
...
+            end
+          elseif propertyData.action then
+            local pathToCustomFunction = "nil"
             local pathToFormatter = "nil"
-            if (WeakAuras.customConditionsFunctions[id]
-              and WeakAuras.customConditionsFunctions[id][conditionNumber]
-              and  WeakAuras.customConditionsFunctions[id][conditionNumber].changes
-              and WeakAuras.customConditionsFunctions[id][conditionNumber].changes[changeNum]) then
-              pathToCustomFunction = string.format("WeakAuras.customConditionsFunctions[%q][%s].changes[%s]", id, conditionNumber, changeNum);
+            if WeakAuras.customConditionsFunctions[id] and WeakAuras.customConditionsFunctions[id][conditionNumber] and WeakAuras.customConditionsFunctions[id][conditionNumber].changes and WeakAuras.customConditionsFunctions[id][conditionNumber].changes[changeNum] then
```
InfusOnWoW, line 446: the problem with setting the line width to 1000 is that stylua thinks this is what we want.

Obviously we gain a bit of consistency by removing all the semicolons, fixing the indentation and spaces around operators.

But this and other lines that are now exessively long are not an improvement.

Stanzilla: Yeah but that's when we can use the inline ignore, doesn't happen too often?

InfusOnWoW: Well, that's just one of many examples. And frankly the gain is just not there, it fixes a few small things everywhere, but looking at what it does, I haven't found a clear example of where the existing formatting was actually unreadable.

As in, I haven't seen anywhere where the indentation was actually misleading.

I'd much rather have a tool that respects the developer instead of this by design user hostile tool.

PR #3816, opened 2022-09-14
Add "Action Button" as aura anchor & glow external option

mrbuds:
# Description

Add "Action Button" as aura anchor & glow external option

Require LibGetFrame-1.0 last version https://github.com/mrbuds/LibGetFrame

PS: No dynamic group anchoring, i don't think this option is relevant for dynamic groups.

## Type of change

## How Has This Been Tested

It was tested with these triggers

- Cast
- Aura
- Combat Log
- Action Usable
- Cooldown Progress (spell)
- Cooldown Progress (item)
- Item Count

## Know Issues
~~External glow may not work on an aura already active after PEW or /reload~~
~~Moving a button on action bars after an aura has been anchored to a button, or a glow active, won't update the glow or anchor.~~
~~I'll have to add a callback tracking action button changes~~

### DIFF
--- a/WeakAuras/AuraEnvironment.lua
+++ b/WeakAuras/AuraEnvironment.lua
@@ -140,6 +140,7 @@ WeakAuras.GetUnitNameplate =  function(unit)
     return LGF.GetUnitNameplate(unit)
   end
 end
+WeakAuras.GetActionButtonsById = LGF.GetActionButtonsById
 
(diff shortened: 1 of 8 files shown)


mrbuds:
This PR isn't going anywhere, there are corner cases that ask to do too much computing to make it 100% reliable, like handling stances, pages, vehicles ..

PR #3869, opened 2022-10-09
Add a way for auras to save data

InfusOnWoW:


InfusOnWoW:
Test aura https://wago.io/C6y3TVTq7

WeakAuras/AuraEnvironment.lua
```diff
@@ -10,6 +10,10 @@ if WeakAuras.IsClassic() then
   LCD:RegisterFrame("WeakAuras")
 end
 
+local LibSerialize = LibStub("LibSerialize")
+local LibDeflate = LibStub:GetLibrary("LibDeflate")
+local Serializer = LibStub:GetLibrary("AceSerializer-3.0")
```
mrbuds, line 15: Serializer is not used

WeakAuras/AuraEnvironment.lua
```diff
@@ -388,7 +455,10 @@ local FakeWeakAurasMixin = {
   },
   override = {
     me = GetUnitName("player", true),
-    myGUID = UnitGUID("player")
+    myGUID = UnitGUID("player"),
+    GetData = function(id)
+      return CopyTable(WeakAuras.GetData(id))
```
emptyrivers, line 460: there's a perf risk here if an aura calls GetData many times. since they're not supposed to do that idk how much i care, but using a cache could be an enhancement if it ends up being a problem users complain too loudly about

```suggestion
      getDataCache = setmetatable({}, {__index = function(self, id) local copy = Copytable(WeakAuras.GetData) self[id] = copy return copy end})
      return getDataCache[id]
```

InfusOnWoW: I actually didn't implement a cache after all. Instead I added an aura warning if an aura calls GetData >= 100 times since the last init. That should catch auras that don't cache the call.

I tested it. btw, and CopyTable is pretty fast, so calling it repeately isn't actually that bad. Thus we can eventually remove the warning since the main prupose is just to warn users that relied on the function being essentially free.

PR #3880, opened 2022-10-13
Smallfix

mrbuds:


WeakAuras/WeakAuras.lua
```diff
@@ -3230,35 +3230,40 @@ local glow_frame_monitor
 local anchor_unitframe_monitor
 Private.dyngroup_unitframe_monitor = {}
 do
-  local function frame_monitor_callback(event, frame, unit)
```
InfusOnWoW, line 3233: In LibGetFrame, you can optimize CacheMonitorMixin:Add, by checking
if type(self.data[key]) ~= "table" and f self.data[key] == nil
before iterating. You basically only need to iterate over all args if type(self.data[key]) == table.

WeakAuras/WeakAuras.lua
```diff
@@ -3230,35 +3230,40 @@ local glow_frame_monitor
...
     local new_frame
-    local update_frame = event == "FRAME_UNIT_UPDATE"
+    local FRAME_UNIT_UPDATE = event == "FRAME_UNIT_UPDATE"
+    local FRAME_UNIT_ADDED = event == "FRAME_UNIT_ADDED"
+    local FRAME_UNIT_REMOVED = event == "FRAME_UNIT_REMOVED"
 
     local dynamicGroupsToUpdate = {}
 
     if type(glow_frame_monitor) == "table" then
       for region, data in pairs(glow_frame_monitor) do
```
InfusOnWoW, line 3242: This is actually quadratic, which generally is pretty unadvisable.
LibGetFrame calls the callback for each mapping that has changed, and this iterates over all regions.
So this code runs "number of added/changed/removed mappings" * "number of regions watching".

InfusOnWoW (approved):

WeakAuras/WeakAuras.lua
mrbuds: Implemented in https://github.com/mrbuds/LibGetFrame/commit/8abfd86e2038ae42ecc05f9082e6a778eaa41f30

PR #3929, opened 2022-10-26
Added support for ElvUI Cooldown module

AcidWeb:
# Description

This change allow ElvUI Cooldown module to replace icon cooldowns handled by default Blizzard code or OmniCC.
Feature is enabled by default for all ElvUI users as Cooldown module is enabled by default. It can be controlled by normal WA options (_Hide Timer Text_) or in ElvUI options panel. Where users also can find options to change style of the CD.

As requested code execute extensive nil checks.

## Type of change

## How Has This Been Tested

At the moment code need to be tested with `development` branch of ElvUI. But changes required will be merged and released soon. If somebody is using incompatible version of ElvUI code fail graciously.

ElvUI CD module operate only on text part of the cooldown frame so other elements like swipe are unaffected.

As requested following functions were tested:

- cooldown:SetReverse 
- cooldown:SetDrawBling
- cooldown:SetDrawSwipeOrg
- cooldown:SetDrawEdge 
- cooldown:SetCooldown
- cooldown:Pause
- cooldown:Resume

## Checklist

- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -38,6 +38,7 @@ globals = {
 	"CUSTOM_CLASS_COLORS",
 	"LibStub",
 	"OmniCC",
+	"ElvUI",
 
 	-- WeakAuras
 	"WeakAuras_DropDownMenu",
--- a/WeakAuras/RegionTypes/Icon.lua
+++ b/WeakAuras/RegionTypes/Icon.lua
@@ -261,6 +261,10 @@ local function create(parent, data)
   cooldown.SetDrawSwipeOrg = cooldown.SetDrawSwipe
   cooldown.SetDrawSwipe = function() end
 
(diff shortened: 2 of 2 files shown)


mrbuds (approved):

InfusOnWoW:
I'm not too keen on that. Since I have really not a high opinion on their abilities to not randomly break their own api.

AcidWeb:
I can't point to last API breakage that occurred in stable branch. Additionally we documented these three endpoints as used by WA so this knowledge will not disappear with time.

AcidWeb:
Technically previous issues were caused by AddonSkins. Not ElvUI. Team wise these projects might overlap a little but they are still separate projects.

PR #3939, opened 2022-10-28
Remove support for Shadowlands

mrbuds:
Remove WeakAuras.IsShadowlands and WeakAuras.IsDragonflight and use WeakAuras.IsRetail always
Remove all Shadowlands specific code


WeakAuras/Prototypes.lua
```diff
@@ -684,68 +684,6 @@ if WeakAuras.IsClassicOrBCCOrWrath() then
...
-    elseif extraOption == 5 then
-      return not (selected or known)
-    end
-    if extraOption == 0 or extraOption == 2 then
-      return selected or known
-    else
-      return selected
-    end
-  end
-elseif WeakAuras.IsDragonflight() then
```
InfusOnWoW, line 703: Ah this was dead code because we don't call CheckTalentByIndex on DF.

mrbuds: yup i was going to rewrite a CheckTalentBySpellId instead, but since data is already refreshed automatically for current spec, Private.talentInfo is good enough for enableTest

PR #3981, opened 2022-11-04
SpinBox Widget

mrbuds:
# Description

SpinBox Widget for replacing sliders

image

Support mousewheel 
Support dragging value with mouseclick down horizontally

I didn't made any layout change, widget is not meant to be this large

It's possible to test other arrow textures, this one has benefit to have pressed & disabled textures and compatible with classic, unlike 10.0 settings arrow that don't have disabled texture and won't work on classic


Nokramx:
i just randomly saw this and want to mention that i always liked that i can't screw anything up by scrolling through the options of a weakaura with the mousewheel. 

I don't know exactly how you implement this but to me it looks like you can accidentally  change values in an aura when you just want to scroll through the options.

PR #3983, opened 2022-11-04
Smallpatches

InfusOnWoW:


mrbuds (approved):
I think you want to remove tooltips commit, otherwise good fixes

PR #4000, opened 2022-11-08
use SetClampedToScreen on title, instead of trying to calculate where…

Ironii:
… options frame should be (fixes an issue where UIParent doesn't fill up whole screen)

# Description

When UIParent is smaller than screen size, current solution keeps snapping options frame because its trying to find its position, but isn't able to calculate it correctly, this fixes that isses.

Fixes #(issue)

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested

- [x] Limited testing with reloads and relogs, including with clean install, one thing that was not tested that well is logging in for the first time with old wtf file, i think on first login with new version the options frame position might be off from previous saved position, but that fixes itself after first login.

## Checklist

### DIFF
--- a/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
+++ b/WeakAurasOptions/OptionsFrames/OptionsFrame.lua
@@ -161,16 +161,27 @@ function OptionsPrivate.CreateFrame()
   frame.window = "default"
 
   local xOffset, yOffset
+
   if db.frame then
-    xOffset, yOffset = db.frame.xOffset, db.frame.yOffset
+    -- Convert from old settings to new
+    odb.frame = db.frame
+    if odb.frame.xOffset and odb.frame.yOffset then
+      odb.frame.xOffset = odb.frame.xOffset + GetScreenWidth() - (odb.frame.width or defaultWidth) / 2
+      odb.frame.yOffset = odb.frame.yOffset + GetScreenHeight()
+    end
+    db.frame = nil
+  end
+
+  if odb.frame then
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -166,11 +166,11 @@ function OptionsPrivate.CreateFrame()
...
 
   if not (xOffset and yOffset) then
-    xOffset = (defaultWidth - GetScreenWidth()) / 2
-    yOffset = (defaultHeight - GetScreenHeight()) / 2
+    xOffset = GetScreenWidth() / 2
+    yOffset = GetScreenHeight() - defaultHeight / 2
   end
 
-  frame:SetPoint("TOPRIGHT", UIParent, "TOPRIGHT", xOffset, yOffset)
+  frame:SetPoint("TOP", UIParent, "BOTTOMLEFT", xOffset, yOffset)
```
InfusOnWoW, line 173: Why do you want to change this? This invalidates the stored settings.

Ironii: Because every coordinate you get is from bottomleft corner of the screen, and in order to use the opposite corner, you have to calculate where it is. With this there is no offsetting the coordinates based on GetScreenWidth (or height), so even if someone overwrites the functions, this will still work. (Overwriting GetScreenWidth is something that i have seen some people do in attempt to fix their ultrawide screen UIs)

PR #4013, opened 2022-11-13
Smallpatches

InfusOnWoW:


mrbuds:
About the nil checks on Private.regions, should additional checks be added in Private.UnloadDisplays for `Private.regions[id].region:Collapse();` and `Private.CancelAnimation(Private.regions[id].region`?

PR #4097, opened 2022-12-04
smallfix

mrbuds:


Stanzilla:
Should probably add a comment or so as to why, blizzard returning the m+ tab etc

mrbuds:
> Should probably add a comment or so as to why, blizzard returning the m+ tab etc

done

PR #4121, opened 2022-12-06
smallfix

mrbuds:


WeakAuras/Types_Retail.lua
```diff
@@ -59,6 +59,7 @@ function Private.InitializeEncounterAndZoneLists()
 			end
 		end
 	end
+	EJ_SelectTier(10) -- see ticket #4119
```
InfusOnWoW, line 62: That's incorrect, we need to restore what it was previously. Not to 10

PR #4172, opened 2022-12-27
reset font object on acquire

emptyrivers:
The update dialog can sometimes set one of these widgets to have GameFontHighlight instead, which is smaller and if that widget is reused elsewhere it looks wrong


Stanzilla:
You forgot to bump the minor version in the widget

Issue #3875, opened 2022-10-10
Shared State between users


emptyrivers:
**Is your feature request related to a problem? Please describe.**

There are a fair number of auras which have a dependency on something like AngryNotes or ERT to essentially replicate the raid leader's configuration to the raiders without having to go through the trouble of publishing new aura versions every time you adjust your strategy on a boss.

So my thinking is, since this is a thing our users do, why not support this in house? 

**Describe the solution you'd like**

In the simplest possible implementation, an option somewhere which when enabled, asks you to select who the "source of truth" is for the state (by default, make it raid lead), and also provides a code box for user to write data into, akin to TSU custom variables.

WeakAuras would then, for each of these auras, disseminate the truth source's state to other group members, whose auras would then receive an event with the updated data.

Pros:
- we can support a broader swath of use cases that our users find.
- fewer addons to manage means ~~greater dominance of WeakAuras~~ users potentially have access to tools they wouldn't otherwise think to find
- fewer people would misuse custom option for custom behavior that a user isn't meant to modify unilaterally

Cons:
- redundant? how many people would bother to learn to use this when they already know how to use other notes addons
- performance? is this an undue burden on the addon message channel? perhaps we should discourage people from putting gigabytes of data into shared state?
- complexity? we would need to account for all the various ways that sync could be broken. Obviously if an addon message gets dropped there's not much we can do, but for it to be reliable we should probably look into how other similar addons ensure reliability?

InfusOnWoW:
Yeah! We do have quite a few other examples that this is something that would be useful, e.g. Zen Tracker, but also Interrupt trackers. What makes those more complicated is that there is a need for everyone to communicate to everyone.

Now, where I think you are correct is that "state" is the right primitive for communication. That is often communication in distributed systems is often modeled as each participant sending events to each other.

But I doubt people would be able to write reliable software on top of that, due to the unpredictable order in which events might come in. With states, it's much more likely that people automatically write code that only depends on the state, and not in which order events came in.

When I think about that kind of feature, I think of it as the final frontier but maybe that's just how I think about it, and maybe there's some kernel of it that can be implemented.

The way I think about it, and I'm really talking about things that I normally don't even mention because they are that far outside of any reasonable timeline.

[...]

PR #4218, opened 2023-01-08
add media custom option

### LINKED ISSUE #1102: custom options could use a sound option
The problem is that any alternative is ugly and a bad solution because
a) adding sounds from sharedmedia sources to a dropdown forces you to kinda explain that it requires an addon for it to work which makes everything look ugly
b) can't make the default dropdown hide if you don't have the sharedmedia addon. if it behaved equally to action tab's sound option and you were lacking the sharedmedia addon, then it would be blank and the user looking at your custom options would feel like "oh i have to input something here"
c) the process of adding all sounds to a table so it can be used in a dropdown is soul draining https://i.gyazo.com/9f235c6f1bc7c338680e55374988c9fc.gif

a sound option where the sound-path is the value?

https://i.gyazo.com/b7f4750b8d6e598035f2f92c50c9b4fd.png

emptyrivers:
Fixes #1102

### DIFF
--- a/WeakAuras/SubscribableObject.lua
+++ b/WeakAuras/SubscribableObject.lua
@@ -1,0 +1,90 @@
+if not WeakAuras.IsLibsOK() then return end
+--- @type string, Private
+local AddonName, Private = ...
+
+local WeakAuras = WeakAuras
+local L = WeakAuras.L
+
+--- @class SubscribableObject
+--- @field events table<string, frame[]>
+--- @field callback table<string, fun():nil>
+--- @field ClearSubscribers fun(self: SubscribableObject)
+--- @field ClearCallbacks fun(self: SubscribableObject)
+--- @field AddSubscriber fun(self: SubscribableObject, event: string, subscriber: frame, highPriority: boolean?)
+--- @field RemoveSubscriber fun(self: SubscribableObject, event: string, subscriber: frame)
+--- @field SetOnSubscriptionStatusChanged fun(self: SubscribableObject, event: string, cb: fun())
+--- @field Notify fun(self: SubscribableObject, event: type, ...: any)
+--- @field HasSubscribers fun(self: SubscribableObject, event: string): boolean
+--- @type SubscribableObject
+local SubscribableObject =
+{
+  --- @type fun(self: SubscribableObject)
+  ClearSubscribers = function(self)
+    self.events = {}
+  end,
+
+  --- @type fun(self: SubscribableObject)
+  ClearCallbacks = function(self)
+    self.callbacks = {}
+  end,
+
+  --- @type fun(self: SubscribableObject, event: string, subscriber: frame, highPriority: boolean?)
+  AddSubscriber = function(self, event, subscriber, highPriority)
+    if not subscriber[event] then
+      print("Can't register for ", event, " ", subscriber, subscriber.type)
+      return
+    end
+
+    self.events[event] = self.events[event] or {}
+    local pos = highPriority and 1 or (#self.events[event] + 1)
+    if TableHasAnyEntries(self.events[event]) then
+      tinsert(self.events[event], pos, subscriber)
+    else
+      tinsert(self.events[event], pos, subscriber)
+      if self.callbacks[event] then
+        self.callbacks[event]()
+      end
+    end
+  end,
+
+  --- @type fun(self: SubscribableObject, event: string, subscriber: frame)
+  RemoveSubscriber = function(self, event, subscriber)
+    if self.events[event] then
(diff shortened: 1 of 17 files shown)


WeakAurasOptions/AuthorOptions.lua
```diff
@@ -975,6 +975,35 @@ typeControlAdders = {
...
+      width = WeakAuras.normalWidth,
+      name = name(option, "default", L["Default"]),
+      desc = desc(option, "default"),
+      values = function()
+        if option.mediaType == "sound" then
+          return OptionsPrivate.Private.sound_types
+        else
+          return AceGUIWidgetLSMlists[option.mediaType]
+        end
+      end,
```
mrbuds, line 1001: This sort sounds like in other widgets
```suggestion
      end,
      sorting = function()
        if option.mediaType == "sound" then
          return OptionsPrivate.Private.SortOrderForValues(OptionsPrivate.Private.sound_types)
        else
          return nil
        end
      end,
```

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -975,6 +975,35 @@ typeControlAdders = {
       step = 1
     }
   end,
+  media = function(options, args, data, order, prefix, i)
+    local option = options[i]
+    args[prefix .. "mediaType"] = {
+      type = "select",
+      width = WeakAuras.normalWidth,
```
mrbuds, line 982: The list isn't wide enough for anything but sounds, can you make this doubleWidth?

WeakAurasOptions/AuthorOptions.lua
```diff
@@ -975,6 +975,35 @@ typeControlAdders = {
...
+      name = name(option, "default", L["Default"]),
+      desc = desc(option, "default"),
+      values = function()
+        if option.mediaType == "sound" then
+          return OptionsPrivate.Private.sound_types
+        else
+          return AceGUIWidgetLSMlists[option.mediaType]
+        end
+      end,
+      order = order(),
```
mrbuds, line 1002: suggestion for using LSM widgets
start of the "media" function add
```Lua
    local controls = {
      statusbar = "LSM30_Statusbar",
      border = "LSM30_Border",
      background = "LSM30_Background",
      font = "LSM30_Font"
    }
```
and add this in args[prefix.."default"]
```Lua
dialogControl = controls[option.mediaType],
```

mrbuds:
suggested changes work only in Author Mode

emptyrivers:
took the liberty of having the selected sound play once (on both author mode when choosing default, and in user mode)

mrbuds (approved):

InfusOnWoW:
I pushed two things to this branch now:

a) PlaySound playing on the master channel, because that works even if the sound is muted and is something we do for Actions/Conditions

b) A bigger rework of how FramTick/TimerTick work.

WeakAuras/SubscribableObject.lua
```diff
@@ -0,0 +1,90 @@
+if not WeakAuras.IsLibsOK() then return end
+--- @type string, Private
+local AddonName, Private = ...
+
+local WeakAuras = WeakAuras;
+local L = WeakAuras.L;
```
Stanzilla, line 6: no semicolons in new files :D

WeakAuras/SubscribableObject.lua
```diff
@@ -0,0 +1,90 @@
...
+--- @field callback table<string, fun():nil>
+--- @field ClearSubscribers fun(self: SubscribableObject)
+--- @field ClearCallbacks fun(self: SubscribableObject)
+--- @field AddSubscriber fun(self: SubscribableObject, event: string, subscriber: frame, highPriority: boolean?)
+--- @field RemoveSubscriber fun(self: SubscribableObject, event: string, subscriber: frame)
+--- @field SetOnSubscriptionStatusChanged fun(self: SubscribableObject, event: string, cb: fun())
+--- @field Notify fun(self: SubscribableObject, event: type, ...: any)
+--- @field HasSubscribers fun(self: SubscribableObject, event: string): boolean
+--- @type SubscribableObject
+local SubscribableObject =
```
emptyrivers, line 19: unless im mistaken, this feels like it's also some groundwork for #3875?

InfusOnWoW: Maybe, I'm not actually sure if it will be. It does tick some boxes, but also quite a few things are different from what they would need to be.

InfusOnWoW (approved):

Stanzilla:
@emptyrivers any last words?

PR #4244, opened 2023-01-17
Aura trigger: use new API for retail

mrbuds:


InfusOnWoW:
Note, I haven't tested your changes and what I added on top throughly at all, which needs to be done, but otherwise this is imoh ready,.

WeakAuras/BuffTrigger2.lua
```diff
@@ -261,9 +260,13 @@ local function UpdateToolTipDataInMatchData(matchData, time)
   end
   local changed = false
 
-  if matchData.unit and matchData.index and matchData.filter then
+  if matchData.unit and matchData.auraInstanceID then
```
InfusOnWoW, line 263: In you changes you basically passed in auraInstanceId as the index. I think that's strange and it's best to separate them.

And you missed the tooltip thing.

WeakAuras/BuffTrigger2.lua
```diff
@@ -303,13 +329,16 @@ local function UpdateMatchData(time, matchDataChanged, unit, index, filter, name
       lastChanged = time,
       filter = filter,
       index = index,
+      points = points,
```
InfusOnWoW, line 332: So I looked at all the other entries in the structure and points is the most interesting to me.

So points is as far as I can see the replacement for the additional return values that some UnitAura calls had. For e.g. Power Shield, the absorb value is returned as the first "point" value. In general it appears that for an aura there's one "point" per effect, e.g. https://www.wowhead.com/de/spell=397734 has 8 effects listed and thus point will contain 8 values. Or Avatar https://www.wowhead.com/de/spell=107574 there will be 7 values.

This isn't always the same information as is in the tooltip, sometimes tooltip values appear to come from somewhere else.

Now, since it's table, it's actually not directly useable as a text replacement and there are several ways we handle that
1) Do nothing at all, we ignore points
2) Add points to the state table, %c can retrieve them and display them but that's it. (And maybe some day in the future add text replacements that can deal with a table, e.g. something like %points.1 (?))
3) Add e.g. point1, point2, point3 as needed that refers to the individual members

I think 2) is actually reasonable for now. I don't have a good reason to believe that they would be that useful, but this gives enough support that people can try using the information.

WeakAuras/WeakAuras.lua
```diff
@@ -3889,6 +3889,29 @@ function WeakAuras.GetAuraTooltipInfo(unit, index, filter)
...
+  if WeakAuras.IsRetail() then
+    local tooltipData = C_TooltipInfo.GetUnitAura(unit, index, filter)
+    local secondLine = tooltipData and tooltipData.lines[2] -- This is the line we want
+    if secondLine then
+      for _, arg in ipairs(secondLine.args) do
+        if arg.field == "leftText" then
+          tooltipText = arg.stringVal or ""
+        end
+      end
+    end
```
mrbuds, line 3903: Is this retail block here only to not break custom auras that could use it WeakAuras.GetAuraTooltipInfo?

InfusOnWoW: Kinda, the check it's not really needed, as it shouldn't be possible to call the code on non-retail. So the check for retail can be removed.

mrbuds: i'v just closed a ticket that use it on retail https://github.com/WeakAuras/WeakAuras2/issues/4314 so i'm ok not break custom code calling it for this wow extension

WeakAuras/BuffTrigger2.lua
mrbuds: agreed for 2)

WeakAuras/BuffTrigger2.lua
mrbuds: for the tests, i wonder the impact the change on index field has on dynamic group sorting

InfusOnWoW (approved):

PR #4252, opened 2023-01-18
fix custom sort

emptyrivers:


WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -329,7 +329,7 @@ local sorters = {
     if #events > 0 then
       sortOn = {}
       for _, event in ipairs(events) do
-        events[event] = true
```
emptyrivers, line 332: :( i can't believe i didn't notice this

Stanzilla:
did you test this in-game?

WeakAuras/Modernize.lua
```diff
@@ -1633,7 +1633,7 @@ function Private.Modernize(data)
 
   if data.internalVersion < 62 then
     if data.regionType == "dynamicgroup" then
-      if data.sort == "CUSTOM" and type(data.sortOn) ~= "string" then
+      if data.sort == "custom" and type(data.sortOn) ~= "string" then
```
Stanzilla, line 1636: do you need another modernize to fix this up?

InfusOnWoW: Yeah you need to do that in a new modernize version, so that the code that we wrongly didn't convert gets converted.

Stanzilla (approved):

PR #4284, opened 2023-01-26
Smallfixes

mrbuds:


Stanzilla:
should be `Empowered Charge`, no?

mrbuds:
This overlay show between max empowered & when the cast end (lost of dps when you don't end cast as soon as possible in this section)
API to get the duration of this section is GetUnitEmpowerStageDuration which has not a better name for users
I used `Charged` because it's not "a charge" but it's "fully charged" maybe `Empowered Fully Charged` is better then?

Stanzilla:
ah yes, and yes, I think that's better

WeakAuras/Prototypes.lua
```diff
@@ -8563,6 +8563,14 @@ Private.event_prototypes = {
...
         end,
         enable = WeakAuras.IsRetail
+      },
+      {
+        name = L["Empowered Fully Charged"],
+        func = function(trigger, state)
+          if not state.stageTotal or not trigger.use_showChargedDuration then return 0, 0 end
+          return state.duration - state.stagesData[state.stageTotal].finish, 0
+        end,
+        enable = WeakAuras.IsRetail
```
InfusOnWoW, line 8573: missing () here

mrbuds: changed for all overlays

mrbuds: reverted because it break call to enable function

```
1x WeakAuras/GenericTrigger.lua:1272: attempt to call field 'enable' (a boolean value)
[string "@WeakAuras/GenericTrigger.lua"]:1272: in function `Add'
[string "@WeakAuras/WeakAuras.lua"]:2955: in function <WeakAuras/WeakAuras.lua:2878>
[string "@WeakAuras/WeakAuras.lua"]:3036: in function `Add'
[string "@WeakAuras/WeakAuras.lua"]:2504: in function `AddMany'
[string "@WeakAuras/WeakAuras.lua"]:1150: in function <WeakAuras/WeakAuras.lua:1121>

[string "@WeakAuras/WeakAuras.lua"]:1273: in function <WeakAuras/WeakAuras.lua:1224>
```

InfusOnWoW (approved):

PR #4288, opened 2023-01-26
Fix atlas textures if use full rotation is enabled for atlas textures

InfusOnWoW:


mrbuds:
I pushed a change to use region.texture.IsAtlas

PR #4295, opened 2023-01-30
Revert "adopt BAG_UPDATE_DELAYED"

emptyrivers:
This reverts commit 3df435cf866a9b8404a6dc642242321d6f1becfd.
We have to do this because somehow blizzard broke B_U_D on wotlk, then fixed it, then broke it again and propogated the bug to retail.


emptyrivers:
if blizzard fixes this in maint tomorrow, then don't merge this

InfusOnWoW:
@Stanzilla So what's blizzard saying? Can we get any information that they'll fix the bug soonish?

Stanzilla:
Asking

Stanzilla:
Should be in the next build, no ETA when that lands though

InfusOnWoW:
Eh, so do they intend to hotfix or not? If they do, I don't think we need it. If they are waiting for 10.0.7, then we probably want this.

mrbuds:
fix is on 10.0.5.47884 ptr

emptyrivers:
confirmed, this is fixed in latest build of WoW

PR #4296, opened 2023-02-01
Purge TBC

mrbuds:
Tested quickly on classic & wotlk without error, models/textures picker works


Stanzilla:
Should name Classic as ClassicEra so it's a little less confusing when you are seeing IsClassic and IsWrathClassic next to each other?

InfusOnWoW:
There's one BCC artifact left in .github/scripts/csv_to_lua.lua, but that's not too important

InfusOnWoW (approved):

mrbuds:
> There's one BCC artifact left in .github/scripts/csv_to_lua.lua, but that's not too important

These scripts are not used anymore as wowtools doesn't make csv files anymore.

Last update of models i did for wotlk & classic was done with csv made with https://www.kruithne.net/wow.export/ for db2 extraction then https://github.com/Marlamin/DBC2CSV
I didn't do one for retail because result was too much massive.

mrbuds:
> Should name Classic as ClassicEra so it's a little less confusing when you are seeing IsClassic and IsWrathClassic next to each other?

i'll do that

mrbuds:
Since Infus's review i have rebase + renamed Classic to ClassicEra in separate commit

PR #4297, opened 2023-02-01
fix

mrbuds:


mrbuds:
A migration from model_path to File ID has to be done, but this fixes weakauras breaking when a subregion model exists

InfusOnWoW (approved):

Stanzilla:
> A migration from model_path to File ID has to be done, but this fixes weakauras breaking when a subregion model exists

Waiting with merge for this then?

mrbuds:
> > A migration from model_path to File ID has to be done, but this fixes weakauras breaking when a subregion model exists
> 
> Waiting with merge for this then?

next commit made this comment void as no migration is needed, data is already there for a while

PR #4304, opened 2023-02-05
Blockburstevents

mrbuds:
# Description

This WIP PR aim at finding code run too much on same frame, and may be use as a bandage fix if it's good enough

It works on the (supposedly wrong) assumption that data fetchable with API is all ready when events start to fire in a single frame
On first event, it save time, and next events for same trigger will be skipped.

It's not good enough yet, but it helps find redundant events.

Also as Rivers pointed out https://www.wowace.com/projects/ace3/pages/api/ace-bucket-3-0 exists and doesn't do my terrible assumption

Findings:

cdReadyFrame: SPELL_UPDATE_COOLDOWN and ACTIONBAR_UPDATE_COOLDOWN both trigger same frame 99.99% of the time, except for known special case of Spinning Crane Kick <https://github.com/WeakAuras/WeakAuras2/pull/593>

bufftrigger2: UNIT_FLAGS and UNIT_NAME_UPDATE both trigger for same units when joining a raid group and not always in the order

Power trigger: UNIT_POWER_FREQUENT may be replaced

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -1844,6 +1844,7 @@ local function RecheckActiveForUnitType(unitType, unit, unitsToRemoveScan)
 end
 
 local Buff2Frame = CreateFrame("Frame")
+Mixin(Buff2Frame, Private.EventBurstBlock)
 Private.frames["WeakAuras Buff2 Frame"] = Buff2Frame
 
 local function EventHandler(frame, event, arg1, arg2, ...)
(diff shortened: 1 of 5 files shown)


InfusOnWoW:
So on the weekend I did take a look at that, and there are quite a few issues here, we can't apply this in the all-encompasing way you did, because:
* If multiple events happen, the api might return different things for later events. E.g. when writing the bt2, I initially had a similar time based check on whether to skip checking units. But it turns out that units can receive multiple UNIT_AURA events in the same frame and crucially the api returns different data on the later events. Now, what can work is to simply gather the events, set a flag and handle in in OnUpdate, since OnUpdate is always the last thing that runs for a given frame. (Afaik)
* For e.g. generic triggers, your change makes it so that only the first event comes through, that's sure to lead to some custom code not working.
* For load, the transformation is wrong, because the scanForLoadsImpl actually decides based on the event, which auras load function needs to be rerun. If you combine events like you do here.

Now with that being said, I generally like the approach of factoring this out into a component that can be mixed into other areas, and it's definately something to keep in mind when looking at the other PR.

PR #4331, opened 2023-02-25
Add support for queued spell costs

### LINKED ISSUE #4214: Spell Cost Overlay considering Queued Actions on resource bar
It would be nice if the Overlay Cost of Casts option in triggers would consider queued actions such as heroic strike and cleave, so that similary to a cast, a portion of the resource bar could be changed to indicate how much rage is being used, and also if the action is being queued. - I'm not sure if this has application outside of warrior. (I believe feral druids are able to queue maul for instance, could be applied here.)

Either an additional button to select Overlay cost of queued actions, or for the already present button to consider queued actions.

raethkcj:
# Description

Fixes #4214

## Type of change

## How Has This Been Tested

For each test: created a Progress Bar Aura with a Player/Unit Info - Power Trigger, with Overlay Cost of Casts checked.

- [X] Tested on a level 70 Warrior in Wrath of the Lich King Classic, using Heroic Strike and Cleave.
- [X] Tested on a level 1 Warrior in Classic Era, using Heroic Strike.
- [X] Tested on several WOTLK classes with no queueable spells, where previous cost overlay functionality still works.
- [ ] Test on retail for compatibility

## Checklist

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] ~~I'm unsure of the implications of the change I made to `statesParameter`. I'll comment on this below.~~

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -3373,14 +3373,35 @@ function Private.ExecEnv.CheckTotemName(totemName, triggerTotemName, triggerTote
   return true
 end
 
-function WeakAuras.GetSpellCost(powerTypeToCheck)
-  local spellID = select(9, WeakAuras.UnitCastingInfo("player"))
-  if spellID then
-    local costTable = GetSpellPowerCost(spellID);
-    for _, costInfo in pairs(costTable) do
-      -- When there is no required aura for a power cost, the API returns an aura ID of 0 and false for hasRequiredAura despite being valid.
-      if costInfo.type == powerTypeToCheck and (costInfo.requiredAuraID == 0 or costInfo.hasRequiredAura) then
-        return costInfo.cost;
+do
+  local classQueueableSpells = {
+    WARRIOR = 78,        -- Heroic Strike
+    HUNTER = 2973,       -- Raptor Strike
+    DRUID = 6807,        -- Maul
+    DEATHKNIGHT = 56815, -- Rune Strike
+  }
+  local class = select(2, UnitClass("player"))
+  local queueableSpell
+  local updateQueueableSpell = function()
+    -- Set the stored spell id to the highest rank of our class's queueable spell
+    queueableSpell = select(7, GetSpellInfo(GetSpellInfo(classQueueableSpells[class])))
+  end
+  local queueableSpellFrame = CreateFrame("Frame")
+  queueableSpellFrame:RegisterEvent("SPELLS_CHANGED")
+  queueableSpellFrame:SetScript("OnEvent", updateQueueableSpell)
+
+  function WeakAuras.GetSpellCost(powerTypeToCheck)
+    local spellID = select(9, WeakAuras.UnitCastingInfo("player"))
+    if not spellID and queueableSpell and IsCurrentSpell(queueableSpell) then
+      spellID = queueableSpell
+    end
+    if spellID then
+      local costTable = GetSpellPowerCost(spellID);
+      for _, costInfo in pairs(costTable) do
+        -- When there is no required aura for a power cost, the API returns an aura ID of 0 and false for hasRequiredAura despite being valid.
+        if costInfo.type == powerTypeToCheck and (costInfo.requiredAuraID == 0 or costInfo.hasRequiredAura) then
+          return costInfo.cost;
+        end
       end
     end
   end
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -3374,27 +3374,41 @@ function Private.ExecEnv.CheckTotemName(totemName, triggerTotemName, triggerTote
 end
 
 do
-  local classQueueableSpells = {
-    WARRIOR = 78,        -- Heroic Strike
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -2972,7 +2974,7 @@ Private.event_prototypes = {
 
       return ret
     end,
-    statesParameter = "unit",
+    statesParameter = "one",
```
raethkcj, line 2977: Since `CURRENT_SPELL_CAST_CHANGED` is not a `UNIT` event, the triggerFunc wasn't getting called for that event when `statesParameter` was set to `"unit"`. It looks like this might also change the key used in the `allstates` table later on, from e.g. `"player"` to `""`? I wasn't sure and would appreciate feedback if there's a better way to add a non-unit event.

WeakAuras/GenericTrigger.lua
```diff
@@ -3373,14 +3373,49 @@ function Private.ExecEnv.CheckTotemName(totemName, triggerTotemName, triggerTote
...
+      ["HUNTER"] = {
+        2973,  -- Raptor Strike
+      },
+      ["DRUID"] = {
+        6807,  -- Maul
+      },
+      ["DEATHKNIGHT"] = {
+        56815, -- Rune Strike
+      },
+    }
```
raethkcj, line 3393: A hard-coded table doesn't feel great, but the only alternative I could think of was looping through the spellbook and scanning tooltips for the Global String `SPELL_ON_NEXT_SWING = "Next melee"`.

WeakAuras/Prototypes.lua
```diff
@@ -2859,6 +2859,7 @@ Private.event_prototypes = {
         AddUnitEventForEvents(result, "player", "UNIT_SPELLCAST_STOP")
         AddUnitEventForEvents(result, "player", "UNIT_SPELLCAST_FAILED")
         AddUnitEventForEvents(result, "player", "UNIT_SPELLCAST_SUCCEEDED")
+        AddUnitEventForEvents(result, false, "CURRENT_SPELL_CAST_CHANGED")
```
mrbuds, line 2862: Can you register this event only for era & wotlk?

WeakAuras/Prototypes.lua
mrbuds: This will break the power trigger for unit=boss/nameplate/group/...

WeakAuras/Prototypes.lua
raethkcj: Sure thing!

WeakAuras/Prototypes.lua
raethkcj: I was worried about something like that. The reason I changed it is that `CURRENT_SPELL_CAST_CHANGED`'s first argument is sometimes `false`, so with `statesParameter` set to `unit`, this check in RunTriggerFunc fails, and the trigger func is never run:

https://github.com/WeakAuras/WeakAuras2/blob/d0f975fe0efc30c04a37a24b4e55c263a3855c7a/WeakAuras/GenericTrigger.lua#L587

raethkcj: Changed it to `if arg1 ~= nil` and reverted the power trigger's `statesParameter` to `unit`

mrbuds: I don't like this solution, if trigger expect an event with a unit as arg1, it needs to receive an event with a unit as arg1.
You can take the cast trigger as an example with WeakAuras.WatchForCastLatency or WeakAuras.WatchForNameplateTargetChange and WA_UNIT_TARGET_NAME_PLATE internal event

raethkcj: Thanks for the feedback, I've refactored it to use a helper function/event following that pattern.

mrbuds (approved):

WeakAuras/GenericTrigger.lua
```diff
@@ -3373,8 +3373,61 @@ function Private.ExecEnv.CheckTotemName(totemName, triggerTotemName, triggerTote
...
+    local class = select(2, UnitClass("player"))
+    queueableSpells = classQueueableSpells[class]
+  end
+
+  local queuedSpellFrame
+  function WeakAuras.WatchForQueuedSpell()
```
InfusOnWoW, line 3400: All of this should only be done on ClassicEraOrWrath

WeakAuras/GenericTrigger.lua
```diff
@@ -3373,8 +3373,61 @@ function Private.ExecEnv.CheckTotemName(totemName, triggerTotemName, triggerTote
...
+          for _, spellID in ipairs(queueableSpells) do
+            -- Check the highest known rank
+            local maxRank = select(7, GetSpellInfo(GetSpellInfo(spellID)))
+            if IsCurrentSpell(maxRank) then
+              self.queuedSpell = maxRank
+              break
+            end
+          end
+        end
+        WeakAuras.ScanEvents("WA_UNIT_QUEUED_SPELL_CHANGED", "player")
```
InfusOnWoW, line 3418: This sends an event even if the queuedSpell didn't change

raethkcj:
Thank you, I've addressed those issues.

PR #4335, opened 2023-03-02
Delay + Count option for event triggers

mrbuds:
Add a way to delay events for generic event triggers

For now it's enable only for Enter/leaving Combat trigger, and Combat Log trigger
I have not think yet about which trigger would benefit from this

Maybe it should be called Delay Timer instead of Offset? Offset is consistent with the option in DBM/BW triggers but it's not unusual that this option is not understood.

Timers are cleared when aura unload


WeakAuras/GenericTrigger.lua
```diff
@@ -756,8 +758,15 @@ function WeakAuras.ScanEventsInternal(event_list, event, arg1, arg2, ... )
     local updateTriggerState = false;
     for triggernum, data in pairs(triggers) do
       local allStates = WeakAuras.GetTriggerStateForTrigger(id, triggernum);
-      if (RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2, ...)) then
-        updateTriggerState = true;
+      local delay = GenericTrigger.GetDelay(data)
+      if delay == 0 then
+        if (RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2, ...)) then
+          if GenericTrigger.MatchEventCount(data, id, triggernum) then
```
InfusOnWoW, line 764: This fundamentally can't work. The trigger has already updated the state, and you are just hiding that the state was changed from immediate further processing, here. But that doesn't make that state change invisible to a lot of other systems.

This will break e.g. conditions, but also multiple triggers. 
For example take this aura:
!WA:2!LrvZUTTrq4i4w0c6lXQXgnO)zKGgKEjnrbjh6nrfjBxilPqrh7CPsljhrU2u7YU7sjlFlQab5SEeeqV1t6rWpblekYdGFe8tqNDPKDAlpqm7SZoZ333m7wQr5HLJkh9HhoNgYzD55Iq4BVJ482dgibvjXKvgliSWeUOdNYub1Q3YVU3YqopnIpM1Dmnd6F56L1JIHext66pvjOXXGq(BpsSY8JUQjzWSq(WaIkLhplxc9c55mv)RK5bWiGP6Mpya98f9QvTRFVU(v98NgLlikkN9GN)upBmZRzZWUn5XEmYqq64MZOQGSuYeq4zZyLN9Zp)kt(LwEDeU)TfPJaWI41Tt9MnNkZG00dIKoZVnu92ZJGaek(iGf7xVzNgh1CgM2cI48iTZ713tVTEZGWCPcHJnVYST70S6BR7vG86VQ3HTFZbT2Brrq9m0xVL(70FV(ULFUE3SVzqol0WTh)to7kavUGTRsKdoalA9HsOrGNIoeI0)GZIiQ8u7zgbviSjzBrS2(fa7qEe8x35ASrbcgj9nO8Jj)pU(maYQIenu5zKYKzsiDGTHQ)CKPsIYOXqYCuH8Gy8mYFzh9MlXLbKWZIfOKg9HTZkRGZv9KjeSzFYQXJS7BDA)nGlgsWawz4Y4my2nBF5p(SNyv7wytl7UFsUQXt5IFTuPsBCvH31WZT6r(TZ(kRtsokhi8dpMgPsCRIlVUOQ0ZHiRZ3TSyw9TRW2L29pffs6GjivfCfstF0PBR2TQR3uGm0eZCBGHwySb(nBfDyQS72qqVy3xNtIqjISRV))sfwxPLwFJ5IOJfKSPhVYOaaJOsAqk0FfN)K7tzBDqRw48IBBF)2h6DWE77V8MsBh(AFKFZdAv)wVDPxa3FfpVPh8W)xp4tK6vU0v(4Y4u(4gc43Zbw4KoF2tFsLxuicMnwy(ToLtnlmiyzqUsXzThbc8(1CJ7MalwLSTn(1cWmZIA3iGxI392BTNexZMj2Q7NqdpJbs5g2t0nKKcfMnPO7TSfWfLpqKeKa04e17Ct5KOgbkmuMAhVH5PkQJRzEEhDphVWuIuASCLO2ymMlSdXg4R)YS7T(5jtJ)vujb7frjZim6q7dlV0tQicvJlbIe6Qew29f6n1vwU(Th7vxHeWx2IKtnXzZDf3HekRHo2gTort1NQRemGYOYK)R7fuwrJaZNJ3y78Q7fC(WscI9ra5l1dC0PoU4XvoZv8WrfxG)6sV)pFanQslyS3TZOlgiWMlIxCSCJRlgiAyCzGwq3AE1R3QsonAXRPVO2jp(qy)97eGeyan2rqzMCdjzLXgvTv6d(8b((a03JKMLq2yUHTulYCMUwd7VeVhIvQDwXgLLvEfmQ8O)(K)5d

The first trigger is a combat log trigger that shows every 3. spell cast for 30s
The second trigger is a custom "PLAYER_STARTED_MOVING" trigger.

Now, if you just start casting random spell it looks like everything is working correctly, that is the information of every 3 spell cast even is shown.

But the "hidden" state is easily exposed via:
1) Start casting spell foo
2) Start casting spell bar
3) Start moving

=> the aura shows the information for spell "bar".

That same problem exists for lots of systems, just hiding that a state has changed won't work.

Now you might be able to add the counting check to the trigger by modifying GenericTrigger's ConstructFunction.
The end of the function that is generated there is essentially
```
if ( **tests** ) then
   create state
   return true
else
  return false
end
```

Whereas if I see that correctly, you'd "just" need to modify that too:
```
if ( **tests** ) then
   if (use_count) then
     count = count + 1
     if not check_count then
       return false
     end
   end     
   create state
   return true
else
  return false
end
```

The other option would be some "watching" trigger that counts the number of times the watched trigger activates, but wathching triggers have a few usuability challenges, which makes me reject that idea.

InfusOnWoW (commented):
The delay changes looked good, but the count check needs more work.

You can either try to improve the count check, or remove it for now.

mrbuds:
Did some changes to the Count option
- moved code to ConstructFunction 
- doesn't need filter to be checked to have the field added to state anymore
- stat's field renamed to `count`
- text replacement tooltip added for %count
- condition added

Things i don't like, but i don't know if there is a better way to achieve it:
- WeakAuras.GetNextCountTrigger and WeakAuras.MatchCountTrigger are in WeakAuras scope, but i think they belongs to Private, and i don't think i can use Private scope from there. Particularly GetNextCountTrigger which if use from custom code can break other auras
- `Private.AuraEnvironmentSetField(id, "triggernum", triggernum)` and getting id + triggernum from aura_env feels too hacky

InfusOnWoW:
You can move WeakAuras.GetNextCountTrigger and WeakAuras.MatchCountTrigger to Private.ExecEnv. Our own triggers have access to that, whereas user custom code doesn't.

I'll have to look at Private.AuraEnvironmentSetField to see what we can do about that.

WeakAuras/GenericTrigger.lua
```diff
@@ -1551,6 +1551,14 @@ function GenericTrigger.Add(data, region)
       end
     end
   end
+
+  if warnAboutCLEUEvents then
+    Private.AuraWarnings.UpdateWarning(data.uid, "spamy_event_warning", "warning",
+                L["COMBAT_LOG_EVENT_UNFILTERED with no filter can trigger frame drops in raid environment. Find more information:\nhttps://github.com/WeakAuras/WeakAuras2/wiki/Custom-Triggers#events"],
+                  true)
+  else
+    Private.AuraWarnings.UpdateWarning(data.uid, "spamy_event_warning")
```
InfusOnWoW, line 1672: For every aura warning there should be in general something that clears the warning.

Also I'm not sure I agree on the choice to print the warning on the console. I forsee lots of feedback on that, and not all positive :)

WeakAuras/GenericTrigger.lua
```diff
@@ -1272,6 +1302,78 @@ end
...
+do
+  local function ParseCron(pattern)
+    local tests = {}
+    for test in pattern:gmatch("[^ ,]+") do
+      local startString, endString, intervalString = test:match("(%d*)-?(%d*)/?(%d*)")
+      local intervalNumber = tonumber(intervalString) or 1
+      local startNumber = startString == "" and 1 or tonumber(startString) or 1
+      local endNumber = tonumber(endString)
+      if not endNumber then
+        endNumber = intervalString and math.huge or startNumber
```
InfusOnWoW, line 1314: I didn't fully understand what checks you made to decide on whether endNumber should be math.huge or startNumber, maybe I missed something here.

WeakAuras/GenericTrigger.lua
```diff
@@ -1272,6 +1302,78 @@ end
...
+      Reset = function(self)
+        self.count = 0
+      end,
+      GetNext = function(self)
+        self.count = self.count + 1
+        return self.count
+      end,
+    }
+    if pattern then
+      counter.tests = ParseCron(pattern)
```
InfusOnWoW, line 1346: I made quite a few changes here:
* The patterns are parsed the constructor, that is when the aura is constructed

WeakAuras/GenericTrigger.lua
```diff
@@ -1272,6 +1302,78 @@ end
...
+      counter.RunTests = function(self, count)
+        for _, test in ipairs(self.tests) do
+          if test:Match(count) then
+            return true
+          end
+        end
+        return false
+      end
+
+      for i = 1, 20 do
```
InfusOnWoW, line 1356: And for the first 20 numbers we check at construction time whether they match. This should cover most use cases.

WeakAurasOptions/LoadOptions.lua
```diff
@@ -898,7 +898,7 @@ function OptionsPrivate.ConstructOptions(prototype, data, startorder, triggernum
       type = "input",
       width = WeakAuras.normalWidth,
       name = L["Count"],
-      desc = L["Occurence of the event, reset when aura is unloaded\nCan be a range of values\nCan include repetition using cron notation\nCan have multiple values separated by a comma or a space\n\nExemples:\n2nd 5th and 6th events: 2 5 6\n2nd to 6th: 2-6\nevery 2 events: /2\nevery 3 events starting from 2nd: 2/3\nevery 3 events starting from 2nd and ending at 11th: 2-11/3"],
+      desc = L["Occurence of the event, reset when aura is unloaded\nCan be a range of values\nCan have multiple values separated by a comma or a space\n\nExamples:\n2nd 5th and 6th events: 2, 5, 6\n2nd to 6th: 2-6\nevery 2 events: /2\nevery 3 events starting from 2nd: 2/3\nevery 3 events starting from 2nd and ending at 11th: 2-11/3"],
```
InfusOnWoW, line 901: I removed the mention of "cron" here, because the syntax is close but not entirely identical to cron.
And cron is most likely not something that people are familar with.

Stanzilla: We could/should link to the wiki here

WeakAuras/GenericTrigger.lua
mrbuds: An example of test that broke with this version of the function

```Lua
-- test function
local function cron(test, count)
  local startString, endString, intervalString = test:match("(%d*)-?(%d*)/?(%d*)")
  local intervalNumber = tonumber(intervalString) or 1
  local startNumber = startString == "" and 1 or tonumber(startString) or 1
  local endNumber = tonumber(endString)
  if not endNumber then
    endNumber = intervalString and math.huge or startNumber
  end
  if count == 0 then
    print(startString, endString, intervalString)
    print(startNumber, endNumber, intervalNumber)
    print("###")
  end
  return (count >= startNumber and count <= endNumber and (count - startNumber) % intervalNumber == 0)
end
```
```
local test = "5"
for i = 0, 10 do
  print(i, cron(test, i))
end 

5		    -- endString == "" and intervalString == ""
5	inf	1 -- endNumber should be == startNumber
###
0	false
1	false
2	false
3	false
4	false
5	true
6	true
7	true
8	true
9	true
10	true
```

WeakAuras/GenericTrigger.lua
mrbuds: I think the benefit for users out-weight the annoyance, but that works only if they don't chose to ignore the message in chat, or don't know what to make of it

WeakAuras/GenericTrigger.lua
InfusOnWoW: Oh right I'll tweak that, the check for "intervalString" is wrong. I need to check that the user actually supplied an interval.

I've pushed a fix. We should squash the commits a bit before merging.

mrbuds:
I have squashed some commits and force pushed

InfusOnWoW (approved):

PR #4346, opened 2023-03-14
Smallpatches

InfusOnWoW:


mrbuds (approved):

WeakAuras/Conditions.lua
```diff
@@ -799,20 +799,20 @@ local function runDynamicConditionFunctions(funcs)
...
-local function handleDynamicConditions(self, event)
+local function handleDynamicConditions(self, event, onlyUpdateState)
   Private.StartProfileSystem("dynamic conditions")
   if (globalDynamicConditionFuncs[event]) then
     for i, func in ipairs(globalDynamicConditionFuncs[event]) do
       func(globalConditionState);
     end
   end
-  if (dynamicConditions[event]) then
+  if (dynamicConditions[event] and not onlyUpdateState) then
```
mrbuds, line 809: This doesn't work because when handleDynamicConditions is called from `dynamicConditionsFrame:SetScript("OnEvent", handleDynamicConditions)` with an event that have parameters, like `UNIT_POWER_FREQUENT` from example of https://github.com/WeakAuras/WeakAuras2/issues/4362, `onlyUpdateState` can have a value and then `runDynamicConditionFunctions` is skipped.

PR #4371, opened 2023-03-23
Smallfix

mrbuds:


WeakAuras/GenericTrigger.lua
```diff
@@ -3213,8 +3214,18 @@ do
         end
       end
     end
-    if count and count ~= "" and count ~= v.count then
-      return false
+    if count and count ~= "" then
+      local stateCount = tonumber(v.count) or 0
+      local match = false
+      for _, test in ipairs(Private.ParseCron(count)) do
```
InfusOnWoW, line 3220: This parses the cron patterns every time we check for a match. I think that's wasteful and we shouldn't do that.

Instead, you can optimize that by on line 5326, where the trigger function is created, do
local counter = Private.CreateTriggerCounter(%q)

This then allows to call counter:Match(triggerCount) either inside DBMTimerMatches or in the trigger function.

WeakAuras/GenericTrigger.lua
```diff
@@ -1342,6 +1342,9 @@ do
         self.count = self.count + 1
         return self.count
       end,
+      SetCount = function(self, count)
```
InfusOnWoW, line 1345: I squashed two very minor thigns into your commit. This new function, and the removal of the
```Private.ParseCron = ParseCron```
line

InfusOnWoW (approved):

PR #4395, opened 2023-04-08
Fix PlayerMoving trigger

InfusOnWoW:


WeakAuras/GenericTrigger.lua
```diff
@@ -3941,6 +3941,11 @@ do
     playerMovingFrame:RegisterEvent("PLAYER_STARTED_MOVING");
     playerMovingFrame:RegisterEvent("PLAYER_STOPPED_MOVING");
     playerMovingFrame:SetScript("OnEvent", PlayerMoveUpdate)
+    playerMovingFrame:SetScript("OnUpdate", function(self)
```
mrbuds, line 3944: If OnUpdate is required i think it's safe to stop listening to PLAYER_STARTED_MOVING and PLAYER_STOPPED_MOVING
And then i guess the channeling event spam workaround is not needed anymore

InfusOnWoW: Right, that simplifies the code a lot.

mrbuds (approved):

PR #4407, opened 2023-04-23
Smallfix

mrbuds:
Encounter trigger is ready, but Text sub element condition is WIP


mrbuds:
SubText condition for editing text is fully working now with %c, %p, etc..

PR #4443, opened 2023-05-06
add unignoring uption for WagoUpdates

Jodsderechte:
# Description

This pullrequest is supposed to add an option to the dropdownlist to be able to Unignore (and ignore) wago options for auras. Code is probably pretty suboptimal so please roast me to make things better thank you.

Having 2 popup dialogs that basically do exactly the same except having different texts should be changed but i wasn't sure how to handle this best.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) -- this depends on how updating is implemented in the companion  (which i don't know)

## How Has This Been Tested

Tried ignoring and unignoring updates which updated the auras fine in the SavedVariables.

## Checklist

Will prob need a change alongside in companion but i didn't doublecheck with buds https://github.com/WeakAuras/WeakAuras-Companion/issues/1543

### DIFF
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -670,6 +670,14 @@ local methods = {
       OptionsPrivate.Ungroup(self.data);
     end
 
+    function self.callbacks.IgnoreWagoUpdate(auraId)
+      if OptionsPrivate.IsWagoUpdateIgnored(auraId) then
+        StaticPopup_Show("WEAKAURAS_CONFIRM_UNIGNORE_UPDATES", "", "", auraId)
+      else
+        StaticPopup_Show("WEAKAURAS_CONFIRM_IGNORE_UPDATES", "", "", auraId)
+      end
+    end
+
     function self.callbacks.OnUpGroupClick()
       if (WeakAuras.IsImporting()) then return end;
       if(self.data.parent) then
@@ -884,6 +892,11 @@ local methods = {
       text = L["Export debug table..."],
       notCheckable = true,
       func = function() OptionsPrivate.ExportToTable(self.data.id) end
+    });
+     tinsert(self.menu, {
+      text = L["Ignore Wago Update"],
+      checked = function() return OptionsPrivate.IsWagoUpdateIgnored(self.data.id) end,
+      func = function() self.callbacks.IgnoreWagoUpdate(self.data.id) end
     });
 
     tinsert(self.menu, {
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -893,11 +893,13 @@ local methods = {
       notCheckable = true,
       func = function() OptionsPrivate.ExportToTable(self.data.id) end
     });
-    tinsert(self.menu, {
-      text = L["Ignore Wago Update"],
-      checked = function() return OptionsPrivate.IsWagoUpdateIgnored(self.data.id) end,
-      func = function() self.callbacks.ToggleIgnoreWagoUpdate(self.data.id) end
-    });
+    if OptionsPrivate.HasWagoUrl(self.data.id) then
+      tinsert(self.menu, {
+        text = L["Ignore Wago Update"],
+        checked = function() return OptionsPrivate.IsWagoUpdateIgnored(self.data.id) end,
+        func = function() self.callbacks.ToggleIgnoreWagoUpdate(self.data.id) end
+      });
+    end
       
     tinsert(self.menu, {
       text = " ",
--- a/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
+++ b/WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
@@ -670,7 +670,7 @@ local methods = {
       OptionsPrivate.Ungroup(self.data);
     end
 
-    function self.callbacks.IgnoreWagoUpdate(auraId)
+    function self.callbacks.ToggleIgnoreWagoUpdate(auraId)
       if OptionsPrivate.IsWagoUpdateIgnored(auraId) then
         StaticPopup_Show("WEAKAURAS_CONFIRM_UNIGNORE_UPDATES", "", "", auraId)
       else
@@ -893,12 +893,12 @@ local methods = {
       notCheckable = true,
       func = function() OptionsPrivate.ExportToTable(self.data.id) end
     });
-     tinsert(self.menu, {
+    tinsert(self.menu, {
       text = L["Ignore Wago Update"],
       checked = function() return OptionsPrivate.IsWagoUpdateIgnored(self.data.id) end,
-      func = function() self.callbacks.IgnoreWagoUpdate(self.data.id) end
+      func = function() self.callbacks.ToggleIgnoreWagoUpdate(self.data.id) end
     });
-
+      
     tinsert(self.menu, {
       text = " ",
       notClickable = true,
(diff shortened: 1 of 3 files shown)


WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -1025,7 +1058,7 @@ function OptionsPrivate.SortDisplayButtons(filter, overrideReset, id)
       button:ResetLinkedAuras()
     end
     for id, aura in pairs(WeakAurasSaved.displays) do
-      if not aura.ignoreWagoUpdate and aura.url and aura.url ~= "" then
+      if (aura.ignoreWagoUpdate and aura.ignoreWagoUpdate == true or not aura.ignoreWagoUpdate) and aura.url and aura.url ~= "" then
```
InfusOnWoW, line 1061: I don't understand this change.

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -670,6 +670,14 @@ local methods = {
       OptionsPrivate.Ungroup(self.data);
     end
 
+    function self.callbacks.IgnoreWagoUpdate(auraId)
```
InfusOnWoW, line 673: This function should be called somthing liek "ToggleIgnoreWagoUpdate" instead.

InfusOnWoW (changes requested):

WeakAurasOptions/WeakAurasOptions.lua
Jodsderechte: i might misunderstand lua syntax but "not aura.ignoreWagoUpdate" should return both for  aura.ignoreWagoUpdate being true or false right? so simply checking for that won't work if we just set the key to false

InfusOnWoW: not aura.ignoreWagoUpdate will return true for "nil" and "false", and the code inside the if should only be run if ignoreWagoUpdate is nil or false, so that looks correct without your change.

I'm not sure I understand your confusion though.

Jodsderechte: true ur totally correct and i'm not sure what i thought when doing that except it being late and me being tired

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -494,6 +494,39 @@ StaticPopupDialogs["WEAKAURAS_CONFIRM_IGNORE_UPDATES"] = {
...
+  preferredindex = STATICPOPUP_NUMDIALOGS,
+}
+
+function OptionsPrivate.IsWagoUpdateIgnored(auraId)
+    local isIgnored = false
+    local auraData = WeakAuras.GetData(auraId)
+      if auraData then
+        for child in OptionsPrivate.Private.TraverseAll(auraData) do
+          if(child.ignoreWagoUpdate == true) then
+            isIgnored = true
```
mrbuds, line 523: you can return true at this point and save time looping other auras

WeakAurasOptions/WeakAurasOptions.lua
mrbuds, line 1061:
```suggestion
      if aura.ignoreWagoUpdate and aura.url and aura.url ~= "" then
```

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -494,6 +494,39 @@ StaticPopupDialogs["WEAKAURAS_CONFIRM_IGNORE_UPDATES"] = {
...
+StaticPopupDialogs["WEAKAURAS_CONFIRM_UNIGNORE_UPDATES"] = {
+  text = L["Do you want to stop ignoring all future updates for this aura"],
+  button1 = L["Yes"],
+  button2 = L["Cancel"],
+  OnAccept = function(self)
+    if self.data then
+      local auraData = WeakAuras.GetData(self.data)
+      if auraData then
+        for child in OptionsPrivate.Private.TraverseAll(auraData) do
+          child.ignoreWagoUpdate = false
```
mrbuds, line 506: could you set ignoreWagoUpdate to nil instead?

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
```diff
@@ -885,6 +893,11 @@ local methods = {
       notCheckable = true,
       func = function() OptionsPrivate.ExportToTable(self.data.id) end
     });
+     tinsert(self.menu, {
+      text = L["Ignore Wago Update"],
+      checked = function() return OptionsPrivate.IsWagoUpdateIgnored(self.data.id) end,
+      func = function() self.callbacks.IgnoreWagoUpdate(self.data.id) end
+    });
```
mrbuds, line 900: This shouldn't be shown if there is no wago url for this aura

I'm not sure if this option belong to the contextual menu, if it was in the Information tab you would see wago url this aura is linked to, i think it make more sense to move it there

mrbuds (changes requested):

WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua
Jodsderechte: my first thought was having it in the information tab aswell but since i remembered it being in the dropdown before i thought best keep it there. also the options in the information tab didn't seem fitting since there is only compatibility and debug options so it'd prob need a new tab?

WeakAurasOptions/WeakAurasOptions.lua
InfusOnWoW:
```suggestion
      if not aura.ignoreWagoUpdate and aura.url and aura.url ~= "" then
```

WeakAurasOptions/InformationOptions.lua
```diff
@@ -100,6 +100,34 @@ function OptionsPrivate.GetInformationOptions(data)
...
+      name = L["Ignore Wago updates"],
+      desc = OptionsPrivate.IsWagoUpdateIgnored(data.id) and L["Do you want to stop ignoring all future updates for this aura"] or L["Do you want to ignore all future updates for this aura"],
+      width = WeakAuras.doubleWidth,
+      get = function() return OptionsPrivate.IsWagoUpdateIgnored(data.id) end,
+      set = function(info, v)
+          local auraData = WeakAuras.GetData(data.id)
+          if auraData then
+            for child in OptionsPrivate.Private.TraverseAll(auraData) do
+              if OptionsPrivate.IsWagoUpdateIgnored(data.id) then
+                child.ignoreWagoUpdate = nil
```
InfusOnWoW, line 115: This toggles individually on each aura, which is the wrong approach. You should toggle based on what IsWagoUpdateIgnored returns for the whole group.

WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -474,25 +474,29 @@ StaticPopupDialogs["WEAKAURAS_CONFIRM_DELETE"] = {
...
-  button2 = L["Cancel"],
-  OnAccept = function(self)
-    if self.data then
-      local auraData = WeakAuras.GetData(self.data)
+function OptionsPrivate.IsWagoUpdateIgnored(auraId)
+    local auraData = WeakAuras.GetData(auraId)
       if auraData then
         for child in OptionsPrivate.Private.TraverseAll(auraData) do
-          child.ignoreWagoUpdate = true
+          if(child.ignoreWagoUpdate == true) then
```
InfusOnWoW, line 481: you don't need to be so explicit here.
if child.ignoreWagoUpdate then
is sufficient.

mrbuds:
> i feel like i'm wasting more of y'alls time with pull request reviews then it would've taken to implement this yourselfs

I can have same feeling of wasting someone else time when asking a change, ultimately we are only robots on internet and feelings don't have their place in this world, this subsystem isn't helping improve the code 🤖 getting reviews from peers do

Jodsderechte:
beeep agreed boop

mrbuds (approved):

PR #4449, opened 2023-05-07
Fix

mrbuds:


Stanzilla:
How big or small is this fix? I can't accept a PR without a size. Is it smallfix, smallerfix or verysmallfix?

mrbuds:
What matters isn't the size but how well it fixes

Stanzilla:
> What matters isn't the size but how well it fixes

I can accept that. Have a nice day.

InfusOnWoW (approved):

PR #4480, opened 2023-05-30
LibCustomNames support

mrbuds:
Add support for https://github.com/Jodsderechte/LibCustomNames
I think Jods addon needs a few reviews from addon authors before merging this to be sure interface isn't going to need too much changes

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -626,14 +626,14 @@ Private.format_types = {
 
       if realm == "never" then
         nameFunc = function(unit)
-          return unit and UnitName(unit)
+          return unit and WeakAuras.UnitName(unit)
         end
       elseif realm == "star" then
         nameFunc = function(unit)
           if not unit then
             return ""
           end
-          local name, realm = UnitName(unit)
+          local name, realm = WeakAuras.UnitName(unit)
           if realm then
             return name .. "*"
           end
@@ -644,7 +644,7 @@ Private.format_types = {
           if not unit then
             return ""
           end
-          local name, realm = UnitName(unit)
+          local name, realm = WeakAuras.UnitName(unit)
           if realm then
             return name .. "-" .. realm
           end
@@ -655,7 +655,7 @@ Private.format_types = {
           if not unit then
             return ""
           end
-          local name, realm = WeakAuras.UnitNameWithRealm(unit)
+          local name, realm = WeakAuras.UnitNameWithRealmCustomName(unit)
           return name .. "-" .. realm
         end
       end
@@ -751,10 +751,11 @@ Private.format_types = {
 
       if realm == "never" then
         nameFunc = function(name, realm)
-          return name
+          return WeakAuras.GetName(name)
         end
       elseif realm == "star" then
         nameFunc = function(name, realm)
+          name = WeakAuras.GetName(name)
           if realm ~= "" then
             return name .. "*"
           end
@@ -762,6 +763,7 @@ Private.format_types = {
         end
       elseif realm == "differentServer" then
         nameFunc = function(name, realm)
+          name = WeakAuras.GetName(name)
           if realm ~= "" then
             return name .. "-" .. realm
           end
@@ -769,6 +771,7 @@ Private.format_types = {
         end
       elseif realm == "always" then
         nameFunc = function(name, realm)
+          name = WeakAuras.GetName(name)
           if realm == "" then
             realm = select(2, WeakAuras.UnitNameWithRealm("player"))
           end
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -76,6 +76,19 @@ local LDBIcon = LibStub("LibDBIcon-1.0")
 local LCG = LibStub("LibCustomGlow-1.0")
 local LGF = LibStub("LibGetFrame-1.0")
 
+local CustomNames = C_AddOns.IsAddOnLoaded("CustomNames") and LibStub("CustomNames") -- optional addon
+if CustomNames then
+  WeakAuras.GetName = CustomNames.Get
+  WeakAuras.UnitName = CustomNames.UnitName
+  WeakAuras.GetUnitName = CustomNames.GetUnitName
+  WeakAuras.UnitFullName = CustomNames.UnitFullName
+else
+  WeakAuras.GetName = function(name) return name end
+  WeakAuras.UnitName = UnitName
+  WeakAuras.GetUnitName = GetUnitName
+  WeakAuras.UnitFullName = UnitFullName
+end
+
 local timer = WeakAurasTimers
 WeakAuras.timer = timer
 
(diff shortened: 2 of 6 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -73,6 +73,17 @@ local LDBIcon = LibStub("LibDBIcon-1.0")
 local LCG = LibStub("LibCustomGlow-1.0")
 local LGF = LibStub("LibGetFrame-1.0")
 
+local LibCustomNames = IsAddOnLoaded("LibCustomNames") and LibStub("LibCustomNames") -- optional addon
```
Jodsderechte, line 76: since we agreed to change the name this should be
```lua
 local LibCustomNames = IsAddOnLoaded("CustomName") and LibStub("CustomNames")
 ```

File: WeakAuras/WeakAuras.toc [hunk 2]
```diff
@@ -18,7 +18,7 @@
 ## DefaultState: Enabled
 ## LoadOnDemand: 0
 ## SavedVariables: WeakAurasSaved
-## OptionalDeps: Ace3, LibCompress, LibSharedMedia-3.0, AceGUI-3.0-SharedMediaWidgets, Masque, GTFO, LibButtonGlow-1.0, LibSpellRange-1.0, LibRangeCheck-2.0, LibDBIcon-1.0, LibClassicDurations, LibClassicCasterino, LibGetFrame-1.0, LibSpecialization, !LibUIDropDownMenu, !!AddonLocale
+## OptionalDeps: Ace3, LibCompress, LibSharedMedia-3.0, AceGUI-3.0-SharedMediaWidgets, Masque, GTFO, LibButtonGlow-1.0, LibSpellRange-1.0, LibRangeCheck-2.0, LibDBIcon-1.0, LibClassicDurations, LibClassicCasterino, LibGetFrame-1.0, LibSpecialization, !LibUIDropDownMenu, !!AddonLocale, LibCustomNames
```
Jodsderechte, line 21: same here

WeakAuras/Types.lua
```diff
@@ -524,6 +524,7 @@ Private.format_types = {
             return ""
           end
           local name, realm = WeakAuras.UnitNameWithRealm(unit)
+          name = WeakAuras.GetName(name)
```
Jodsderechte, line 527: this won't work as chars in lib are saved in format Name-Realm you can either build the string before request or use another overwrite for  Lib.UnitFullName(unit) intead of UnitFullName e.g. in  WeakAuras.UnitNameWithRealm(unit) WeakAuras.lua line 5713ff

WeakAuras/WeakAuras.lua
```diff
@@ -73,6 +73,17 @@ local LDBIcon = LibStub("LibDBIcon-1.0")
...
+if LibCustomNames then
+  WeakAuras.GetName = LibCustomNames.Get
```
Jodsderechte, line 78: i basically added the whole set of available UnitName functions incase they are needed in certain spaces https://github.com/Jodsderechte/CustomNames/blob/main/README.md

InfusOnWoW:
There's actually already a sort-of standard for custom names in the role playing addon community, apparently the common protocol is this: https://github.com/Total-RP/Total-RP-3/wiki/Mary-Sue-Protocol

I haven't looked to deep into it, the protocol is a bit suboptimal for our use cases and the whole functionality is obviously beyond what we need. But maybe we should investigate that a bit.

Jodsderechte:
I was trying to look into actual implementation of this but trying to find the actual doc for that had me running in circles. i eventually found https://github.com/wow-rp-addons/LibMSP which seems to be the most up to date version. However their wiki is just a parsed version of https://moonshyne.org/msp/ which didn't help me in the first place. However i feel like my database would explode trying to save all the names of the people you were randomly in group with/moused over.  I'd also need to add options to turn off that nicknaming which kinda defeats the purpose of the "lib" (thats not a lib) to be very lightweight. It also seems like you can only ever have one addon active that is implementing libMSP or they create conflicts https://github.com/wow-rp-addons/LibMSP/wiki/Original-Mary-Sue-Protocol-documentation#addon-conflicts-and-how-to-avoid-them but i haven't checked that in detail. Maybe contacting those rp addon devs to implementing the saving of names in the databse would be the better bet.

InfusOnWoW:
As far as I undestood, there can be multiple addons requesting name information, but only one responding to name request information. Which is a pretty reasonable.

With the library/protocol you can request the nickname for a given player/realm but that's an async method, which is a bit problematic for WA's usage of unit formatting functions.

Jodsderechte:
Also %name text replacements don't work yet unsure if that is something you want to support. Same for condition -> UnitName/Realm and Trigger -> UnitName/Realm (if unit exists ofc. Could also be added for combatlog triggers but you'd need to add Realm if you want to GetName)  (either add both or none i guess)

mrbuds:
> Also %name text replacements don't work yet unsure if that is something you want to support. Same for condition -> UnitName/Realm and Trigger -> UnitName/Realm (if unit exists ofc. Could also be added for combatlog triggers but you'd need to add Realm if you want to GetName) (either add both or none i guess). I also changed Lib.Get() to also allow for single name fetching (it will assume realm = GetPlayerRealm())

I think having only text formatters support, without touching aura's triggers & loading, is good enough at least for now
It also add helper functions to your addon for custom coded auras for complex needs

WeakAuras/Types.lua
mrbuds: WeakAuras.UnitNameWithRealm can't return a fantasy value as it's used in triggers

Jodsderechte:
Did some ingame testing it seemed to work without any issues on retail: do note i do not have classic or tbc installed so can't adhere to how it works there.

Regarding the MarySue protocol i talked to meo [here](https://discord.com/channels/168296152670797824/168296152670797824/1121762863544025108) about it and came to the conclusion it's not realy sth worth adding right now.

Adding it would provide 2 issues.

1.  since as infus already mentioned requesting data of MSP is async would require adding callbacks to the unit formater.
2.  the idea for CustomNames was to give the player controle over how other players appear on his screen. Using information others can set themselfes would result in a mess like with DetailsNicknaming which wasn't the goal.

That being said adding a plugin later on to CustomNames to allow it to fetch info of the MarySue protocol would easily be achievable and be the cleaner solution (as otherwise you'd need to atleast be able to toggle using that info anyways)

Jodsderechte:
CustomNames support has now been added to DBM. https://github.com/DeadlyBossMods/DeadlyBossMods/pull/1125
This means there is now support for Dbm,[MRT](https://github.com/Jods-GH/CustomNames_MRT), [ELvUI](https://github.com/Jods-GH/CustomNames_ElvUI) , [Details and Plater](https://wago.io/CustomNames) aswell as someone created a [grid plugin ](https://www.curseforge.com/wow/addons/customnames_grid2)

Stanzilla (approved):
DIDN'T TEST IN GAME BUT CODE LGTM!

Jodsderechte:
From a quick test on beta seems to still work correctly. However i didn't do a more advanced check only sitting around in OG as i currently don't have a sub.

Issue #4483, opened 2023-06-03
DBM Stage triggers don't work with combat-related load conditions.


Kolmarwow:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues.

### Description

If you have a trigger with Other Addons -> DBM Stage -> Stage Counter and a combat-related load condition such as In Combat / In Encounter / Encounter ID(s) then the WA won't activate, without those load conditions it behaves as expected. 

### WeakAuras Version

WeakAuras 5.5.4

### World of Warcraft Flavor

Wrath of the Lich King Classic

### World of Warcraft Region

EU

### Tested with only WeakAuras

- [X] Yes

### Lua Error

```Text
^ by "Tested with only WeakAuras" I mean Tested with only WAs + DBM as naturally this trigger needs DBM to function, DBM is on version 3.4.40-wrath. No lua errors.
```

### Reproduction Steps

1. Create a new aura.
2. Create a trigger with Other Addons -> DBM Stage -> Stage Counter = 2 (for example).
3. Enable a combat-related load condition such as "In Encounter".
4. Aura won't activate, but if you remove the load condition then it will work as expected.

InfusOnWoW:
Can you try the version at https://github.com/WeakAuras/WeakAuras2/suites/13350946956/artifacts/730072741 ?

Kolmarwow:
Downloaded that and got this error when I went to make another DBM Stage test aura:

```8x Lua error in aura 'New': Trigger 1
WeakAuras Version: 5.5.4-9-gc0191a3
WeakAuras/GenericTrigger.lua:3414: attempt to call method 'GetState' (a nil value)
[string "@WeakAuras/GenericTrigger.lua"]:3414: in function `GetDBMStage'
...
```

I went into line 3414 of that file and changed that to "GetStage" and the error went away but the issue still persists.

PR #4484, opened 2023-06-03
Smallpatches

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -5761,7 +5761,7 @@ Private.event_prototypes = {
         trigger.use_cloneId and "true" or "false",
         trigger.use_extend and tonumber(trigger.extend or 0) or 0,
         trigger.use_remaining and "true" or "false",
-        trigger.remaining or 0,
+        type(trigger.remaining) or 0,
```
mrbuds, line 5764: I don't understand the change
`local triggerRemaining = ` will receive "string" or "number" or 0 if it's nil, or am i wrong?
Did you mean to do `type(trigger.remaining) == "number" and trigger.remaining or 0,` or `tonumber(trigger.remaining) or 0` instead?

InfusOnWoW: Oh yeah.

WeakAuras/GenericTrigger.lua
```diff
@@ -3414,7 +3411,7 @@ do
   end
 
   function WeakAuras.GetDBMStage()
-    return currentStage, currentStageTotal
+    return DBM:GetState()
```
mrbuds, line 3414: I don't see any GetState in my local DBM repo, did you mean GetStage?

mrbuds: There is no reason #4483 does not also happen with bigwigs triggers, but their GetStage function is in BossPrototype.lua so i don't know the equivalent to this DBM:GetStage() for them

InfusOnWoW: oh right again...

hmm no idea on how to get to that from bigwigs

mrbuds (approved):

InfusOnWoW:
I'll remove the reverts and also the reporter for the dbm stage trigger claims that the change doesn't work, so I need to investigate that.

PR #4495, opened 2023-06-13
More descriptive names for Triggers and Conditions in UI

mrbuds:
image
image
image

### DIFF
--- a/WeakAurasOptions/ConditionOptions.lua
+++ b/WeakAurasOptions/ConditionOptions.lua
@@ -1401,14 +1401,14 @@ local function addControlsForIfLine(args, order, data, conditionVariable, totalA
     local isFirst = path[#path] == 1;
     if (isFirst) then
       if (needsTriggerName) then
-        optionsName = optionsName .. string.format(L["Trigger %s"], check.trigger);
+        optionsName = optionsName .. OptionsPrivate.GetTriggerTitle(data, check.trigger)
       end
     else
       if (needsTriggerName) then
         if (parentType == "AND") then
-          optionsName = optionsName .. string.format(L["and Trigger %s"], check.trigger);
+          optionsName = optionsName .. string.format(L["and %s"], OptionsPrivate.GetTriggerTitle(data, check.trigger));
         else
-          optionsName = optionsName .. string.format(L["or Trigger %s"], check.trigger);
+          optionsName = optionsName .. string.format(L["or %s"], OptionsPrivate.GetTriggerTitle(data, check.trigger));
         end
       end
     end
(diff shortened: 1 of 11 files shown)


mrbuds:
Updated screenshot with last changes

Known issue:
- changes made thru SpinBox widget doesn't update condition's title

WeakAurasOptions/TriggerOptions.lua
```diff
@@ -253,10 +253,34 @@ function OptionsPrivate.ClearTriggerExpandState()
   maxTriggerNumForExpand = 0
 end
 
+function OptionsPrivate.GetTriggerTitle(data, triggernum, prefix)
```
InfusOnWoW, line 256: prefix appears to be always set to true.

mrbuds: Indeed, i wasn't sure i would have always use a prefix, at first, argument removed.

WeakAurasOptions/ConditionOptions.lua
```diff
@@ -313,6 +314,10 @@ local function addControlsForChange(args, order, data, conditionVariable, totalA
       conditions[i].changes[j].value = v;
       WeakAuras.ClearAndUpdateOptions(data.id)
     end
+    setValueReload = function(info, v)
+      setValue(info, v)
+      WeakAuras.FillOptions()
```
InfusOnWoW, line 319: So the way AceConfig typically works is that after a control has fired an OnValueChanged event, the settter of that option is called, and then the options are reopened.
See, https://github.com/WoWUIDev/Ace3/blob/5250653df38b943a99c3826636d13a807a2261ed/AceConfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua#L658, which in the part at https://github.com/WoWUIDev/Ace3/blob/5250653df38b943a99c3826636d13a807a2261ed/AceConfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua#L844 calls AceConfigDialog:Open.

We use that by in our setters rebuilding the options table, with all necessary changes.

For "range" controls though, the options are not opened automatically but only on "OnMouseUp". This is most likely to prevent the widgets from being rebuilt while dragging. [1]

WeakAuras.FillOptions does essentially the same as AceConfigDialog:Open, that's why it mostly works.
There's another common case where we need to open the options manually, that is when we use a confirmation dialog

[1] This change actually seems to work even with drag and drop, which I can't really explain. It should be rebuilding the options and thus changing the widget, so that shouldn't work.

InfusOnWoW (approved):

Issue #4502, opened 2023-06-19
Improving and Unifying bossmod callbacks


MysticalOS:
This isn't an overnight thing and something to be worked on by next tier or at least by next expansion. I have many notes on it here:
https://github.com/DeadlyBossMods/DBM-Unified/issues/231#issuecomment-1596023699

But basic gist is, BW and DBM callbacks are mostly cross compatible as is, and for majority of auras that aren't using a count or stupidly using msg text scanning when they could be using spellID, they work 1 for 1 just fine. However, when an author writes a weak aura, they target either bw or DBM and not both. It results in a weak aura that COUNT work with both mods, but doesn't.

There are a few steps to be able to unify the callbacks and they are basically
1. Get on same page with stage callback (bw and DBM problem not WA)
2. Support count arg in timer and announce objects so spellid + count can be used instead of string matching (DBM just did this, so did BW. It's better this is done in boss mods and not as reliant for weak aura's to do it in your handler cause this also just benefits all addons better or WAs using custom code instead of built in trigger parsing.)
3. BW needs to improve type filter. DBM lets you filter by ALL timer types, but bw is just "cast" and "cd" which means an author of a weak aura still needs to scan timer text itself for any timer object not using a spellID (such as a stage timer). This is the reason why the DBM type filter lets you choose a multitude of types. A good weak aura should be able to avoid requiring localized text scanning.

What needs doing on WA short term:
1. Adding support for the recently added count filter i just pushed to DBM. (https://github.com/DeadlyBossMods/DBM-Unified/commit/a9dae8eab96be1eb23b2001bd22710728ae5658a)
2. More stuff as BW adds it?
3. Scrap the "timer id" field on DBM timer start config. It's confusing and should never be used here. That arg is for custom code handlers only such as timer replacement weak auras or other addons like player which need unique timer guids (that's what timerId is basically). for WA general triggers though. Just need spellid and message like BW timer start

Long term:
1. After BW and DBM have gotten more unified (on same page), merge callbacks together
2. Create an even bigger emphasis to stop using msg text scanning when we're both providing ways to avoid doing msg text scanning. There will probably still be super niche cases it's still needed but typically, they're over used when they shouldn't be and not only does it cause conflicts if boss mods don't match timer/alert text, but also in multi language support.

mrbuds:
Having one trigger for both would indeed be better for users.

[...]

PR #4511, opened 2023-06-23
DBM Triggers: use count from callback instead of parsing text

mrbuds:
Asked in https://github.com/WeakAuras/WeakAuras2/issues/4502

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -3265,7 +3265,7 @@ do
 
   local function dbmEventCallback(event, ...)
     if event == "DBM_TimerStart" then
-      local id, msg, duration, icon, timerType, spellId, dbmType = ...
+      local id, msg, duration, icon, timerType, spellId, dbmType, _, _, _, _, _, timerCount = ...
       local now = GetTime()
       local expirationTime = now + duration
       bars[id] = bars[id] or {}
@@ -3276,7 +3276,7 @@ do
       bar.icon = icon
       bar.timerType = timerType
       bar.spellId = tostring(spellId)
-      bar.count = msg:match("%((%d+)%)") or msg:match("（(%d+)）") or "0"
+      bar.count = timerCount and tostring(timerCount) or "0"
       bar.dbmType = dbmType
 
       local barOptions = DBT.Options or DBM.Bars.options


WeakAuras/GenericTrigger.lua
```diff
@@ -3265,7 +3265,7 @@ do
 
   local function dbmEventCallback(event, ...)
     if event == "DBM_TimerStart" then
-      local id, msg, duration, icon, timerType, spellId, dbmType = ...
+      local id, msg, duration, icon, timerType, spellId, dbmType, _, _, _, _, _, timerCount = ...
```
InfusOnWoW, line 3268: What are the other callback values, anything interesting in there?

mrbuds: from https://github.com/DeadlyBossMods/DBM-Unified/commit/a9dae8eab96be1eb23b2001bd22710728ae5658a

fireEvent("DBM_TimerStart", id, msg, timer, self.icon, self.type, self.spellId, colorId, self.mod.id, self.keep, self.fade, self.name, guid, timerCount)

```
--ID: Internal DBM timer ID
--msg: Timer Text (Do not use msg has an event trigger, it varies language to language or based on user timer options. Use this to DISPLAY only (such as timer replacement UI). use spellId field 99% of time
--timer: Raw timer value (number).
--Icon: Texture Path for Icon
--type: Timer type (Cooldowns: cd, cdcount, nextcount, nextsource, cdspecial, nextspecial, stage, ai. Durations: target, active, fades, roleplay. Casting: cast)
--spellId: Raw spellid if available (most timers will have spellId or EJ ID unless it's a specific timer not tied to ability such as pull or combat start or rez timers. EJ id will be in format ej%d
--colorID: Type classification (1-Add, 2-Aoe, 3-targeted ability, 4-Interrupt, 5-Role, 6-Stage, 7-User(custom))
--Mod ID: Encounter ID as string, or a generic string for mods that don't have encounter ID (such as trash, dummy/test mods)
--Keep: true or nil, whether or not to keep bar on screen when it expires (if true, timer should be retained until an actual TimerStop occurs or a new TimerStart with same barId happens (in which case you replace bar with new one)
--fade: true or nil, whether or not to fade a bar (set alpha to usersetting/2)
--spellName: Sent so users can use a spell name instead of spellId, if they choose. Mostly to be more classic wow friendly, spellID is still preferred method (even for classic)
--MobGUID if it could be parsed out of args
local guid
--timerCount if current timer is a count timer. Returns number (count value) needed to have weak auras that trigger off a specific timer count without using localized message text
```

we convert colorId into a "type" field for 
```
Private.dbm_types = {
  [1] = L["Add"],
  [2] = L["AOE"],
  [3] = L["Targeted"],
  [4] = L["Interrupt"],
  [5] = L["Role"],
  [6] = L["Phase"],
  [7] = L["Important"]
}
```

InfusOnWoW:
The commit was just 4 days ago? https://github.com/DeadlyBossMods/DBM-Unified/commit/a9dae8eab96be1eb23b2001bd22710728ae5658a

I think we can't do that yet?

mrbuds:
It can wait indeed

InfusOnWoW:
Apparently now released in https://github.com/DeadlyBossMods/DBM-Retail/tree/10.1.13, on 27.6.

PR #4537, opened 2023-07-18
New "Boss Mod" triggers working with both DBM & BigWigs

### LINKED ISSUE #4502: Improving and Unifying bossmod callbacks

mrbuds:
# Description

This PR adds new triggers working with both DBM & BigWigs, which fixes #4502

Boss Mod Announce
Boss Mod Stage
Boss Mod Timer

All code for DBM & BW triggers were moved from `Prototypes.lua` & `GenericTrigger.lua` to `BossMods.lua`

New public function `WeakAuras.GetBossStage`

It has a few deprecated functions kept for compatibility with custom code

```
WeakAuras.GetAllDBMTimers
WeakAuras.GetDBMTimerById
WeakAuras.GetDBMTimer
WeakAuras.GetBigWigsTimerById
WeakAuras.GetAllBigWigsTimers
WeakAuras.GetBigWigsStage
WeakAuras.RegisterBigWigsTimer
WeakAuras.RegisterDBMCallback
```

Boss Mod Timer doesn't carry a "type" option, as both bossmod don't have common type options, and will always ignore matches for a cast bar.

## Type of change

## How Has This Been Tested
- [x] /pull 5 (both)
- [x] create test bars (both)
- [x] with jods timeline custom aura (both)
- [x] used in aberrus (BW)
- [ ] used in aberrus (DBM)


MysticalOS:
don't existing bw triggers have type. or at least a checkbox for cast/not cast? i feel we still have to match this or auraa that rely on it won't move over to unified. i know bw doesn't support advanced typing so it can't really go beyond cd and cast but that's probably fine for now. if it helps i can also just add a simp type arg that just returns true/false in our handler for caatbar or not but looking at this code you basically have that anyways by string finding case "cast" already.

if i'm mistaken this on this i'll correct post when i get home. family keeps forgetting my "stop scheduling shit on tuesday's" rule :/

MysticalOS:
I feel that if there isn't parity with feature of old trigger, no one will use new one at all. The type option is used extensively in many of weak aura examples i gave that targeted BW triggers, cause they need to know whether it's a CD or cast bar and they do trigger off cast bars sometimes.

I can understand no migration since that's extra work, but to ensure people even use the new trigger in future tiers, it has to do everything old one does, or they'd just always use old one. Parity to old BW trigger should be at base a minimum. I can understand why DBMs more thorough advanced typing isn't supported (cause BW obviously doesn't support it). but outs does allow very easily mapping to existing BW cast/nocast trigger mapping.

mrbuds:
It's not a problem of "extra work", we won't do anything that break previously made auras in the middle of an expansion.

mrbuds:
> no i get that, that's why no migration. existing WAs shouldn't be changed now I agree. But i'm saying if 10.2 hits ptr, and someone is creating NEW auras, they're not gonna use the new trigger if it lacks features old one has.

We can ask Ora to check on wago, but i think 99.99% of auras using these triggers doesn't use type options

MysticalOS:
if that's case, we can table the "type" stuff for now until it's further simplified and streamlined.

I just pushed https://github.com/DeadlyBossMods/DBM-Unified/commit/491d8a5ec0e709bae51110c43e81d0ebe62d2198 to address the over complexity of DBMs type system. Whether or not BW does parity to it, at very least, it still makes it easier to parse ours.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1508 @@
...
+  },
+  timedrequired = true
+}
+Private.category_event_prototype.addons["DBM Announce"] = L["DBM Announce"]
+
+Private.event_prototypes["DBM Timer"] = {
+  type = "addons",
+  events = {},
+  internal_events = {
+    "DBM_TimerStart", "DBM_TimerStop", "DBM_TimerStopAll", "DBM_TimerUpdate", "DBM_TimerForce", "DBM_TimerResume", "DBM_TimerPause"
```
Mitalie, line 397: DBM_TimerStopAll no longer exists.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1508 @@
...
+  },
+  timedrequired = true
+}
+Private.category_event_prototype.addons["Boss Mod Announce"] = L["Boss Mod Announce"]
+
+Private.event_prototypes["Boss Mod Timer"] = {
+  type = "addons",
+  events = {},
+  internal_events = {
+    "BossMod_TimerStart", "BossMod_TimerStop", "BossMod_TimerStopAll", "BossMod_TimerUpdate", "BossMod_TimerForce", "BossMod_TimerResume", "BossMod_TimerPause"
```
Mitalie, line 1298: BossMod_TimerStopAll doesn't exist.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1508 @@
...
+    if self.nextExpire then
+      self.recheckTimer = timer:ScheduleTimerFixed(self.RecheckTimers, self.nextExpire - now, self)
+    end
+  end,
+
+  EventCallback = function(self, event, ...)
+    if event == "BigWigs_Message" then
+      WeakAuras.ScanEvents("BigWigs_Message", ...)
+      if self.isGeneric then
+        WeakAuras.ScanEvents("BossMod_Announce", ...)
```
Mitalie, line 721: Boss Mod Announce trigger expects message and icon as first and second arg of BossMod_Announce (as in DBM_Announce), but BigWigs_Message has them as third and fifth arg, respectively.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1508 @@
...
+      test = "true",
+      init = "false"
+    },
+  },
+  automaticrequired = true,
+}
+Private.category_event_prototype.addons["BigWigs Timer"] = L["BigWigs Timer"]
+
+-- Unified
+if BigWigsLoader or not DBM then
```
Mitalie, line 1174: Since this check runs on load, it might be a good idea to add `BigWigs` and `DBM-Core` to OptionalDeps.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1508 @@
...
+
+-- deprecated functions for compatibility for custom coded auras
+WeakAuras.GetAllDBMTimers = function() return Private.ExecEnv.BossMods.DBM:GetAllTimers() end
+WeakAuras.GetDBMTimerById = function(...) return Private.ExecEnv.BossMods.DBM:GetTimerById(...) end
+WeakAuras.GetDBMTimer = function(...) return Private.ExecEnv.BossMods.DBM:GetTimer(...) end
+WeakAuras.GetBigWigsTimerById = function(...) return Private.ExecEnv.BossMods.BigWigs:GetTimerById(...) end
+WeakAuras.GetAllBigWigsTimers = function() return Private.ExecEnv.BossMods.BigWigs:GetAllTimers() end
+WeakAuras.GetBigWigsStage = function(...) return Private.ExecEnv.BossMods.BigWigs:GetStage(...) end
+WeakAuras.RegisterBigWigsTimer = function() end
+WeakAuras.RegisterDBMCallback = function() end
```
Mitalie, line 1508: While the no-op funcs prevent errors due to calling nonexistent functions, custom coded aura calling them only works if the events it expects have been registered by other triggers. Ignore if that's intentional.

MysticalOS:
Oh yeah, DBM_TimerStopAll was deprecated long ago when it was ensured TimerStop fires for every timer no matter what. even when mod:stop is called any timer that's stopped fires the correct stop callback per object.

Mitalie:
It doesn't necessarily have to be in this PR, especially if announce triggers are rarely used, but DBM Announce and Boss Mod Announce triggers could be improved by adding support for spellID / key and count, reducing or removing the need to use locale-sensitive messages as trigger options. DBM_Announce has spellID as 4th and count as 7th arg, BigWigs_Message has key as 2nd arg and count would need to be parsed from message similarly to BigWigs_StartBar.

Also, not a new issue but BigWigs Message trigger has BigWigs Addon option to test against the first arg of BigWigs_Message, however the first arg is a table value (<https://github.com/search?q=repo%3ABigWigsMods%2FBigWigs%20BigWigs_Message&type=code>) which will never match the string-valued option. It should probably be removed, or test against `<table>.moduleName` instead.

MysticalOS:
announce triggers are heavily used on echo of neltharian for bomb assignments, specifically for hearts/bombs. But the interesting note of them is they used string matching even when spellid and count was available in the BW announce object. I intentionally changed DBMs strings to match perfectly so they'd work in a unified. but I also didn't realize that weak auras never actually added support for spellid or count in the dbm announce object (and thus not in the boss mod one either). I'd argue it's definitely used enough to warrant addition before merge, or at least a new PR put in immediately after (we can do that if it helps).

Mitalie:
`Boss Mod Timer` trigger passes `extraArg1`/`"noCastBar"` to `Generic:TimerMatches` as the 6th argument, but `DBM:TimerMatches` expects it in the 7th position. The unexpected 6th arg causes `TimerMatches` to always return false.

The above bug would cause `Boss Mod Timer` to never activate when backed by DBM, if not for the call to `Generic:GetTimer` omitting `extraArg1`. While this allows the trigger to function at least partially, the discrepancy doesn't seem intentional.

mrbuds:
@Mitalie thanks for your review! I pushed a commit that should fix the issues you found

Mitalie (commented):
Found a few more things.

You can go ahead and mark my first review's comments as resolved - apparently I can't do that myself, it's restricted to the PR submitter and people with write access to the repository.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1598 @@
...
+      local triggerCounter = %q
+      local counter
+      if triggerCounter and triggerCounter ~= "" then
+        counter = Private.ExecEnv.CreateTriggerCounter(triggerCounter)
+      else
+        counter = Private.ExecEnv.CreateTriggerCounter()
+      end
+      local isBW = Private.ExecEnv.BossMods.Generic == Private.ExecEnv.BossMods.BigWigs
+      local isDBM = Private.ExecEnv.BossMods.Generic == Private.ExecEnv.BossMods.DBM
+      local extraArg1, extraArg2
```
InfusOnWoW, line 1395: I dislike that the generic trigger has to check which of the boss mods is active.

I'd much prefer if the BossMods "modules" would provide a function, e.g. CooldownTimerMatches(...), that behaved in the desired way and had the same api for both dbm and bw. So that the trigger has to not care about any differences.

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1598 @@
...
+      elseif dbmType == 5 then
+        r, g, b = barOptions.StartColorRR, barOptions.StartColorRG, barOptions.StartColorRB
+      elseif dbmType == 6 then
+        r, g, b = barOptions.StartColorPR, barOptions.StartColorPG, barOptions.StartColorPB
+      elseif dbmType == 7 then
+        r, g, b = barOptions.StartColorUIR, barOptions.StartColorUIG, barOptions.StartColorUIB
+      else
+        r, g, b = barOptions.StartColorR, barOptions.StartColorG, barOptions.StartColorB
+      end
+      bar.dbmColor = {r, g, b}
```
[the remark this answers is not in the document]
QartemisT: We can likely expose the colorVars table in DBT.

Though also, you should probably be doing `barOptions['StartColor' .. colorVariables[dbmType] .. 'G']` etc., rather than a nasty switch mess?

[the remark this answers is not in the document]
QartemisT: https://github.com/DeadlyBossMods/DBM-Unified/commit/08166b43c2f031de988597ffaab21c5fd74f3b75 Should help with this =]

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1598 @@
...
+
+-- deprecated functions for compatibility for custom coded auras
+WeakAuras.GetAllDBMTimers = function() return Private.ExecEnv.BossMods.DBM:GetAllTimers() end
+WeakAuras.GetDBMTimerById = function(...) return Private.ExecEnv.BossMods.DBM:GetTimerById(...) end
+WeakAuras.GetDBMTimer = function(...) return Private.ExecEnv.BossMods.DBM:GetTimer(...) end
+WeakAuras.GetBigWigsTimerById = function(...) return Private.ExecEnv.BossMods.BigWigs:GetTimerById(...) end
+WeakAuras.GetAllBigWigsTimers = function() return Private.ExecEnv.BossMods.BigWigs:GetAllTimers() end
+WeakAuras.GetBigWigsStage = function(...) return Private.ExecEnv.BossMods.BigWigs:GetStage(...) end
+WeakAuras.RegisterBigWigsTimer = function() Private.ExecEnv.BossMods.BigWigs:RegisterTimer() end
+WeakAuras.RegisterDBMCallback = function() Private.ExecEnv.BossMods.DBM:RegisterTimer() end
```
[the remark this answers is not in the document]
InfusOnWoW: On second thought, we probably shouldn't immediately deprecate it. Lots of people use the timeline aura, and won't update soonish.

Though we should still move these forwards to the aura envrionment's FakeWeakAurasMixin.

WeakAuras/BossMods.lua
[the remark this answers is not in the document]
MysticalOS: I'd like to add that these are NOT option types, just literally color assignments, type is an entirely differet arg in the actual callback, these should never be used to assign type to a weak aura, ever. In fact I just refactored them to do away with them being types in the GUI and now they're just color 1 through color 8 cause users had same confusion. thinking it set a timer type when all it sets is bar color. The types were just what that color was assigned to by default and nothing more. In fact, for this PR, i don't even see how they're relevant at all since what color DBM makes a bar isn't used in these triggers and as far as I've seen type is also omitted. if type is added though then refer to https://github.com/WeakAuras/WeakAuras2/pull/4537#issuecomment-1642607655

WeakAuras/BossMods.lua
```diff
@@ -0,0 +1,1598 @@
...
+  Private.ExecEnv.BossMods.BigWigs.isGeneric = true
+  Private.ExecEnv.BossMods.BigWigs.isInstalled = BigWigsLoader ~= nil
+elseif DBM then
+  Private.ExecEnv.BossMods.Generic = Private.ExecEnv.BossMods.DBM
+  Private.ExecEnv.BossMods.DBM.isGeneric = true
+  Private.ExecEnv.BossMods.DBM.isInstalled = true
+end
+
+-- make Private.ExecEnv.BossMods.Generic.GetStage public
+WeakAuras.GetBossStage = Private.ExecEnv.BossMods.Generic.GetStage
```
InfusOnWoW, line 1212: There is a subtle error here. The BigWigs implementation requires a self parameter.

WeakAuras/BossMods.lua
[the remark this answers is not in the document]
InfusOnWoW: The 'dbmType' is a misnaming on our end, it is what is called colorType in the dbm sources, so yes that's about the color not about the timer type.

MysticalOS:
`local barOptions = DBT.Options or DBM.Bars.options`
Pretty sure we eliminated DBM.Bars last expansion. No one should be running that, the force disable in DBM would make sure DBM for 9.x versions wasn't loading

MysticalOS:
I'd also still love to see Count arg added to "BossMod_Announce" object since it's a single line of parsing it from text (similar to what you already do for BW object) and using the count arg DBM provides (no need to even parse it) Then it's a knock out of the park in full usability next tier

mrbuds:
> I'd also still love to see Count arg added to "BossMod_Announce" object since it's a single line of parsing it from text (similar to what you already do for BW object) and using the count arg DBM provides (no need to even parse it) Then it's a knock out of the park in full usability next tier

Afaik BigWigs_Message event doesn't have a count arg, neither of the 3 message/announce triggers have a count option

MysticalOS:
right, that's why all the weak aura packs in aberus were using string scanning on echo of nelth.

if you look at any of popular packs they were scanning for msg match = "bomb (1)" "bomb (2)" etc. since there was no means to go "if spellid = 12345 and count = 4" on announce.

but yeah of original objects don't even have it then maybe worth another separate PR/issue then and not this one. :)

mrbuds:
i have not seen a single aura for neltharion using announce/message triggers, only bigwigs/dbm timers, but i didn't look at all of them

MysticalOS:
https://wago.io/Y2peTP20U
This one
Screenshot 2023-08-07 at 9 21 36 PM

MysticalOS:
Fortunately it had very few installs. So I'd call it a niche case, but it still exposes the design flaw of that currently this actually is the ONLY way to cherry pick specific counts (in announce object at least, timers support count).

InfusOnWoW (approved):

InfusOnWoW:
We can certainly add that, though it doesn't need to be part of this PR. So let's merge this first.

PR #4550, opened 2023-07-26
Improve formating of spellId & time options

mrbuds:
Fixes ##4549


Stanzilla (approved):

WeakAuras/RegionTypes/Text.lua
```diff
@@ -169,76 +174,124 @@ local function modify(parent, region, data)
...
     end
-  end
 
-  local customTextFunc = nil
-  if(Private.ContainsCustomPlaceHolder(data.displayText) and data.customText) then
-    customTextFunc = WeakAuras.LoadFunction("return "..data.customText)
-  end
+    local customTextFunc = nil
+    if(Private.ContainsCustomPlaceHolder(self.displayText) and data.customText) then
+      customTextFunc = WeakAuras.LoadFunction("return "..data.customText)
```
InfusOnWoW, line 229: That shouldn't be done in ConfigureTextUpdatem since the function should be already loaded.

WeakAuras/RegionTypes/Text.lua
```diff
@@ -169,76 +174,124 @@ local function modify(parent, region, data)
...
+    local UpdateText
+    if self.displayText and Private.ContainsAnyPlaceHolders(self.displayText) then
+      local getter = function(key, default)
+        local fullKey = "displayText_format_" .. key
+        if (data[fullKey] == nil) then
+          data[fullKey] = default
+        end
+        return data[fullKey]
       end
+      local formatters = Private.CreateFormatters(self.displayText, getter)
```
InfusOnWoW, line 215: I would prefer it if the formatters were created outside ConfigureTextUpdate.

That is in the loop above, which checks for CustomPlaceHolders also create formatters for any potential text. (With slight adjustments to not create duplicated formatters.) Store these in the region and pass it into ReplacePlaceHolders. Having too many formatters in there is not a problem.

WeakAuras/RegionTypes/Text.lua
```diff
@@ -169,76 +174,124 @@ local function modify(parent, region, data)
...
   end
 
-  if not UpdateText then
-    local textStr = data.displayText
-    textStr = textStr:gsub("\\n", "\n");
-    SetText(textStr)
+  function region:ConfigureSubscribers()
+    local visible = self:IsShown()
+    if self.Update then
+      if visible then
```
InfusOnWoW, line 271: That can't be right, there's no code to register the subscribers if the text becomes visible.

I think this check is bogus for texts.

WeakAurasOptions/RegionOptions/Text.lua
```diff
@@ -267,7 +291,7 @@ local function createOptions(id, data)
   };
 
   OptionsPrivate.commonOptions.AddCodeOption(options, data, L["Custom Function"], "customText", "https://github.com/WeakAuras/WeakAuras2/wiki/Custom-Code-Blocks#custom-text",
-                          37, function() return not OptionsPrivate.Private.ContainsCustomPlaceHolder(data.displayText) end, {"customText"}, false);
+                          37, hideCustomTextOption, {"customText"}, false);
```
InfusOnWoW, line 294: There's a bug in the sub text options, which is on my to do list, but since you are copying it, it's probably best if we'd fix it before.

For the custom text function, we essentially check each potential text value to check if we should show the input box.

We need to do something analog about text formatters, that's in an aura where the normal text contains no placeholders, but a condition can set the text to e.g. "%p", we should display the text formatter settings on the display text.

Essentially in the loop that adds the formatter options, also iterate all potential texts.

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -189,7 +196,9 @@ local function CreateNameOptions(aura_options, data, trigger, size, isExactSpell
         WeakAuras.UpdateThumbnail(data)
         WeakAuras.ClearAndUpdateOptions(data.id)
       end,
-      validate = isExactSpellId and WeakAuras.ValidateNumeric or nil
+      validate = isExactSpellId and WeakAuras.ValidateNumeric or nil,
+      control = "WeakAurasInputFocus",
+      getWithFocus = function() return trigger[optionKey] and trigger[optionKey][i] or "" end
```
InfusOnWoW, line 201: I'm not sure if that works with multi-selection.

WeakAuras/Prototypes.lua
```diff
@@ -5539,9 +5595,12 @@ Private.event_prototypes = {
       {
         name = "spellId",
         display = L["Spell Id"],
-        type = "string",
+        type = "spell",
```
InfusOnWoW, line 5598: This changes the type from a string to a number, is that backwards compatible? Do existing auras continue to work?

WeakAurasOptions/LoadOptions.lua
```diff
@@ -510,7 +510,7 @@ function OptionsPrivate.ConstructOptions(prototype, data, startorder, triggernum
         if (arg.showExactOption) then
           options["exact"..name] = {
             type = "toggle",
-            width = WeakAuras.normalWidth - 0.1,
```
InfusOnWoW, line 513: I need to look at this again.

InfusOnWoW (changes requested):

WeakAuras/Prototypes.lua
```diff
@@ -817,6 +817,53 @@ function WeakAuras.ValidateNumeric(info, val)
...
+  end
+end
+
+Private.tinySecondFormat = function(value)
+  if type(value) == "string" then value = tonumber(value) end
+  if type(value) == "number" then
+     local fraction = value - math.floor(value)
+     local ret
+     if value > 3600 then
+        ret = ("%i:%02i:%02i"):format(math.floor(value / 3600), math.floor((value % 3600) / 60), value % 60)
```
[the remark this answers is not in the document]
mrbuds: but %i does no?

WeakAurasOptions/ActionOptions.lua
```diff
@@ -229,6 +229,7 @@ function OptionsPrivate.GetActionOptions(data)
         width = WeakAuras.normalWidth,
         name = L["Sound"],
         order = 8.4,
+        itemControl = "WeakAurasMediaSound",
```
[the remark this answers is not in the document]
mrbuds: Difference is you can preview sounds when you are doing the selection, there is no need to select each sound one by one to test how it sound like

WeakAuras/Prototypes.lua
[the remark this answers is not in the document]
InfusOnWoW: Ah yes it does.

WeakAurasOptions/RegionOptions/Text.lua
mrbuds: Damn it's tricky to get OptionsPrivate.AddTextFormatOption grouped under same header_anchor across multiple call

WeakAuras/Prototypes.lua
mrbuds: no it's still a string
image

InfusOnWoW: Oh right, then that should be ok.

WeakAurasOptions/RegionOptions/Text.lua
InfusOnWoW: I think you need to extend AddTextFormatOption to take multiple inputs, since you want to keep the seenSymbols variable in between OptionsPrivate.Private.ParseTextStr calls.

mrbuds: i got this done in commit i just pushed, but it's incomplete (explained in commit message)

WeakAuras/RegionTypes/Text.lua
```diff
@@ -179,43 +179,46 @@ local function modify(parent, region, data)
...
           and change.property == "displayText"
           and type(change.value) == "string"
-          and Private.ContainsCustomPlaceHolder(change.value)
           then
-            containsCustomText = true
-            break
+            if not containsCustomText and Private.ContainsCustomPlaceHolder(change.value) then
+              containsCustomText = true
+            end
+            allFormatters[change.value] = Private.CreateFormatters(change.value, getter)
```
InfusOnWoW, line 204: I don't think you should be keeping separate list of formatters, that is if two conditions both change the text to %foo, I would expect just one %foo-formatter, and one %foo option. 

So you probably need to extend CrateFormatters to take multiple strings as input. (Or inline it).

WeakAurasOptions/LoadOptions.lua
mrbuds: It wasn't shown at the same place for all triggers (aura vs cast) and didn't have some position whither a scrollbar is shown or not
I made it always shown on the left of the input

WeakAurasOptions/LoadOptions.lua
```diff
@@ -583,18 +583,22 @@ function OptionsPrivate.ConstructOptions(prototype, data, startorder, triggernum
...
                 if (trigger[realname] and trigger[realname] ~= "") then
-                  if useExactSpellId then
-                    local spellId = tonumber(trigger[realname])
-                    if (spellId and spellId ~= 0) then
-                      return tostring(spellId);
+                  local isSpellID = WeakAuras.SafeToNumber(trigger[realname])
+                  if isSpellID then
+                    local spellName, _, _, _, _, _, spellID = GetSpellInfo(WeakAuras.SafeToNumber(trigger[realname]))
+                    if spellName and spellID then
+                      return ("%s (%s)"):format(spellID, spellName)
```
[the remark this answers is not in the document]
mrbuds: An other way to show it i imagined (but complicated?) is this 
Wow_2023-07-31_03-19-49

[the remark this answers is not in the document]
InfusOnWoW: Right, it's probably fine as is. All the information is there.

WeakAurasOptions/ActionOptions.lua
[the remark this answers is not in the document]
InfusOnWoW: Okay, if you prefer that, then no objection from me.

WeakAurasOptions/BuffTrigger2.lua
mrbuds: It does not error, here is a video of the behavior with 2 auras with aura trigger

https://github.com/WeakAuras/WeakAuras2/assets/189656/22d88b43-495e-4ed7-a7b8-6db4f02e48bf

InfusOnWoW (approved):

PR #4570, opened 2023-08-14
UI++

mrbuds:
WIP

it was forked from https://github.com/WeakAuras/WeakAuras2/pull/4550 16 first commit of this PR will be gone after rebase

image

image

image

image

image

image

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -35,6 +35,7 @@ globals = {
 
 	-- misc custom
 	"AceGUIWeakAurasMultiLineEditBoxInsertLink",
+	"AceGUIWeakAurasMultiLineEditBoxWithEnterInsertLink",
 	"AceGUIWidgetLSMlists",
 	"CUSTOM_CLASS_COLORS",
 	"LibStub",
--- a/WeakAuras/BossMods.lua
+++ b/WeakAuras/BossMods.lua
@@ -594,7 +594,7 @@ Private.event_prototypes["DBM Timer"] = {
     return ret:format(
       trigger.use_count and trigger.count or "",
       trigger.use_id and trigger.id or "",
-      trigger.use_spellId and trigger.spellId or "",
+      trigger.use_spellId and tostring(trigger.spellId) or "",
       trigger.use_message and trigger.message or "",
       trigger.use_message and trigger.message_operator or "",
       trigger.use_cloneId and "true" or "false",
@@ -1220,7 +1220,7 @@ Private.event_prototypes["BigWigs Timer"] = {
     ]=]
     return ret:format(
       trigger.use_count and trigger.count or "",
-      trigger.use_spellId and trigger.spellId or "",
+      trigger.use_spellId and tostring(trigger.spellId) or "",
       trigger.use_text and trigger.text or "",
       trigger.use_text and trigger.text_operator or "",
       trigger.use_cloneId and "true" or "false",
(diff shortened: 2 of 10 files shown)


MysticalOS:
Good changes. But the ones that already support multiple (like encounter Ids) I feel is kind of redundant no?

It makes sense with additional support for > or < like you may need with a level range but the fields that are just 123,456,789 as a simple entry, additional boxes seems like UI clutter. But I suppose the idea is if you plan to unify things and eliminate supporting multiple in a single box, then it makes more sense...well until you get someone who needs 8 encounter Ids in a single weak aura, and now you got only 3 boxes. 

So I guess my question is, npcs, spellids, encounterIds, that already support multiple, and at least based on above screen shots aren't adding anything new like > or <, what is the plan there?

InfusOnWoW:
The Encounter Id change was a mistake, for npcIs I have changes that make that use a single text field instead of multiple.

For spells we also show additional information if e.g. a spell id is entered, so I'm not sure yet what to do about that. Some users are likely creating a single Cast trigger for all relevant casts in dungeons. That's a lot of names/spell ids.

I do have some additional changes on top of this, that fix a few bugs and change a few things, which I need to finish.

InfusOnWoW:
@mrbuds I pushed what I have so far to the branch, there are a few things I want to still look at, including the spell input mentioned above.

mrbuds:
> The Encounter Id change was a mistake, for npcIs I have changes that make that use a single text field instead of multiple.
> 
> For spells we also show additional information if e.g. a spell id is entered, so I'm not sure yet what to do about that. Some users are likely creating a single Cast trigger for all relevant casts in dungeons. That's a lot of names/spell ids.
> 
> I do have some additional changes on top of this, that fix a few bugs and change a few things, which I need to finish.

How i saw it, i wanted convert these multi entry single field, into single entry multi field, then (TODO) improve code to use hash table, and (TODO) make migration to split these single fields

\+ for single field:
- easier to copy/paste a one line list between auras
- more compact when using big list

\+ for multiple fields
- all entries are visible and easier to edit when list size is wider than widget
- spellid fields show additional information

If we were to keep single fields with possible long input, an improvement to the widget may be done to improve visibility, like a multi line widget with auto-height?

MysticalOS:
> How i saw it, i wanted convert these multi entry single field, into single entry multi field, then (TODO) improve code to use hash table, and (TODO) make migration to split these single fields
I've seen auras that are every encounterId in the instance. or as buds said a single interrupt aura with like every spellid that's interruptible or a gtfo or whatever. Or zone auras that list 8+ zone Ids such as a M+ aura using zoneids as load condition.

this will turn that single box into a massive scroll of boxes.

spellid, creatureid, encounterId, and zoneid are probably the 4 it's fairly common to bloat up quickly.

But I also see the concern for having some boxes support multi entries and some being multi boxes. from an end user standpoint, it lacks consistency.

At end of the day, a long scrollbox of boxes isn't end of world, but some might see it as a step backwards.

Jodsderechte:
I personally vastly prefer the multi box style over the multiple fields in a single box. It's not only way easier to edit, but also way better to spot at a glance what is going on in the trigger. Yes, it can blow up if people add a ton of things into that, but with a single box you just don't see the things anymore. Maybe a collapsible group and copy/paste functionality could be added if more than X fields are present.

mrbuds:
Just testing things to see how it looks/fit

current
image

multiline = true
image

width = WeakAuras.doubleWidth
image

width + multiline
image


image

InfusOnWoW:
All of them would work better if we hid the unused input boxes.

mrbuds:
> All of them would work better if we hid the unused input boxes.

Added a "multilineWhenEnable" option, only applied to encounter id load option, and combat log trigger's source name for now as examples

InfusOnWoW (approved):

PR #4592, opened 2023-09-09
UI+++

mrbuds:
right now, only last commit is new compared to ui++ branch


InfusOnWoW:
I looked into how we are creating the fake states for options and whether we can find some generic way to do that. 
But wasn't really convinced of any of my ideas on how that should look like.

I move your code a bit around so that RunTriggerFunc doesn't need to know about fake states, and while at it noticed that one of the fake state related things in RunTriggerFunc can be moved out of it.

InfusOnWoW (approved):

PR #4608, opened 2023-09-17
Smallpatches

InfusOnWoW:


WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -1035,6 +1035,13 @@ local function GetBuffTriggerOptions(data, triggernum)
       width = WeakAuras.doubleWidth,
       hidden = function() return not (trigger.type == "aura2" and (trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party")) end
     },
+    inRange = {
+      type = "toggle",
+      name = L["Ignore out of casting range"],
+      order = 69.81,
+      width = WeakAuras.doubleWidth,
+      hidden = function() return not (trigger.type == "aura2" and (trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party")) end
+    },
```
mrbuds, line 1044: For all triggers with this test:

WOTLK: register `UNIT_IN_RANGE_UPDATE` trigger an error (but register works on wotlk ptr)
CLASSIC_ERA: register works, but i couldn't test if it is in a working state, on retail it trigger when i switch target to an other enemy, it doesn't do that on classic_era, so it's worth testing if it effectively work first

Also "Ignore out of casting range" could benefit a better tooltip to see difference compared to "Ignore out of checking range"
like "Uses UnitInRange() to check if in range. Matches default raid frames out of range behavior, which is between 40 yards and 25 yards depending on your spec."

and for "Ignore out of checking range": 
"Uses UnitIsVisible() to check if in range. Game client loads objects with a distance of around 100 yards.\nThis is polled every second."

WeakAuras/BuffTrigger2.lua
```diff
@@ -2221,6 +2238,105 @@ Buff2Frame:RegisterEvent("PARTY_MEMBER_ENABLE")
...
+  frames = {
+
+  },
+  --- @type table<string, table<string, {player: number?, pet: number?}>>
+  unitTypePetMode = {
+
+  },
+  --- @type fun(self: perUnitFrames, unitType: string, event: string, petMode: "PlayersAndPets"|"PetsOnly"|nil)
+  Register = function(self, unitType, event, petMode)
+    if WeakAuras.IsRetail() then
```
mrbuds, line 2257: I don't think this is the best place to test IsRetail()

mrbuds (changes requested):

WeakAurasOptions/BuffTrigger2.lua
mrbuds: Can you add a In Range condition for all of these triggers?

PR #4623, opened 2023-09-29
fixes

mrbuds:
fix issue reported on discord:

```
Hey, I'm trying to make a simple WA for Wake of Ashes that glows 1 second before it becomes off cooldown. 
Here is a link to the WA: https://wago.io/2EGRN4UT-
Here is a screenshot of the trigger: https://i.imgur.com/AVzBswZ.png
Here is a screenshot of the conditions: https://i.imgur.com/mYUwKse.png

In combat, condition 2 doesn't fire at all. Out of combat it does, but only sometimes. I'm guessing this has something to do with the CDR from Seal of Order (https://www.wowhead.com/spell=385129/seal-of-order) ? The same setup works fine with Blade of Justice, Judgment, and Hammer of Wrath.
```

I think this fix is not enough if the time is slow down but couldnt test it

### DIFF
--- a/WeakAuras/Conditions.lua
+++ b/WeakAuras/Conditions.lua
@@ -371,9 +371,15 @@ local function CreateTestForCondition(uid, input, allConditionsTemplate, usedSta
     -- If adding a new condition type, don't forget to adjust the validator in the options code
 
     if (cType == "timer" and value) then
-      recheckCode = "  nextTime = state[" .. trigger .. "] and not state[" .. trigger .. "].paused"
-                    .. " and state[" .. trigger .. "]" .. string.format("[%q]",  variable)
-                    .. " and (state[" .. trigger .. "]" .. string.format("[%q]",  variable) .. " - " .. value .. ")\n"
+      if useModRate then
+        recheckCode = "  nextTime = state[" .. trigger .. "] and not state[" .. trigger .. "].paused"
+        .. " and state[" .. trigger .. "]" .. string.format("[%q]",  variable)
+        .. " and (state[" .. trigger .. "]" .. string.format("[%q]",  variable) .. " - " .. value .. " * (state[" .. trigger .. "].modRate or 1.0))\n"
+      else
+        recheckCode = "  nextTime = state[" .. trigger .. "] and not state[" .. trigger .. "].paused"
+        .. " and state[" .. trigger .. "]" .. string.format("[%q]",  variable)
+        .. " and (state[" .. trigger .. "]" .. string.format("[%q]",  variable) .. " - " .. value .. ")\n"
+      end
       recheckCode = recheckCode .. "  if (nextTime and (not recheckTime or nextTime < recheckTime) and nextTime >= now) then\n"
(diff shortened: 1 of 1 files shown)


WeakAuras/Conditions.lua
```diff
@@ -162,7 +162,7 @@ function Private.ExecEnv.ScheduleConditionCheck(time, uid, cloneId)
         checkConditions[uid](region);
         Private.ActivateAuraEnvironment()
       end
-    end, time - GetTime())
+    end, (time - GetTime()) / (modRate or 1.0))
```
InfusOnWoW, line 165: I'm not sure that this is correct. It's been a while since I last looked at modRate, but I think the expirationTime in the state is always correct and modRate basically only affects how remaining time is formatted.

mrbuds: Line 278 the test on remainingTime correctly consider modRate
Issue i found is that the condition function is run too early when this test is not true yet

InfusOnWoW: Right but this is then the wrong place to fix it. The calculation of nextTime needs to be adjusted. Also in line 278 the usage of modRate is guarded by useModRate, so that should be done too.

mrbuds: Changing nextTime was the first thing i tried with the following change, but it's wrong, didn't work correctly

```Lua
    if (cType == "timer" and value) then
      if useModRate then
        recheckCode = "  nextTime = state[" .. trigger .. "] and not state[" .. trigger .. "].paused"
        .. " and state[" .. trigger .. "]" .. string.format("[%q]",  variable)
        .. " and (state[" .. trigger .. "]" .. string.format("[%q]",  variable) .. " - " .. value .. ") / (state[" .. trigger .. "].modRate or 1.0)\n"
      else
        recheckCode = "  nextTime = state[" .. trigger .. "] and not state[" .. trigger .. "].paused"
        .. " and state[" .. trigger .. "]" .. string.format("[%q]",  variable)
        .. " and (state[" .. trigger .. "]" .. string.format("[%q]",  variable) .. " - " .. value .. ")\n"
      end
      recheckCode = recheckCode .. "  if (nextTime and (not recheckTime or nextTime < recheckTime) and nextTime >= now) then\n"
      recheckCode = recheckCode .. "    recheckTime = nextTime\n";
      recheckCode = recheckCode .. "  end\n"
```

InfusOnWoW: Becasue:
**(state[" .. trigger .. "]" .. string.format("[%q]",  variable)** is the expirationTime and **value**  the time threshold time.

You need to modify the threshold time by multiplying it with modRate.

Consider this example:
We are at t = 100.
The expirationTime is 150.
The remaining time check is 10.
The modRate is 2.0 (that is time ticks twice as fast)

Then the check must run at 
expirationTime - (modRate * threshold)
=> 150 - 10 * 2
=> 130

mrbuds: You are right, pushed updated commit

PR #4639, opened 2023-10-09
fix

mrbuds:


WeakAuras/Prototypes.lua
```diff
@@ -1024,6 +1024,7 @@ function WeakAuras.IsPlayerSpellOrOverridesAndBaseIsPlayerSpell(spell)
 end
 
 function WeakAuras.IsSpellKnownForLoad(spell, exact)
+  spell = tonumber(spell) or spell
```
InfusOnWoW, line 1027: I think we should revert the breaking change, looking at it again it looks to be wrong.

mrbuds: done

InfusOnWoW (approved):

PR #4662, opened 2023-10-18
fix

mrbuds:
This use font loading workaround use by https://www.curseforge.com/wow/addons/libddi-1-0

I'm not 100% sure it's correct to use same objectName with potentially different settings for same font, but it would be bad to create too much of these objects


InfusOnWoW:
I can't reproduce the font issue, the font and also another shipped with prat just work correctly. Probably has something to do with how fast the fonts can be loaded.

InfusOnWoW (approved):

PR #4676, opened 2023-10-28
Colorize "Class and Specialization" options fixes #4657

### LINKED ISSUE #4657: Class and Specialization List - Difficult to Read - Class color text?
The dropdown lists for Class and Specialization are hard to read quickly. It's a very long list of text, and the class/spec icons are very small. For my eyes at least, the two icons bleed together because they're so small and close together. The colors do not stand out so I have a hard time quickly scanning for the spec I'm looking for. I end up having to scroll back and forth till I find it, the icons aren't big enough to jump out. All the required information is present, it's just hard to read. 

This also applies to the multi-selection check box list (In the Load tab). However, the checkbox list is _slightly_ easier to read because the options are slightly larger and are more separated. I still experience difficulty quickly finding what I'm looking for, but it's not as bad as the dropdown. 

Could the text of the spec names in the dropdown be class colors? (This wouldn't help colorblind folks unfortunately, I imagine they have an even harder time scanning the list with their eyes). I feel that class colors here would make each entry have a quick standout association. This may have problems contrasting with the background, I'm unsure, but that's the best thing I can think of. 

Icons or text size being larger in the dropdown is also an option, but I think that coloring each row with text color would be better. 

**Additional context**
image
image

mrbuds:
image

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -1398,8 +1398,9 @@ local function update_specs()
       if tabName then
         tinsert(WeakAuras.spec_types_specific[classFileName], "|T"..(icon or "error")..":0|t "..(tabName or "error"));
         if WeakAuras.IsRetail() then
+          local classColor = WA_GetClassColor(classFileName)
           Private.spec_types_all[specId] = CreateAtlasMarkup(GetClassAtlas(classFileName:lower()))
-          .. "|T"..(icon or "error")..":0|t "..(tabName or "error");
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
That looks better than I expected. Still, for the currency PR I wanted to experiment with headers in a drop down, which we then could use here too. I'll probably do that this weekend.

InfusOnWoW (approved):

PR #4680, opened 2023-10-30
fix

mrbuds:


WeakAuras/Profiling.lua
```diff
@@ -394,9 +412,25 @@ function WeakAuras.PrintProfile()
...
+  for k, v in pairs(profileData.systems) do
+    local event = unitEventToMultiUnit(k)
+    if systemRegrouped[event] == nil then
+      systemRegrouped[event] = CopyTable(v)
+    else
+      if v.elapsed then
+        systemRegrouped[event].elapsed = (systemRegrouped[event].elapsed or 0) + v.elapsed
+      end
+      if v.spike then
+        systemRegrouped[event].elapsed = (systemRegrouped[event].spike or 0) + v.spike
```
InfusOnWoW, line 426: you are assigning to elapsed here, should be spike

WeakAuras/Prototypes.lua
```diff
@@ -7966,7 +7968,7 @@ Private.event_prototypes = {
           threatpct, rawthreatpct, threatvalue, threattotal = 100, 100, 0, 100
         end
       ]];
-      return ret .. unitHelperFunctions.SpecificUnitCheck(trigger);
+      return ret:format(trigger.unit == "group" and "true" or "false") .. unitHelperFunctions.SpecificUnitCheck(trigger);
```
InfusOnWoW, line 7971: The Threat Situation trigger doesn't support the "group" unit type.

WeakAuras/Prototypes.lua
```diff
@@ -8036,6 +8038,58 @@ Private.event_prototypes = {
...
+      {
+        name = "name",
+        display = L["Unit Name"],
+        type = "string",
+        store = true,
+        hidden = true,
+        test = "true"
+      },
+      {
+        name = "realm",
```
InfusOnWoW, line 8050: Can this ever have a value? Since you can only have threat on npcs

InfusOnWoW: Actually due to how UnitNameWithRelam works, it probably always the realm of the player.

WeakAuras/Prototypes.lua
```diff
@@ -7955,6 +7955,8 @@ Private.event_prototypes = {
       trigger.unit = trigger.unit or "target";
       local ret = [[
         unit = string.lower(unit)
+        local name, realm = WeakAuras.UnitNameWithRealm(unit)
```
InfusOnWoW, line 7958: We can use use UnitName here, since I think we don't need to realm at all.

WeakAuras/Prototypes.lua
```diff
@@ -7955,6 +7955,8 @@ Private.event_prototypes = {
...
+        local smart = %s
```
InfusOnWoW, line 7959: You can simplfy this, smart is always false since there's no "group" unit (nor will there ever be one)

InfusOnWoW (approved):

PR #4685, opened 2023-11-02
Adding coloring for WeakAuras that never load

m33shoq:
# Description

So my change is that WeakAuras lines are now colored red if they have load "Never" cheked

image

Current way to identify is WeakAura loaded(look for small papersheet on the display button) is not very obvious and informative. There is often a WeakAuras that must be loaded on certain bosses and they will always be shown as unloaded, so abillity to quickly identify is WA set to never load should be a good addition.

I saw this feature on Echo WeakAura's fork during RWF so inspired to make it real for everyone.
(new to github and english is not my primary language so sorry if my description is not good enough or i made something wrong)

## Type of change

## How Has This Been Tested

Tested on my own WAs
Tried creating new WA and set it to load never

## Checklist

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -1138,6 +1138,12 @@ function OptionsPrivate.SortDisplayButtons(filter, overrideReset, id)
       child:DisableLoaded();
     end
 
+    if child.data.load.use_never then
+      child.background:SetVertexColor(1, 0.12, 0.12, 0.25)
+    else
+      child.background:SetVertexColor(0.5, 0.5, 0.5, 0.25)
+    end
+
     if useTextFilter then
       if(id:lower():find(filter, 1, true)) then
         aurasMatchingFilter[id] = true


InfusOnWoW:
That's the wrong approach for several reasons:

Red signifies some kind of error, an never loaded aura is not an error.
Never loaded is overall one of the least used load options, it should not have the most aggressive display. 
This solves a narrow use case for one specific situation, while ignoring the wider problems with the display of the loaded auras.

Both the sorting of auras into Loaded/Unloaded and the load indicator were designed in a very differnt era and both fail in various ways. Not just in the way you describe where for a certain encounter you have a number of "Never loaded" auras. But we also get frequently confussed users that ask why a certain aura is in the unloaded section, e.g. for auras that are set to load in M+, zones etc.

The aim with the "Unloaded" vs "Loaded" distinction was to display the "relevant" auras for the user. Clearly with how many relevant auras end up in "Unloaded", this is no longer working well and needs a better approach. Adding a hack for a use case that is only relevant to a small minority of users is heading into the wrong direction. And then we are stuck with highlihting "Never load".

Circling back, the problem is not that an "never loaded" aura is displayed as "unloaded", that's correct. But that the other auras in the group, that are only loaded on demand are displayed as "unloaded". These auras are relevant, and that they can't be distinguished from "never loaded" is the problem.

PR #4710, opened 2023-11-14
fix

mrbuds:


InfusOnWoW (approved):

WeakAuras/WeakAuras.lua
```diff
@@ -6004,6 +6002,10 @@ do
     return Traverse(data, true, false, true)
   end
 
+  local function TraverseAuras(data)
+    return Traverse(data, false, false, false)
```
InfusOnWoW, line 6006: I think that can't be right, this is includeSelf = false, includeGroups = false and includeLeafs = false.

The latter two are passed to shouldInclude, making shouldInclude always return false.

I think you want actually want false, false, true, which is exactly TraveseLeafs.

mrbuds: Indeed, edited

InfusOnWoW (approved):

PR #4717, opened 2023-11-17
Add effective combo points trigger

weswigham:
# Description

This adds a new trigger kind for the "effective" combo point value the player currently has, considering charged combo points. This makes writing combo point trigger logic using charged combo points *much* simpler. (This number is what most of the APL in simc is written against)

## Type of change

## How Has This Been Tested

- Made a weak aura with the new trigger and a `>= 6` condition. Used echoing reprimand to charge combo points, and got to 3 combo points, and verified that the weak aura appears (since 3 becomes effectively 7 when charged).
<details><summary>WA Import Code</summary>
!WA:2!1nvqUTTruulOffGOX1vUWO1PleCaCrAacqdsmCqTDHPTyKluIvOOQDxrpKZNItk5mdMzOTLxkuGKnzJpcADxPJqobec9eOfDvw5tq(dPbsrrT2W)F0BM379N30WRvElAl67(bLOWKX4G64Hb9o6vDuxECsIgm3t(iktlZiJdGlnHjcvoXekdnSCiKoMtYzXHMufOtfz0XZ1PeQ4IFVEZF)sZi84uHQVGXn((h9IUbY2395v31qUsCH2iYTigkPed4dNdCZcsbUkXWIpHrnPU7JTksSHj46T81gIY4eLW4mDQJl(X4mXOyJgbk9YBQUT8FCnJLGBb(3YwfAiKePfQOUaj7LckC2cDrufBdkssyxol8G9heeoiyF)GfsXfGkuibfbzE9n2B3RTNq1YNv6ixRtscG65CO9bI8ir7kFRl)MiRHb1IpZxfxt0sil7iQ2XhhKG2zIfGvAFwg9vakd)b970RNFftR3ARPuicLxa6ev3o967nS31f8BDOJ8RjvIiOU3s1FTKCJ7EUNlOHOLGZKBE3GYGrK4XHjzcHk9AnKLu5o3ED8cCteCJCfpf7Q2VUGqXdJ0oiykAcFyK9(z91kxDo2grI)JryyJtDZeeQxKHKHUCn)8ImdZXfNiXRv(to(XzeT2w5Qzxb2IjwsgGnF3TPStRtznNQQ4WonCnOWJIf8e2iNRjCMnUi4BvUS3haIggyuaFKj9lkx1Ll4WCAHQcrOnvO0aUvQEIfQ94k3XnNW4EL)mUHYDk3TCVYFb)(v)3vKp4UhCsfeZ0ihnL3))fvDHVfmDQvaSQinodezc1V2e)PEd(IGLmU8hNAeXNJkfr8TnEB6gm6uVQiF7xjUyYfcf9efro5KBlMLOWKfAB8gP5n1Vg9SlzT3t67FOAC9u8Jl9KcgDw6ZlE6d32n6WHVyo(CdrFSSwnZ(xAVS7S6RGdQuyJgnAoftPaT6H5F(LZy8AtH705gmLakoj73QL97BP9F2J3(XpR15)9PFk
</details>

https://github.com/WeakAuras/WeakAuras2/assets/2932786/bfdfe1a8-65ef-45de-aec4-03e5042822e4

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -7261,6 +7261,114 @@ Private.event_prototypes = {
     delayEvents = true,
     timedrequired = true
   },
+  ["Effective Combo Points"] = {
+    type = "unit",
+    canHaveDuration = false,
+    events = function(trigger)
+      local unit = "player"
+      local result = {}
+      AddUnitEventForEvents(result, unit, "UNIT_POWER_FREQUENT")
+      AddUnitEventForEvents(result, unit, "UNIT_MAXPOWER")
+      AddUnitEventForEvents(result, unit, "UNIT_DISPLAYPOWER")
+      AddUnitEventForEvents(result, unit, "UNIT_NAME_UPDATE")
+
+      if WeakAuras.IsRetail() then
+        AddUnitEventForEvents(result, unit, "UNIT_POWER_POINT_CHARGE")
+      else
+        AddUnitEventForEvents(result, unit, "UNIT_TARGET")
+      end
+      return result;
+    end,
+    internal_events = {},
+    force_events = {{"UNIT_CHANGED_player", "player"}},
+    name = L["Effective Combo Points"],
+    init = function(trigger)
+      return ""
+    end,
+    statesParameter = "one",
+    args = {
+      {
+        name = "power",
+        display = L["Power"],
+        type = "number",
+        init = not WeakAuras.IsRetail() and "powerType == 4 and GetComboPoints('player', 'player-target') or UnitPower('player', 4)"
+        or "(tContains(GetUnitChargedPowerPoints('player') or {}, UnitPower('player', 4)) and 7 or UnitPower('player', 4)) / WeakAuras.UnitPowerDisplayMod(4)",
+        store = true,
+        conditionType = "number",
+        multiEntry = {
+          operator = "and",
+          limit = 2
+        },
+      },
+      {
+        name = "value",
+        hidden = true,
+        init = "power",
+        store = true,
+        test = "true"
+      },
+      {
+        name = "total",
+        hidden = true,
+        init = not WeakAuras.IsRetail() and "(math.max(1, UnitPowerMax('player', 4))"
+                                      or "math.max(1, UnitPowerMax('player', 4)) / WeakAuras.UnitPowerDisplayMod(4)",
+        store = true,
+        test = "true"
+      },
+      {
+        name = "stacks",
+        hidden = true,
+        init = "power",
+        store = true,
+        test = "true"
+      },
+      {
+        name = "progressType",
+        hidden = true,
+        init = "'static'",
+        store = true,
+        test = "true"
+      },
+      {
+        name = "percentpower",
+        display = L["Power (%)"],
+        type = "number",
+        init = "total ~= 0 and (value / total) * 100 or nil",
+        store = true,
+        conditionType = "number",
+        multiEntry = {
+          operator = "and",
+          limit = 2
+        },
+      },
+      {
+        name = "deficit",
+        display = L["Power Deficit"],
+        type = "number",
+        init = "total - value",
+        store = true,
+        conditionType = "number",
+        multiEntry = {
+          operator = "and",
+          limit = 2
+        },
+      },
+      {
+        name = "maxpower",
+        display = WeakAuras.newFeatureString .. L["Max Power"],
+        type = "number",
+        init = "total",
+        store = true,
+        conditionType = "number",
+        multiEntry = {
+          operator = "and",
+          limit = 2
+        },
+      },
+    },
+    overlayFuncs = {},
+    automaticrequired = true
+  },
   ["Evoker Essence"] = {
     type = "unit",
     events = {},
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -3192,12 +3192,13 @@ Private.event_prototypes = {
         local name, realm = WeakAuras.UnitNameWithRealm(unit)
         local smart = %s
         local powerType = %s;
+        local measureChargedComboPoints = %s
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
What's the use case for this? Is the use case converting an APL to a WeakAuras setup? Because that's not something I would consider relevant for WA.

weswigham:
The use case is just for reminders like the one above - in fact, for rogues, almost everywhere they'd use a combo points condition, this is probably moreso what they want (the only time you'd want "real" combo points is for a slice and dice reminder or a combo points resource bar). Technically you can emulate this with a ton of triggers monitoring echoing reprimand buffs and combo points counts, but this lets you use one trigger instead of 7 and a custom trigger combinator.

Stanzilla:
Hm and it would not make sense to have this as an option on the normal combo points trigger?

weswigham:
This seems inline with the top-level Death Knight Rune and Essence power triggers, imo. Especially since this can't reasonably be combined with a lot of the options available on the base power trigger; but yeah, it *could* be an option, just with a lot more checks scattered around.

Stanzilla:
Yeah I was just thinking out loud. You have to excuse us, we all don't play rogue so our perspective here is a bit limited, mine was only trying to see it from a end user perspective, they would probably be looking at/for the main combo points trigger first and not even consider that there is a second one, you know?

weswigham:
Maybe, I'd feel that reading "effective combo points" in the top level trigger list is probably just as visible, though. It also feels pretty weird to clutter the general power options list with options that only apply to one power, and in this case, for one class (since feral druids don't have charged combo points).

mooklepticon:
Good timing. I was literally trying to do this last night. I found a few on wago, but nothing like what I wanted. I typically create a priority list, like to use envenom.
1) Action usable: envenom.
2) Combo points >4.

With Echoing Reprimand, I **want** to create this:
1) Action usable: envenom.
2) Combo points == an anima charge combo point.

I found some on Wago that show you when you have one, but don't really seem to make an elegant way to show an aura if it's the right number.  There are some that seem to try, but I couldn't get it to work with the above style.  I ended up doing this.

Aura 1: Envenom usable and CP>4
Aura 2: Envenom usable, and Echoing Rep = 3, and CP=3
Aura 3: Envenom usable, and Echoing Rep = 4, and CP=4
Aura 4: Envenom usable, and Echoing Rep = 5, and CP=5

It's tedious, but it works. The tricky part was figuring out the different ER spell ID's because using the name alone doesn't work.

mrbuds:
> It also feels pretty weird to clutter the general power options list with options that only apply to one power

It's possible to show an option only if power type = combo points, look at the "Overlay Charged Combo Points" option

InfusOnWoW:
I don't like the addition of a new trigger just for one talent either, and especially not that there would be two ways to track combo points in slightly different ways in a non-obvious setting.

So this has to be either:
* Not included
* An Option in the Power trigger
* Moving the Combo Point trigger out of the "Power" trigger, and then an option.

weswigham:
> So this has to be either:
>
> * Not included

🙁

> * Moving the Combo Point trigger out of the "Power" trigger, and then an option.

Maybe it should be at this point - combo points are a lot of special cases in there - but it'd break a lot of existing WAs, so maybe not?

> * An Option in the Power trigger

Doable. And now done, tested similarly as in the OP.

<summary>New test WA import code
<details>
```
!WA:2!1nvZUnsrq4yzHqm8xwFic8csrbPfHqkhwbwlhaHhN4nbz2ygpzZYPX9mDnEAyMUBvDpjX7rlKyVWf)i4ZCYpc8emYAFcY590(eSvpJvweI4ltvL)QQ)QVQQwd7u0H3H)IVevL2CHeWZopC0Pp5y86ZstnG9dxZfgDoBEiCTD1qHuyY2)jQR2yYyC1v)AdQpBN1mzsMchResBqWPp(Kq9()RmJsvybZgPJSIcyRxl9UjLgRQWH4CnNzHa4sqAVHvsrzwrYfcUnZVp5ISeRqjn9cmwgA9ItR5INp9X6TWIIzZa08rpa3A(I7DtPbIyXgfg)ZkomDHZVKGpDPwDfG25A4924cER70vCiUmnnKSXtoE04HNpY39h(U80DEBfpby51v9gtzCnRNq5jUED0G(tcJMe2pi8M66gPOAXOoO7b)W3x9(bJDbdKScW4TWOH88t5gVBlZyeOYemz8XJgfuxGUD6T8wwov)5o7cGzkryqgdNb8bQIyvT2BMwD)yNQd4Ys5wPWtFpN6Dje24749FVJ(G7EavO4reNHP6hC3GYHzSK5rP5kfMT0a5P1uWF0Xdd9tvsRE3HO457)lLmovm2(HHROUmaM5gKD3RQ7gYnML87ZO1pj3pxX4pTUvLKuGzXwwoPi7fuuMBfEbj5mJzVQEE(KSL0yiEo4mw4EVjKZNUDZ8znBMTxH1pNBG6BPEq)13DdXNttfrsKndbtMkNpFjtkClIkzVQpE4)qIomXIGCMn7DR66lvsydVeRre52tqdKOKCZchu3Jw13VGjKdR(rkHQ(v(vdQoI(U7)nI(lUBIPriryO3OT(()VOAmcCG5RCequFS4T4kfYVaz6fxS1a)n6KtKoV6WvwvcPYUY(jT(ZSde8Q3bN3iBVAN1PiTHsnln4A)6MR7HUqUM6HJdokMEMuXmVhwk4RZ(UYV5REKF8rN)4n0XlH9mDddcsu5k8NAt)w3mzguhOvRwTxrl6aV(g)p(G1cztxqP59AApcqjl)Pne8V6yc(2dF0H96C5lF2B(d
```
</details>
</summary>

image

InfusOnWoW:
A slightly different implementation was merged, which does accomplish the same thing in a imho better way.

Stanzilla:
@weswigham please let us know how that new option works for you, it's in the alpha builds already!

PR #4743, opened 2023-12-05
currencytrigger fixes

jjholleman:
fixes:
> If a currency group header is collapsed on the Currency tab, the dropdown list doesn't contain any items from that group


jjholleman:
I'd like to check if I can also fix new currencies discovered not showing up until reload, but that will be much later today.

PR #4750, opened 2023-12-07
add lootspec trigger

jjholleman:
# Description

Adds a Loot Specialization trigger option under Player/Unit info > Conditions.

## Type of change


WeakAuras/Prototypes.lua
```diff
@@ -9768,6 +9768,11 @@ Private.event_prototypes = {
         tinsert(events, "PLAYER_DIFFICULTY_CHANGED")
       end
 
+      if trigger.use_loot_specialization_id ~= nil then
+        tinsert(events, "PLAYER_LOOT_SPEC_UPDATED")
+        tinsert(events, "PLAYER_TALENT_UPDATE")
```
InfusOnWoW, line 9773: I don't think PLAYER_TALENT_UPDATE is needed

WeakAuras/Prototypes.lua
```diff
@@ -9903,6 +9909,25 @@ Private.event_prototypes = {
         enable = WeakAuras.IsRetail(),
         hidden = not WeakAuras.IsRetail(),
       },
+      {
+        name = "loot_specialization_id",
+        display = L["Loot Specialization"],
+        type = "multiselect",
+        init = "WeakAuras.IsRetail() and (GetLootSpecialization() == 0 and GetSpecializationInfo(GetSpecialization()) or GetLootSpecialization())",
```
InfusOnWoW, line 9916: you don't need to check for IsRetail here.

WeakAuras/Prototypes.lua
```diff
@@ -9903,6 +9909,25 @@ Private.event_prototypes = {
...
+        values = "spec_types_all",
+        enable = WeakAuras.IsRetail(),
+        hidden = not WeakAuras.IsRetail(),
+      },
+      {
+        name = "lootSpec",
+        display = L["Loot Specialization"],
+        type = "string",
+        init = "WeakAuras.IsRetail() and select(2, GetSpecializationInfoByID(loot_specialization_id))",
```
InfusOnWoW, line 9925: nor here.

WeakAuras/Prototypes.lua
jjholleman: That should have been PLAYER_SPECIALIZATION_CHANGED

InfusOnWoW:
Eh, I guess there are a few people asking sometimes for that...

WeakAuras/Prototypes.lua
```diff
@@ -9903,6 +9909,25 @@ Private.event_prototypes = {
...
+      {
+        name = "lootSpec",
+        display = L["Loot Specialization"],
+        type = "string",
+        init = "select(2, GetSpecializationInfoByID(loot_specialization_id))",
+        enable = "loot_specialization_id",
+        hidden = true,
+        store = true,
+        test = "true",
+      },
```
mrbuds, line 9930: this needs a 

        enable = WeakAuras.IsRetail(),
        hidden = not WeakAuras.IsRetail(),
        
getting this error on SoD
```
[string ""]:14: attempt to call global 'GetSpecializationInfoByID' (a nil value)
[string ""]:14: in function <[string ""]:2>
[string "=[C]"]: in function `xpcall'
[string "@WeakAuras/GenericTrigger.lua"]:707: in function <WeakAuras/GenericTrigger.lua:630>
[string "@WeakAuras/GenericTrigger.lua"]:1027: in function `ScanWithFakeEvent'
[string "@WeakAuras/GenericTrigger.lua"]:1370: in function `LoadDisplays'
[string "@WeakAuras/WeakAuras.lua"]:1939: in function `LoadDisplays'
[string "@WeakAuras/WeakAuras.lua"]:1723: in function <WeakAuras/WeakAuras.lua:1567>
[string "@WeakAuras/WeakAuras.lua"]:1916: in function `Resume'
[string "@WeakAurasOptions/OptionsFrames/OptionsFrame.lua"]:170: in function <...dOns/WeakAurasOptions/OptionsFrames/OptionsFrame.lua:150>
[string "=[C]"]: in function `Hide'
[string "@Interface_Vanilla/FrameXML/UIParent.lua"]:2665: in function `HideUIPanel'
[string "@SharedXML/SharedUIPanelTemplates.lua"]:213: in function `HideParentPanel'
```

WeakAuras/Prototypes.lua
```diff
@@ -9903,6 +9909,25 @@ Private.event_prototypes = {
...
+        values = "spec_types_all",
+        enable = WeakAuras.IsRetail(),
+        hidden = not WeakAuras.IsRetail(),
+      },
+      {
+        name = "lootSpec",
+        display = L["Loot Specialization"],
+        type = "string",
+        init = "select(2, GetSpecializationInfoByID(loot_specialization_id))",
+        enable = "loot_specialization_id",
```
mrbuds, line 9926: wut

jjholleman:
@mrbuds I've cleaned up the mess now. 
I found there is a "Class and Specialization" trigger and moved the lootspec option there. I think it fits there better than under conditions. It also meant I could remove the IsRetail checks from the args table.

mrbuds:
> @mrbuds I've cleaned up the mess now. I found there is a "Class and Specialization" trigger and moved the lootspec option there. I think it fits there better than under conditions. It also meant I could remove the IsRetail checks from the args table.

I think it's better to have it in a separate trigger because there is no mechanism to have 2 icons from same trigger and icon for spec can be different from icon for loot spec

InfusOnWoW:
I'm not too enthusiastic about the trigger in general, though people are asking about a lootspec trigger somewhat regularly. So there's enough demand that rejecting the contribution wouldn't be right.

All of adding a separate trigger, adding it to the class/spec or conditions, feel non-ideal to me. Though I think I also prefer a separate trigger. Then again, I don't have a strong opinion either way.

jjholleman:
There are a couple of different approaches:

- Make Loot Specialization its own trigger. That would increase the list of triggers, I'm not sure if that really is a problem.
- Keep it under Class and Specialization, but overwrite the icon only if a lootspec filter is chosen. This could be seen as weird behaviour for users because its not transparent which option supplies the icon.
- Move it under Conditions, with or without icon.

Not any approach is perfect, although from a user pov I think option 1 is fine.

InfusOnWoW:
@jjholleman Can we expect an update to the PR with a standalone trigger?

jjholleman:
> @jjholleman Can we expect an update to the PR with a standalone trigger?

Are you spying on me? :D

WeakAuras/Prototypes.lua
```diff
@@ -6464,6 +6464,63 @@ Private.event_prototypes = {
...
+         end
+         local _, lootSpecName, _, lootSpecIcon = GetSpecializationInfoByID(lootSpecId)
+      ]]
+    end,
+    args = {
+      {
+        name = "name",
+        display = L["Loot Specialization"],
+        type = "multiselect",
+        init = "lootSpecId",
```
InfusOnWoW, line 6492: This stores the lootSpecId, since that's needed for the test. I suggest changing the name of this to "lootSpecId" and adjusting the "lootSpecId" below to be actually initialized by lootSpecName and be named "name".

PR #4755, opened 2023-12-10
Currency Trigger: support classic wrath, disable on classic era

jjholleman:
Classic Wrath's API for CurrencyInfo is not the same as on Retail. 
On Classic Era, currencies don't exist.

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -1324,29 +1324,31 @@ Private.GetCurrencyListInfo = function(index)
   if WeakAuras.IsRetail() then
     return C_CurrencyInfo.GetCurrencyListInfo(index)
   elseif WeakAuras.IsWrathClassic() then
-    local name, isHeader, isExpanded, isUnused, isWatched, _, icon, _, hasWeeklyLimit, _, _, itemID = GetCurrencyListInfo(index)
-    local _, currentAmount, _, earnedThisWeek, weeklyMax, totalMax, isDiscovered, rarity = itemID and GetCurrencyInfo(itemID)
-    local currencyInfo = {
-      name = name,
-      description = "",
-      isHeader = isHeader,
-      isHeaderExpanded = isExpanded,
-      isTypeUnused = isUnused,
-      isShowInBackpack = isWatched,
-      quantity = currentAmount,
-      trackedQuantity = 0,
-      iconFileID = icon,
-      maxQuantity = totalMax,
-      canEarnPerWeek = hasWeeklyLimit,
-      quantityEarnedThisWeek = earnedThisWeek,
-      isTradeable = false,
-      quality = rarity,
-      maxWeeklyQuantity = weeklyMax,
-      totalEarned = 0,
-      discovered = isDiscovered,
-      useTotalEarnedForMaxQty = false,
-    }
-    return currencyInfo
+    local _, isHeader, isExpanded, isUnused, isWatched, _, icon, _, hasWeeklyLimit, _, _, itemID = GetCurrencyListInfo(index)
+    local name, currentAmount, _, earnedThisWeek, weeklyMax, totalMax, isDiscovered, rarity = itemID and GetCurrencyInfo(itemID)
+    if name then
+      local currencyInfo = {
+        name = name,
+        description = "",
+        isHeader = isHeader,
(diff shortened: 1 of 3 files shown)


WeakAuras/Types.lua
```diff
@@ -1320,6 +1320,66 @@ for id, str in pairs(Private.combatlog_spell_school_types) do
...
+    return C_CurrencyInfo.GetCurrencyListSize()
+  end
+  if WeakAuras.IsWrathClassic() then
+    return GetCurrencyListSize()
+  end
+end
+
+Private.ExpandCurrencyList = function(index, expand)
+  if WeakAuras.IsRetail() then
+    return C_CurrencyInfo.GetCurrencyListLink(index, expand)
```
mrbuds, line 1366: this seems wrong, shouldn't it be C_CurrencyInfo.ExpandCurrencyList?

jjholleman: yes. 👀

WeakAuras/Types.lua
```diff
@@ -1320,6 +1320,66 @@ for id, str in pairs(Private.combatlog_spell_school_types) do
...
+        totalEarned = 0,
+        discovered = isDiscovered,
+        useTotalEarnedForMaxQty = false,
+      }
+      return currencyInfo
+    end
+  end
+end
+
+Private.GetCurrencyListSize = function()
```
InfusOnWoW, line 1355: You can simplfy here by:
 if WeakAuras.IsRetail() then
     Private.GetCurrencyListSize = C_CurrencyInfo.GetCurrencyListSize()
  elseif WeakAuras.IsWrathClassic() then
    Private.GetCurrencyListSize = GetCurrencyListSize()
  end

jjholleman: This results in errors on wrath and wrong behaviour on retail

InfusOnWoW: Huh, why? the API is exactly the same, you call with no arguments and get returned the size.

jjholleman:
```
if WeakAuras.IsRetail() then
  Private.GetCurrencyListSize = C_CurrencyInfo.GetCurrencyListSize()
elseif WeakAuras.IsWrathClassic() then
  Private.GetCurrencyListSize = GetCurrencyListSize()
end
```
`Private.CurrencyListSize` will always have the same value after this block. But the list size changes when there are collapsed headers and we expand/collapse them.

With `Private.CurrencyListSize()` the value will always be current.

InfusOnWoW: Eh right, typo in there:
```
if WeakAuras.IsRetail() then
  Private.GetCurrencyListSize = C_CurrencyInfo.GetCurrencyListSize
elseif WeakAuras.IsWrathClassic() then
  Private.GetCurrencyListSize = GetCurrencyListSize
end
```

WeakAuras/Types.lua
```diff
@@ -1320,6 +1320,66 @@ for id, str in pairs(Private.combatlog_spell_school_types) do
...
+Private.ExpandCurrencyList = function(index, expand)
+  if WeakAuras.IsRetail() then
+    return C_CurrencyInfo.ExpandCurrencyList(index, expand)
+  end
+  if WeakAuras.IsWrathClassic() then
+    return ExpandCurrencyList(index, expand)
+  end
+end
+
+Private.GetCurrencyIDFromLink = function(currencyLink)
```
InfusOnWoW, line 1373: Also here, change it to:

```
if WeakAuras.IsRetail() then
  Private.GetCurrencyIDFromLink = C_CurrencyInfo.GetCurrencyIDFromLink
elseif WeakAuras.IsWrathClassic() then
  Private.GetCurrencyIDFromLink = function(currencyLink)
    local currencyID = string.match(currencyLink, "|Hcurrency:(%d+):")
    return currencyID
  end
end
```

WeakAuras/Types.lua
jjholleman: Yep, that's it. As a reward it spit out more bugs to fix 😁

mrbuds: roo roo

InfusOnWoW (approved):

PR #4767, opened 2023-12-18
Update flipbooks for StopMotion

Nightwarden24:
# Description

Includes the following changes:

- Add new flipbooks and fix some others where the data was incorrect
- Make the flipbook display proportional in the texture picker
- Set region size to match the flipbook tile size when selecting a texture
- Fix StopMotion thumbnail

## Type of change

### DIFF
--- a/WeakAurasOptions/OptionsFrames/TexturePicker.lua
+++ b/WeakAurasOptions/OptionsFrames/TexturePicker.lua
@@ -234,7 +234,7 @@ local function ConstructTexturePicker(frame)
         pickedwidget = widget;
       end
     end
-    local width, height
+    local width, height, flipbookInfo
     if(pickedwidget) then
       pickedwidget:Pick();
       if not pickedwidget.texture.IsStopMotion then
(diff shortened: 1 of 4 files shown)


WeakAurasOptions/OptionsFrames/TexturePicker.lua
```diff
@@ -243,13 +243,19 @@ local function ConstructTexturePicker(frame)
...
+          height = flipbookInfo.tileHeight
+        end
       end
     end
 
     wipe(group.selectedTextures)
     group.selectedTextures[texturePath] = true
 
-    SetAll(self.baseObject, self.path, self.properties.texture, texturePath, width, height, self.adjustSize)
+    SetAll(self.baseObject, self.path, self.properties.texture, texturePath, width, height, flipbookInfo and true or self.adjustSize)
```
InfusOnWoW, line 258: I think this change here is the wrong place to add that. Keep this line as is, and instead modify the calls to OpenTexturePicker to pass in true as the last parameter.

InfusOnWoW (approved):

PR #4818, opened 2024-01-20
Add dynamic information last active trigger

axondk:
# Description

I've added the option to get dynamic information from the last active trigger.
This helps to mantain always the same order when using multiple triggers and makes editing group conditions much easier.

Thanks for your amaizing work.

## Type of change


mrbuds:
I fail to imagine a use case this solve that wouldn't be solved by properly ordering triggers

InfusOnWoW:
That's fundamentally against how WA should work. The order in which triggers activates should never matter.

PR #4819, opened 2024-01-20
Add status icon system for display buttons

mrbuds:
# Description

This allow display buttons to have a list of status icons
Loaded icon and Warning icons are moved to this list
Groups shows an icon per unique status icon of each leaf

This solve a few ux issues:
- warning icon could show only one warning, now it can show multiple warning on same button
- warnings were not spread to parent group(s), user had to expand them
- with nested groups this help identify which ones have auras with different loaded status

https://github.com/WeakAuras/WeakAuras2/assets/189656/02894413-7f6e-44de-bdab-e9d364ebe07e

### DIFF
--- a/WeakAurasOptions/WeakAurasOptions.lua
+++ b/WeakAurasOptions/WeakAurasOptions.lua
@@ -101,6 +101,7 @@ function OptionsPrivate.DuplicateAura(data, newParent, massEdit, targetIndex)
       if not massEdit then
         local button = OptionsPrivate.GetDisplayButton(parentData.id)
         button.callbacks.UpdateExpandButton()
+        button:UpdateParentWarning()
       end
       OptionsPrivate.ClearOptions(parentData.id)
     end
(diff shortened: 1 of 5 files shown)


WeakAurasOptions/WeakAurasOptions.lua
```diff
@@ -579,6 +579,12 @@ function WeakAuras.ToggleOptions(msg, Private)
         -- The button does not yet exists if a new aura is created
         displayButtons[id]:UpdateWarning()
       end
+      local data = Private.GetDataByUID(uid)
+      for parent in Private.TraverseParents(data) do
+        if displayButtons[parent.id] then
+          displayButtons[parent.id]:UpdateWarning()
+        end
+      end
```
InfusOnWoW, line 587: You also need to call UpdateWarning() in a few more places, which I think can all be identified by looking for UpdateExpandButton. Though do note that UpdateExpandButton needs to be only called on the direct parent, where as UpdateWarning needs to be called on each parent

Places that seem to be missing:
* CreateNewGroupFromSelection
* Actions["Group"] (That's for drag/drop)
* Actions["Ungroup"]
* In update.lua, which can move auras around.
* OnClickGrouping (which is the grouping icon on the right)

PR #4824, opened 2024-01-22
Smallpatches

InfusOnWoW:


WeakAuras/WeakAuras.lua
```diff
@@ -2538,11 +2538,22 @@ function Private.AddMany(tbl, takeSnapshots)
...
     if data.regionType == "dynamicgroup" or data.regionType == "group" then
       groups[data] = true
     end
+    coroutine.yield()
+  end
+
+  for data in pairs(groups) do
+    WeakAuras.PreAdd(data)
+    coroutine.yield()
+  end
```
mrbuds, line 2551: Is there a need to call PreAdd twice on groups?

InfusOnWoW: No that's not necessary.

mrbuds (approved):

PR #4826, opened 2024-01-23
Create the atlas files from wago.tools automatically

InfusOnWoW:
Fixes: #4776


InfusOnWoW:
The result of the action can be seen at https://github.com/InfusOnWoW/WeakAuras2/pull/2/files

Checking out [update-atlasfiles](https://github.com/InfusOnWoW/WeakAuras2/tree/update-atlasfiles) allows testing the generated atlas file list. 

I did spot checks of the difference between what this generates and what the old method of naming the atlas files generated, and indeed #4776 is correctly desribing a more accurate algorithm.

mrbuds (approved):

.github/workflows/atlas-update.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Update Atlas files from wago.tools
+        run: |
+          cd .github/scripts/
+          ./wagotools-atlas.sh
+        shell: bash
+
+      - name: Cleanup luarocks
+        run: |
+             rm -rf .luarocks
```
Stanzilla, line 30: why is that needed?

Stanzilla: Ah fair, could have put it in .gitignore but doesn't really matter

PR #4839, opened 2024-01-26
AuthorOptions: Allow the author to enforce a sorted entry list

InfusOnWoW:


WeakAurasOptions/AuthorOptions.lua
```diff
@@ -2121,6 +2132,7 @@ local function addUserModeOption(options, args, data, order, prefix, i)
             end
             WeakAuras.ClearAndUpdateOptions(data.id, true)
           end,
+          sorting = option.sortAlphabetically and OptionsPrivate.Private.SortOrderForValues(values) or nil
```
emptyrivers, line 2135: This seems fine to me, but it looks like that sorter is case sensitive, which might get us a ticket or two from people confused that `Apple` sorts before `abacus`.

InfusOnWoW: That's something I could solve once we get tickets :)

PR #4845, opened 2024-01-29
fetch all history on build

emptyrivers:
so the packager doesn't break

### DIFF
--- a/.github/workflows/build-beta.yml
+++ b/.github/workflows/build-beta.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 500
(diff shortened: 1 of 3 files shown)


.github/workflows/build-beta.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 500
+          fetch-depth: 0
```
emptyrivers, line 18: this prompts the action to also fetch tags, which was apparently dropped in v3 in favor of `fetch-tags: true`. We could add that flag, but eh.

mrbuds (approved):
did not look but i approve

PR #4846, opened 2024-01-29
Player Location: Parent Zone support

jjholleman:
# Description

Adds support for Parent Zone IDs in the Player Location load option.
Parent Zone IDs must be prefixed with 'p'.
E.g. if you want it to load in all child maps of "Dragon Isles" : 'p1978'.

The child maps are gathered recursively meaning it would also load in childs of childs of childs.
E.g. 'p1978' makes it load in Dragon Isles > Thaldraszus > Valdrakken.

The Parent Zone is added to the tooltip, and I've added newlines to the bottomText for better readability.

image

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -5953,6 +5953,12 @@ function Private.ExecEnv.ParseZoneCheck(input)
         local prevChar = input:sub(start - 1, start - 1)
         if prevChar == 'g' or prevChar == 'G' then
           self.zoneGroupIds[id] = true
+        elseif prevChar == 'p' or prevChar == 'P' then
+          self.zoneIds[id] = true
+          local info = C_Map.GetMapChildrenInfo(id, nil, true)
+          for _,childInfo in pairs(info) do
+             self.zoneIds[childInfo.mapID] = true
+          end
         elseif prevChar == 'a' or prevChar == 'A' then
           self.areaNames[C_Map.GetAreaInfo(id)] = true
         elseif prevChar == 'i' or prevChar == 'I' then
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -5953,7 +5953,7 @@ function Private.ExecEnv.ParseZoneCheck(input)
         local prevChar = input:sub(start - 1, start - 1)
         if prevChar == 'g' or prevChar == 'G' then
           self.zoneGroupIds[id] = true
-        elseif prevChar == 'p' or prevChar == 'P' then
+        elseif prevChar == 'c' or prevChar == 'C' then
           self.zoneIds[id] = true
           local info = C_Map.GetMapChildrenInfo(id, nil, true)
           for _,childInfo in pairs(info) do
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -164,7 +164,7 @@ end
 local function get_zoneId_list()
   local currentmap_id = C_Map.GetBestMapForUnit("player")
   local instanceId = select(8, GetInstanceInfo())
-  local bottomText = L["Supports multiple entries, separated by commas. Group Zone IDs must be prefixed with 'g', e.g. g277. Supports Area IDs from https://wago.tools/db2/AreaTable prefixed with 'a'. Supports Instance IDs prefixed with 'i'."]
+  local bottomText = L["Supports multiple entries, separated by commas. \nGroup Zone IDs must be prefixed with 'g', e.g. g277. \nParent Zone IDs must be prefixed with 'p'. \nSupports Area IDs from https://wago.tools/db2/AreaTable prefixed with 'a'. \nSupports Instance IDs prefixed with 'i'."]
   if not instanceId and not currentmap_id then
     return ("%s\n\n%s"):format(Private.get_zoneId_list(), bottomText)
   elseif not currentmap_id then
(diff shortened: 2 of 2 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -5887,6 +5887,12 @@ function Private.ExecEnv.ParseZoneCheck(input)
         local prevChar = input:sub(start - 1, start - 1)
         if prevChar == 'g' or prevChar == 'G' then
           self.zoneGroupIds[id] = true
+        elseif prevChar == 'p' or prevChar == 'P' then
+          self.zoneIds[id] = true
+          local info = C_Map.GetMapChildrenInfo(id, nil, true)
```
InfusOnWoW, line 5892: I wonder how many users would be annoyed if this was the default behaviour.

jjholleman:
@InfusOnWoW do you mean that its recursive or that it includes the parent?

InfusOnWoW:
That entering a zone id automatically includes all children.

jjholleman:
Honestly I'm not sure either. Right now I can't think of a use-case where it would be a problem, since most of the time the childs of childs are either instances, micro dungeons (like caves in the maw and in zereth morthis), or places like Tyrhold Reservoir.
Loading it also in the instances can be annoying, but there is a no-instance load option to filter those out.

InfusOnWoW:
Ah you misunderstood me, I meant:
Instead of introducing a "p" prefix, change the behaviour of the existing plain zone id. So that just entering e.g. "2112", matches recursively, so not just the zone itself but also all its children.

Now, that would obviously be a behaviour change, and we shouldn't do those easily.

jjholleman:
Interesting, I didn't think about that.
Though, it needs to make sense for the user too.
Can we think about situations where we specifically don't want to load on the child maps?

InfusOnWoW:
I'll try to do some reasearch.

jjholleman:
I gave a dev version where childmaps are added to the load automatically to a couple of friends with the only comment: "tell me if you find any weird things".
And yes, basically all of them returned with "Why is my [Insert Profession] aura loading in dungeons?"
Eventually they did find how to stop those auras from loading in dungeons.
None of them could give me the correct reason why those auras were loading in instances.

InfusOnWoW:
@jjholleman sorry I missed your comment. Can you upload what you had and the auras that misbehaved?

jjholleman:
@InfusOnWoW 
[WeakAuras.zip](https://github.com/WeakAuras/WeakAuras2/files/14393628/WeakAuras.zip) version where parent zones are included by default when entering a zone id as `2025`.

Simple herb tracking icon for Saxifrage that would normally only load in the main DF zones:
```
!WA:2!DrvZUTTrqyRQweuwGIy1u7eGCWifTNmm8p15qpvtfPyxOi5qs)tojTKCi5wtTlZUlLKnqbA0LMl9IEe0LEPN0JGFciek6dqouGIEZpafD2LsUXHhwoZSZoZ89TZovAwRFTWAHV9RMsd4mxEUiaE8kZj5QeUOtMIYzslXLDIIKG6WVpL)VRGFZiSaC)J5uMYVEJ2EnCMhW5PH8Hm3H0mO31lvBegdjcsGjsp1rQicLLnLrvw(r4pzI1yLGghdc5N)nIfI)5yQc63M0hE4h9p)9KCj0nGNZujVtM7ddaMYnpkIoAw36h461117ahVSv1UfdmqqdCt4d7W65y8D6ryW2OUoa3E(JfaEEh3JB0QLT6YmWwNXXYmin9OqP187eO3jn)QVauomSYKwxRt4YcTNDock)SuYLGyAi4JfOhgxXHnADCZtAnbZEj4SYwvZhdaVs9xWdHFFLBqUeems6PitGK1VEZfaKDawqbkhcsFjtKqAKHZl(embsIkxquqYuetoqSMG)U1kwFoQ6tcUiwGao8TpiRMcgP6ktiyPFE5nzLShzmAwI4I(e0Hfc2modMC72p5RLz3)9IqDEkx8dvQuP67kTUSOSp4eVozFHXi2)WXqrdoJgQsSpavVPmx0rqOX4BMx2e9Qfv01M9)XCPIgDjcqbxHGZdnA3Ut7gfRlqCP9zQXXatzuf)MSaemv29BkOxTXlZjHiXq2WZ7oyFzMMBSnKlcptqYgF2cHYcyavs9tHElW871ONT6rTB3WPRDhpVoVW5ONFO38BtTl9k4rlq053jp6Dn9bDoXR1rTB8xZJt5dBkGxNdSGlp(J3ERD3VeD6nUwV8m9flEFwDSwtF65(5kfN1zaiWgSPAZTawSk5lNPLxcTjAL63sn6g0NV0sIXZLfNTwjX4VBajfQAklVeAWfmqkRA2PffLx1KnBKLarIFcqJtuVXoLtcJMPFaCf2UGVy65RWWWuR50ppvrTelSN94D3E393CdCDxZ6EM1Vfx3z)DCcsjs5AfVYYw3PxkGCPwyQW0u7zECItNYEWYHk6UINrLe8IkmrmA50PF6QF7NXPtJx6wVjegTVHjFAXN2Sy9InVgisWvjmC39MhUGO7QNaiKaMLq5yTp6SwSPDFcLzoybCVIOI4Ieu8Z(ql2xX59RmlsGZbWGJ9EvNQ4bdkFj)Wk)s)NqdN4sgrrxIbN)V51HKMLqQEtzNttDa0z23TUtJgT9X6jIgBTBonC2lzJoDVG9on)1ebLPJnKKvdVbwovcNJGdkGEodnVVMQbdTCc(mkR8XnQzv0RxnP)(BTZ2BTtTb)X5)3d
```

The issue with this: go inside an instance.

InfusOnWoW:
Okay, looking at that code and the aura, that rules out ncluding the child auras by default. Breaking something like that is not worth it.

So I think you original change is the right one. I don't like "p" and the description though. I think "parent" zone id is not realy a something that is intuitive. But maybe that's just me.

I would suggest changing it to maybe "r" for recursive or "c" for children? And changing the description to e.g.

Supports multiple entries, separated by commas. To include child zone ids recursively, prefix with r.
Group Zone IDs must be prefixed with 'g', e.g. g277.
[...]

InfusOnWoW:
I did that change now, so imho it's ready now.

jjholleman:
@InfusOnWoW I'm happy with this. 👍

InfusOnWoW (approved):

PR #4872, opened 2024-02-11
harden WeakAuras against bad migrations

emptyrivers:
or rather, re-do the hardening, but in a slightly different way to before If PreAdd errors during AddMany, then that data & all descendants will not hydrate from the database & the user is asked to contact us for further help

Also, fix the data repair tool by taking a snapshot *before* PreAdd is called.

### DIFF
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -2541,34 +2541,57 @@ function Private.AddMany(tbl, takeSnapshots)
 
   local order = loadOrder(tbl, idtable)
   coroutine.yield()
-  local groups = {}
-  for _, data in ipairs(order) do
-    WeakAuras.PreAdd(data)
-    if data.regionType == "dynamicgroup" or data.regionType == "group" then
-      groups[data] = true
+
+  if takeSnapshots then
+    for _, data in ipairs(order) do
+      Private.SetMigrationSnapshot(data.uid, data)
+      coroutine.yield()
     end
-    coroutine.yield()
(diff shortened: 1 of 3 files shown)


WeakAuras/WeakAuras.lua
```diff
@@ -2541,34 +2541,57 @@ function Private.AddMany(tbl, takeSnapshots)
...
+  end
+
   local groups = {}
+  local bads = {}
   for _, data in ipairs(order) do
-    WeakAuras.PreAdd(data)
-    if data.regionType == "dynamicgroup" or data.regionType == "group" then
-      groups[data] = true
+    if data.parent and bads[data.parent] then
+      bads[data.id] = true
```
emptyrivers, line 2556: Honestly, i'm not super sure if this is the best idea, but I can't think of what else to do in this case. Current behavior with Add() is to just not send the bad data to pAdd(), which I think would spill the group's contents onto the screen in an ugly way.

emptyrivers:
side note: the linked issue ran into this problem not from a bad migration, but from corrupt custom options data (& corrupt in a way the current validation doesn't anticipate). I elected not to harden  that validation routine, mostly because I can't imagine how it could have happened without user fuckery (`select` type option without `values`???).

InfusOnWoW (approved):

WeakAuras/WeakAuras.lua
InfusOnWoW: Not sure either, the print output is good though.

PR #4877, opened 2024-02-17
fix buildType detection

emptyrivers:
...by having all of the code run, and then successively commenting out lines that are wrong for the build

honestly, this is how we should have done it anyways, but i didn't fully understand how the bw packager handles it at first

https://github.com/BigWigsMods/packager?tab=readme-ov-file#build-type-keywords

### DIFF
--- a/.github/workflows/build-beta.yml
+++ b/.github/workflows/build-beta.yml
@@ -37,7 +37,7 @@ jobs:
         run: |
           sed -i \
           -e "s/--@experimental@/--\[=====\[@experimental@/" \
-          -e "s/--@end-experimental@/--@end-non-experimental@\]=====\]/" WeakAuras/Init.lua
+          -e "s/--@end-experimental@/--@end-experimental@\]=====\]/" WeakAuras/Init.lua
 
       - name: Create Package
         if: github.event_name == 'push' && contains(github.ref, 'refs/tags/')
--- a/.github/workflows/build-beta.yml
+++ b/.github/workflows/build-beta.yml
@@ -35,10 +35,9 @@ jobs:
 
       - name: Flag Non-Experimental Build
         run: |
-          sed -i -e "s/--@experimental@/--\[=====\[@experimental@/" WeakAuras/Init.lua \
-          -e "s/--@end-experimental@/--@end-experimental\]=====\]/" WeakAuras/Init.lua \
-          -e "s/--\[=====\[@non-experimental@/--@non-experimental@/" WeakAuras/Init.lua \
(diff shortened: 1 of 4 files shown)


WeakAuras/Init.lua
```diff
@@ -368,6 +368,21 @@ local flavorFromTocToNumber = {
 }
 local flavor = flavorFromTocToNumber[flavorFromToc]
 
+
```
emptyrivers, line 371: so the order is deliberate here, WeakAuras.buildType is repeatedly rewritten to progressively less stable flavors until it's correct. Old code tried to have the line order of this reversed, which required a more complex house of cards. I could have fixed that but it would have required thinking about regular expressions and I didn't feel like doing that.

.github/workflows/build-beta.yml
```diff
@@ -35,10 +35,9 @@ jobs:
 
       - name: Flag Non-Experimental Build
         run: |
-          sed -i -e "s/--@experimental@/--\[=====\[@experimental@/" WeakAuras/Init.lua \
-          -e "s/--@end-experimental@/--@end-experimental\]=====\]/" WeakAuras/Init.lua \
-          -e "s/--\[=====\[@non-experimental@/--@non-experimental@/" WeakAuras/Init.lua \
-          -e "s/--@end-non-experimental@\]=====\]/--@end-non-experimental/" WeakAuras/Init.lua
+          sed -i \
+          -e "s/--@experimental@/--\[=====\[@experimental@/" \
+          -e "s/--@end-experimental@/--@end-non-experimental@\]=====\]/" WeakAuras/Init.lua
```
InfusOnWoW, line 40: There's an inconsequential typo in the replacement string, should be --@end-experimental@\]=====\]

.github/workflows/build.yml
```diff
@@ -41,10 +41,9 @@ jobs:
 
       - name: Flag Non-Experimental Build
         run: |
-          sed -i -e "s/--@experimental@/--\[=====\[@experimental@/" WeakAuras/Init.lua \
-          -e "s/--@end-experimental@/--@end-experimental\]=====\]/" WeakAuras/Init.lua \
-          -e "s/--\[=====\[@non-experimental@/--@non-experimental@/" WeakAuras/Init.lua \
-          -e "s/--@end-non-experimental@\]=====\]/--@end-non-experimental/" WeakAuras/Init.lua \
+          sed -i \
+          -e "s/--@experimental@/--\[=====\[@experimental@/" \
+          -e "s/--@end-experimental@/--@end-non-experimental@\]=====\]/" WeakAuras/Init.lua
```
InfusOnWoW, line 46: Same typo here too.

InfusOnWoW (approved):

PR #4893, opened 2024-03-03
Smallpatches

InfusOnWoW:


WeakAurasOptions/RegionOptions/DynamicGroup.lua
```diff
@@ -551,6 +551,20 @@ local function createOptions(id, data)
       bigStep = 0.01,
       isPercent = true
     },
+    sharedFrameLevel = {
+      type = "toggle",
+      width = WeakAuras.normalWidth,
+      name = L["Share Framelevels"],
```
mrbuds, line 557: I feel this option's name is bit too cryptic, i was thinking about "Flatten Frames Levels" but i don't know if its much better

mrbuds (approved):

WeakAurasOptions/RegionOptions/DynamicGroup.lua
Stanzilla: Auto Frame Levels and then in the tooltip what it does?

mrbuds: Without this options auras level are incremental, and with this ticked they all have same level, in both case you can consider it "auto" so that doesn't fit

InfusOnWoW: I agree with you, "Flat" is maybe a better word.

PR #4894, opened 2024-03-03
Split Loaded/Standby section into two separate sections

### LINKED ISSUE #4821: Please split the Loaded/Standby section into two groups
Adding the "Standby" weak auras into the "Loaded" section has made the list a bit overloaded for some users, and challenging to find and update weak auras that are active.  I have dozens of auras in standby currently, that are gated on zones, encounter ids, class/spec settings that are all included in the section.  Before this change I had < 10 loaded auras in the "Loaded" section.

I think it would be nice if there were 3 sections "Loaded", "Standby" and "Not Loaded" sorted in that order.  It would hopefullly provide whatever functionality the surfacing of Standby auras added, while preserving the sort priority of the Loaded auras.

If we could sort all the Loaded auras to the top of the current combined category, it would be a quick improvement.

**Additional context**
image
As an active Raider and M+ player, my standby list is quite large.  I do reguarely clean it up so this is only content relevant to this season, but it is still a huge amount.

lqnrd:
# Description

I really like the idea introduced in January to have "Standby" as a third state of displays. There were quite a lot of displays lost somewhere at the bottom of the list.

One big drawback to having a combined "Loaded/Standby" section however, is that the top most section can get a little cluttered. I myself have a lot of displays for specific dungeon bosses (think mythic+), for raid bosses, for whole dungeons, or even for one particular world map. Most of these will almost never be loaded during a given session, but they are still taking up place at the top of the list. I won't delete them, because they *will* be used once or twice a week, and checking "never load" every time for each of these is just not feasable.

With this commit, we now have three different sections:
- Loaded: everything that's loaded and running right now (like always-on displays, health and mana bars, cast bars)
- Standby: everything that *could* be loaded (like encounter or dungeon specific stuff)
- Not loaded: everything that would never be loaded in the current session

If you really don't want to have this, you are still free to close this pull request, and if there's something I overlooked in the code, please let me know!

Fixes #4821 

## Type of change

## How Has This Been Tested

The changes have been tested on the current version available on curse, which is 5.10.1.

- [x] adding a new display, rename, drag and drop with groups, delete display again

## Checklist


InfusOnWoW:
The linked ticket explains why the top section is the way it is. I want all potentially relevant auras to appear in the top section. That is err on the side of showing too many auras in that section.  

Specifally, the has been quite a bit of user confusion on why auras don't show in the loaded section after importing. Introducing a standby section would defeat that.

Also at this time, I don't want to introduce any wide changes to that display just after users have adjusted to the new categorization.

PR #4899, opened 2024-03-05
only obliterate whispers when we actually find a weakauras link in them

### LINKED ISSUE #4898: Version 5.11.1 seems to block whispers from off server players

- [X] I have searched the existing open and closed issues.

### Description

Starting today with the latest 5.11.1 build. I am unable to get whispers from people on my connected realms but not my actual server. I have another mod that plays a sound on whispers and I would hear the alert, but no messages. I disabled everything til I found it was limited to just Weakauras. I rolled back to version 5.10.1 and problem went away.

### WeakAuras Version

Weakaruas 5.11.1

### World of Warcraft Flavor

Retail (Default)

### World of Warcraft Region

US/NA

- [X] Yes

### Lua Error

_No response_

### Reproduction Steps

Disable all mods, but Weakauras. Have someone from one of the other connected realms whisper me. No whisper. Have someone on my realm whisper me. I get the whispers.

### Last Good Version

Weakaruas 5.10.1

### Screenshots

_No response_

### Export String

_No response_

emptyrivers:
Fixes #4898 (probably, my guinea pig went AFK before i could test this change).

### DIFF
--- a/WeakAuras/Transmission.lua
+++ b/WeakAuras/Transmission.lua
@@ -148,6 +148,7 @@ local function filterFunc(_, event, msg, player, l, cs, t, flag, channelId, ...)
   local newMsg = "";
   local remaining = msg;
   local done;
+  local anyLinkFound = false
   repeat
     local start, finish, characterName, displayName = remaining:find("%[WeakAuras: ([^%s]+) %- (.*)%]");
     if(characterName and displayName) then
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
That looks like the issue should have existed before my change, or maybe I'm too tired to understand it.

WeakAuras/Transmission.lua
```diff
@@ -156,12 +157,13 @@ local function filterFunc(_, event, msg, player, l, cs, t, flag, channelId, ...)
       newMsg = newMsg..remaining:sub(1, start-1);
       newMsg = newMsg.."|Hgarrmission:weakauras|h|cFF8800FF["..characterName.." |r|cFF8800FF- "..displayName.."]|h|r";
       remaining = remaining:sub(finish + 1);
+      anyLinkFound = true
     else
       newMsg = newMsg .. remaining
```
emptyrivers, line 162: Previously, this branch (which should be immediately taken in all cases where we get a whisper that doesn't have a WA link in it) would not append `remaining` onto `newMsg`. Thus, the loop would end with `newMsg` being just the empty string, and we would do nothing to the whisper. Now, it is never the case that `newMsg` is an empty string (unless the message itself is an empty string, i suppose), so we would always enter the block below, even if it's a perfectly innocent whisper.

InfusOnWoW (approved):

PR #4906, opened 2024-03-10
Rework Profiling window

mrbuds:
# Description

List of auras is sortable
Buttons moved on top
Start buttons consolidated into 1 dropdown
UI partially moved to an xml file

https://github.com/WeakAuras/WeakAuras2/assets/189656/2e1a6f99-c59e-48ae-93c5-dca74995017e


Stanzilla:
Can we try using the mono space font for the aura names as well? And use the same darker background for the summary output window

mrbuds:
> Can we try using the mono space font for the aura names as well? And use the same darker background for the summary output window

image

Boneshockx:
Is there a good reason at this point to have report summary do what it does now? I feel like its only used to copy and paste the report. If so 'report summary' might aswell be changed into a 'copy to clipboard' button if that's possible in wow lua.

mrbuds:
Added a systems view

image

> If so 'report summary' might aswell be changed into a 'copy to clipboard' button if that's possible in wow lua.

afaik it's not possible to touch clipboard directly.

mrbuds:
Moved the onLoad stuff to onShow, but wonder if it's not preferable to make it it's own load-on-demand WeakAurasProfiling addon

WeakAuras/Profiling.lua
```diff
@@ -459,250 +404,107 @@ function WeakAuras.PrintProfile()
...
+  spikeTooltip = L["Maximum time used on a single frame"],
+  timeTooltip = L["Cumulated time used during profiling"]
+}
+
+function WeakAurasProfilingLineMixin:Init(e)
+  -- button.pct:SetText(pct)
+	self.progressBar.name:SetText(e.name)
+  self.time:SetText(("%.2fms"):format(e.time))
+	self.spike:SetText(ColoredSpike(e.spike))
+  self.progressBar:SetValue(e.pct)
```
InfusOnWoW, line 417: The indentation here is strange.

WeakAuras/Profiling.lua
```diff
@@ -459,250 +404,107 @@ function WeakAuras.PrintProfile()
...
-    end
-  end)
+  self.ColumnDisplay:LayoutColumns(COLUMN_INFO)
 
+  local view = CreateScrollBoxListLinearView()
+	view:SetElementInitializer("WeakAurasProfilingLineTemplate", function(frame, elementData)
+		frame:Init(elementData)
+	end)
+	ScrollUtil.InitScrollBoxListWithScrollBar(self.ScrollBox, self.ScrollBar, view)
+  ScrollUtil.RegisterAlternateRowBehavior(self.ScrollBox, ApplyAlternateState)
```
InfusOnWoW, line 496: here too the indentation is mixed

WeakAuras/Profiling.lua
```diff
@@ -723,60 +525,273 @@ function RealTimeProfilingWindow:Init()
...
+	end
+
+	local dropDown = self.buttons.modeDropDown
+  UIDropDownMenu_SetWidth(dropDown, 120)
+  UIDropDownMenu_JustifyText(dropDown, "LEFT")
+  UIDropDownMenu_Initialize(dropDown, Initializer)
+end
+
+function WeakAurasProfilingMixin:OnShow()
+  --[[ 3x WeakAuras/Profiling.lua:593: Action[SetPoint] failed because[SetPoint would result in anchor family connection]: attempted from: WeakAurasProfilingFrame:SetPoint.
```
InfusOnWoW, line 639: huh?

WeakAuras/Profiling.lua
```diff
@@ -723,60 +525,273 @@ function RealTimeProfilingWindow:Init()
...
   end
-  if currentProfileState == "encounter" then
-    self.encounterButton:SetText(L["Cancel"])
+  ]]
+end
+
+
+local lastRefresh
+function WeakAurasProfilingMixin:RefreshBars(_, force)
+  if force or (not lastRefresh or lastRefresh < GetTime() - 1) then
```
InfusOnWoW, line 653: This is being called repeately even when no profiling is active.

WeakAuras/Profiling.lua
```diff
@@ -723,60 +525,273 @@ function RealTimeProfilingWindow:Init()
...
+      local elapsed = data[name].elapsed
+      local pct = 100 * elapsed / total
+      local spike = data[name].spike
+      self:UpdateBar(name, elapsed, pct, spike)
+    end
+  end
+
+  self.bars:Sort()
+  if profileData.systems.wa then
+    local timespent = debugprofilestop() - profileData.systems.time.start
```
InfusOnWoW, line 697: If the profiling is stopped using the current time (via debugprofilestop) is wrong. Apart from the onUpdate bug above, this function is also called on switching the profiling mode.

WeakAuras/Profiling.lua
```diff
@@ -459,250 +404,107 @@ function WeakAuras.PrintProfile()
...
-  self.barsFrame = barsFrame
-  barsFrame:SetPoint("TOPLEFT", 7, -20)
-  barsFrame:SetPoint("BOTTOMRIGHT", -3, 30)
-  barsFrame:Show()
-
-  local statsFrameText = self:CreateFontString(nil, "OVERLAY", "GameFontNormal")
-  self.statsFrameText = statsFrameText
-  statsFrameText:SetPoint("BOTTOMLEFT", 15, 25)
+  self.buttons.report:SetText(L["Report Summary"])
+  self.buttons.report.tooltip = L["A detailed overview of your auras and WeakAuras systems\nCopy the whole text to Weakaura's Discord if you need assistance."]
```
InfusOnWoW, line 460: I think this is a bit too hidden now.

WeakAuras/Profiling.lua
```diff
@@ -723,60 +525,273 @@ function RealTimeProfilingWindow:Init()
...
+			local info = UIDropDownMenu_CreateInfo()
+			info.notCheckable = true
+			info.text = entry.text
+			info.func = entry.func
+			UIDropDownMenu_AddButton(info)
+		end
+	end
+
+	local dropDown = self.buttons.startDropDown
+  local dropDownButton = self.buttons.start
```
InfusOnWoW, line 600: Is there any way we can make the 3 widgets in the top line have the same size?

InfusOnWoW:
There are lots of whitespace issues in multiple places, seems like a lot of tabs.

WeakAuras/Profiling.lua
mrbuds: This is still shown in the text summary, it's not more hidden than it used to be

WeakAuras/Profiling.lua
mrbuds: The dropdown in the middle height is weird due to the texture it use, i think this is slightly better

image

mrbuds:
> There are lots of whitespace issues in multiple places, seems like a lot of tabs.

did a pass of linter on both xml & lua files

InfusOnWoW (approved):
The positioning are no longer saved, and there's still commented code in there for that. But apart from that it looks good.

Silvereyes1372:
Hey, I'm not sure if the gradient is relevant for the "system" part. I've seen multiple Discord posts from people with no issues but wondering why the system spikes are red. Perhaps no color or another threshold would be better?

Additionally, maybe removing the color for the spikes in "Time in wa" and "Libgetframe" in the Report summary would be beneficial.

Sorry if I'm overstepping; I did some code (with chatgpt ) for the Report summary only. It's probably Ugly. I'll leave it here in case it makes your life easier somehow.


```lua
local function PrintOneProfile(popup, name, map, total, isSystemEvent) -- add isSystem arg
  if map.count ~= 0 then
    popup:AddText(name .. "  ERROR: count is not zero:" .. " " .. map.count)
  end

  local percent = ""
  if total then
    percent = (", %.2f%%"):format(100 * map.elapsed / total)
  end

  local spikeInfo = ""
  if not isSystemEvent and map.spike then      -- Check for not isSystem
    spikeInfo = ColoredSpike(map.spike)
    elseif map.spike then
    spikeInfo = ("|cff999999(%.2fms)|r"):format(map.spike)     -- grey if isSystemEvent
  end

  popup:AddText(("%s |cff999999%.2fms%s (%s)|r"):format(name, map.elapsed, percent, spikeInfo))
end
```

and in **WeakAuras.PrintProfile()** function theses lines : 

```lua
PrintOneProfile(popup, "|cff9900ffTime inside WA:|r", profileData.systems.wa, nil, true) -- add true arg line 378 for Time in wa
PrintOneProfile(popup, k, systemRegrouped[k], profileData.systems.wa.elapsed, true) -- add true arg line 414 for Systems
PrintOneProfile(popup, id, map,nil,true) -- add true arg line 421 For Lib (didnt test because I have no lib here)

```

mrbuds:
> The positioning are no longer saved, and there's still commented code in there for that. But apart from that it looks good.

fixed

PR #4910, opened 2024-03-15
update issue template

emptyrivers:
Add a new field to issue templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -51,14 +51,25 @@ body:
   validations:
     required: true
 
-- type: checkboxes
+- type: dropdown
   id: testing
   attributes:
     label: Tested with only WeakAuras
-    description: Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?
+    description: |
+      Sometimes, other addons can interfere with WeakAuras. We recommend testing with only WeakAuras enabled to see if the issue persists.
+      You can use the [Bisector Addon](https://www.curseforge.com/wow/addons/bisector) to make this easier. After installing that addon:
+      1. Type in game: `/bisect start +WeakAuras !WeakAurasCompanion`. Your UI should reload.
+      2. Test if the issue is still present:
+        - **YES** Type `/bisect bad`. A window should appear with some text. Copy that text and paste it in the Bisector Report field below.
+        - **NO** Type `/bisect good`. A window should appear with some text. You can either:
+          - Copy that text and paste it in the Bisector Report field below.
+          - Type `/bisect continue` to try to find which addon is causing the issue. Follow the prompts that Bisector prints in your chat.
+      3. Type in game `/bisect reset` to restore your addons to their original state.
     options:
-    - label: "Yes"
-      required: true
+      - I got this issue with only WeakAuras enabled
+      - I have a Bisector Report
+  validations:
+    required: true
 
 - type: textarea
   attributes:
@@ -108,4 +119,11 @@ body:
   validations:
     required: false
 
+- type: textarea
+  attributes:
+    label: Bisector Report
+    description: If you have a Bisector Report, please paste it here.
+    render: Text
+  validations:
+    required: false
 


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -51,14 +51,24 @@ body:
...
+- type: dropdown
   id: testing
   attributes:
     label: Tested with only WeakAuras
-    description: Did you try having WeakAuras as the only enabled addon and everything else (especially something like ElvUI) disabled?
+    description: Did you try having WeakAuras as the only enabled addon? Or, did you run a [Bisector](https://www.curseforge.com/wow/addons/bisector) test to see if the issue is caused by another addon?
     options:
-    - label: "Yes"
-      required: true
+      - I did neither of these 😱
```
emptyrivers, line 60: This is the default because let's be honest, 99% of folks who clicked the box before didn't actually do what they promised they did.

InfusOnWoW: We should then strengthen the request to do so. Maybe add to bisector a mode where only an addon and its depencies are enabled, and then suggest that people use that?

E.g. 
"Some issues are caused by incompatibilities with other addons, if possible, please test whether your issue is caused by another addon. You can either manually disable all other addons or use LINK-TO-BISECTOR and INSTRUCTIONS. After testing RESTORE-INSTRUCTIONS reenables other addons. If another addon causes the issues, bisector can help you identify the problem via GUIDE-LINK"

Stanzilla: maybe some "force solo test" command for bisector that takes an addon name as arg and then disables everything else that the addon does not depend on, restores to default after testing is over. gives an output that we can require people pasting that includes that solo test was ran?

InfusOnWoW: Yes that what I wanted to suggest.

InfusOnWoW: To be a bit more clearer. We don't know how the current "requirement" to check the "tested with only WA" checkbox changes people's behaviour, because we only get data from those that ignored it.

From my feeling, and I might be wrong about that, we had very color picker doesn't work reports compared to how many people asked about it on the discord. So I think it's actually working.

Thus I wouldn't want to change the checkbox to be optional without directing the reporter to try without other addons. Now I think disabling all other addons is not as easy as it could be, so recommending bisector for that would be good. 

I think elvui even has an explicit mode that does disable all other addons, so maybe we should do something similar by bundling bisector. Though addon incompability isn't that common of a problem.

emptyrivers: I just published version 1.1.0 which provides a fast method of testing ${addonName} with all other addons disabled:

```sh
/bisect start
/bisect hint +WeakAuras # signals to Bisector that the issue involved does involve WA
/bisect hint !WeakAurasCompanion # workaround to a thing where companion addon is force-enabled by WA, but isn't a dependency of WA
/bisect continue

# after reloading, WA should be the only loaded addon (aside from Bisector & some 

# if issue persists...
debug tools like bugsack)
/bisect bad # awesome, issue is reproducible without any other addons

# or if issue vanishes...
/bisect good # there's at least one other addon needed, continue with normal algorithm

```

InfusOnWoW:
Bisector could be more user-friendly by printing instructions on the next step on:
* reloadui while a bisect is running, this should probably mention /bisect good/bad/reset as pontial next commands.
* Once bisector detects a minimal set it should mention /bisect reset too.
Imho bisector should mention /bisect reset for pretty much any situation, so that users know how to get out of the situation

Apart from that I'm not sure how much I would want to direct users to use bisector. I suspect normal users will be quite challenged by that command line interface. 
 
Apparently, and I haven't tested that, the elvui instructions are:
* Run "/edebug on"
*  Test. Still have the issue? 
* Run "/edebug off" to reset

The equivalent in bisector is
* Install bisector
/bisect start
/bisect hint +WeakAuras  !WeakAurasCompanion
/bisect continue
...Test
/bisect reset

Obviously bisector can do more, but in the context of our bug reports I'm not sure we need that, that's why I suggested splitting the advice in two parts:

1) How to quickly test whethe disabling all other addons fixes it, and this should really be as frictionless as possible. I want people to do this, and do it quickly. Imho we could also bundle bisector and add a "/wa debug on" command. 

2) How to do a more advanced debugging, this is imho something most users do not need to know, and thus is information that I'd hide in a link.

Stanzilla:
> Bisector could be more user-friendly by printing instructions on the next step on:
> 
> * reloadui while a bisect is running, this should probably mention /bisect good/bad/reset as pontial next commands.
> * Once bisector detects a minimal set it should mention /bisect reset too.
>   Imho bisector should mention /bisect reset for pretty much any situation, so that users know how to get out of the situation
> 
> Apart from that I'm not sure how much I would want to direct users to use bisector. I suspect normal users will be quite challenged by that command line interface.
> 
> Apparently, and I haven't tested that, the elvui instructions are:
> 
> * Run "/edebug on"
> * Test. Still have the issue?
> * Run "/edebug off" to reset
> 
> The equivalent in bisector is
> 
> * Install bisector
>   /bisect start
>   /bisect hint +WeakAuras  !WeakAurasCompanion
>   /bisect continue
>   ...Test
>   /bisect reset
> 
> Obviously bisector can do more, but in the context of our bug reports I'm not sure we need that, that's why I suggested splitting the advice in two parts:
> 
> 1. How to quickly test whethe disabling all other addons fixes it, and this should really be as frictionless as possible. I want people to do this, and do it quickly. Imho we could also bundle bisector and add a "/wa debug on" command.
> 2. How to do a more advanced debugging, this is imho something most users do not need to know, and thus is information that I'd hide in a link.

yeah our instructions are probably too long

emptyrivers:
new in bisector 1.2:

There's a rudimentary API for addons to consume if they wish to whitelabel Bisector's functionality. I didn't spend too long designing it, so I would consider it in an unstable state until/if I get feedback (possibly from myself?) about what's best

`/bisect start` accepts hint strings, as a convenience feature. Additionally, `/bisect start` immediately reloads the UI with the next set of addons to test (if any +hints are given, this will test first only with +hints enabled, similar to the convenience feature I added in 1.1)

After reloading the UI, Bisector prints a welcome message prompting the user on what to do next.

Stanzilla:
> new in bisector 1.2:
> 
> There's a rudimentary API for addons to consume if they wish to whitelabel Bisector's functionality. I didn't spend too long designing it, so I would consider it in an unstable state until/if I get feedback (possibly from myself?) about what's best
> 
> `/bisect start` accepts hint strings, as a convenience feature. Additionally, `/bisect start` immediately reloads the UI with the next set of addons to test (if any +hints are given, this will test first only with +hints enabled, similar to the convenience feature I added in 1.1)
> 
> After reloading the UI, Bisector prints a welcome message prompting the user on what to do next.

So all the user has to run to get started now is `/bisect hint +WeakAuras !WeakAurasCompanion`?

emptyrivers:
> So all the user has to run to get started now is `/bisect hint +WeakAuras !WeakAurasCompanion`?

`/bisect start +WeakAuras !WeakAurasCompanion`, actually.

I also just pushed an update which causes Bisector to immediately print a report after the first test if the user gives a +hint, even if test result was good (if result was bad, proving the issue exists with just WA, they would have gotten a report anyways). The user can continue with the normal algorithm if they wish by `/bisect continue`, or they can copy the report as proof they tested with only WA enabled & use `/bisect reset` to move on with their life.

emptyrivers:
Instructions to use bisector are approximately, 

- `/bisect start +WeakAuras !WeakAurasCompanion`
- Can you reproduce?
  - then `/bisect bad`, copy the report & give it to us.
  - Otherwise, `/bisect good`. Then, you can choose to:
    - copy the report that shows up immediately & give it to us
    - `/bisect continue` & run `/bisect good/bad` until Bisector gives you another report to find the addon(s) interfering with WA.
- `/bisect reset` once you've pasted the report

InfusOnWoW:
Those steps are fine now, though since we are asking the user to install another addon and have a potentially unusual situation with all other addons disabled, we should prepare the user with a bit of motivation, something like this:

"Sometimes problems with WeakAuras are caused by another addon interferring with WeakAuras. You should test that by either manually disabling all other addons, or using the bisector addon. We recommend Bisector because it restores which addons are enabled after diagnosis."

emptyrivers:
how about this?
image

emptyrivers:
i pushed this up to main branch on my fork, you can go to https://github.com/emptyrivers/WeakAuras/issues/new to see a live preview

InfusOnWoW:
yeah looks good to me. Though @Stanzilla usually does those kind of things.

Stanzilla (approved):
period missing after "Your UI should reload" otherwise lgtm!

PR #4914, opened 2024-03-17
Spinbox: Remove mouse wheel support

InfusOnWoW:


Stanzilla:
Because people trigger it by accident?

mrbuds (approved):

InfusOnWoW:
> Because people trigger it by accident?

That, but also that on changing some spinboxes we reload the options at which point the wheel stops on working. So it's unreliable at best.

PR #4917, opened 2024-03-18
add Location trigger

### LINKED ISSUE #4759: Add more load options to built in trigger conditions
Some of the load options, like Zone Name/Zone ID/Mythic+ Affix, are not a built in option for triggers.

Add more of the load options to the built in trigger support. Instance Type/Difficulty/Size Type can already be found under Player/Unit Info -> Conditions. Options like Zone Name/Zone ID/Mythic+ Affix and maybe more could be added here too.

It is possible to achieve this functionality already using custom triggers, but having them added as built in trigger options might be relatively simple but still a nice quality of life improvement.

**Additional context**
Player/Unit Info -> Conditions may not be the right category for these options. The Instance Type/Difficulty/Size Type are found here too, even though these maybe don't totally fit. Perhaps the addition of a new category, named Location or something else, would be appropriate.

Boneshockx:
# Description

This pull request will add a new trigger "~~Player~~ Location" with filters related to zones and instances.
Since the "Conditions" trigger already contained several instance filters, these are now moved to the Player Location trigger.

I've added code to Modernize.lua that removes the instance filters from Conditions and creates a new Player Location trigger.
~~If the leftover Conditions trigger contains no active filters as a result of this, it is removed.~~
Instance filters in Conditions triggers have a "deprecated" string added and auras with these are show an AuraWarning. https://github.com/WeakAuras/WeakAuras2/pull/4917#issuecomment-2013946233

Also fixes #4759 

### Location filters

- Player Location ID(s): The same one as in Load, with added functionality to toggle recursive child maps on/off
- Recursive: toggle only shows if the above input contains an id prefixed with `c` eg. `c1978`
- Zone Name: This leaves open the ability to filter by name instead of IDs
- Subzone Name: minimap zone text
- Map Type
- Indoors
- Instance Name: hidden but stored in state `%instance`
- Instance Size Type
- Instance Difficulty: now also stored in state `%instanceDifficulty`
- Instance Type

WowT_lmktN1wiVU

## Type of change

## How Has This Been Tested

The Modernize has been tested with this aura: https://wago.io/qDhHF9TX-
It contains 2 triggers:
1. Condition trigger with "Alive" and "Instance Size Type" checked
2. Condition trigger with "Instance Difficulty" checked

After importing this aura, two new triggers are added ~~and the 2nd condition trigger is removed~~:
3. Player Location trigger with "Instance Size Type" checked
4. Player Location trigger with "Instance Difficulty" checked


WeakAuras/Modernize.lua
```diff
@@ -2013,6 +2013,69 @@ function Private.Modernize(data)
...
+    -- Remove triggers
+    for i in pairs(delete_triggers) do
+      data.triggers[i] = nil
+    end
+
+    -- Add triggers
+    for _,triggerData in ipairs(new_triggers) do
+      tinsert(data.triggers, triggerData)
+    end
+  end
```
mrbuds, line 2077: This migration ignore conditions, display texts, and trigger activation (i don't fathom how custom could be handled)

Boneshockx: ah I suspected it wouldn't be that simple. I'll look into it.
I was already leaning a bit to just keeping instance filters untouched, so this might be a good reason to just leave it

emptyrivers: Since a new expansion (for both classic & mainline) is on the horizon, another option is to instead add an AuraWarning for auras that use conditions, and then drop support for them on the expansion.

InfusOnWoW:
I don't think you can realistically move these settings to a different trigger, there are too many cases that you canot accurately convert. So the options are, 
1) adding them all in the conditions trigger
2) adding them to a new trigger, and leaving the conditions trigger unchanged

Boneshockx:
I like @emptyrivers idea. 
Don't delete the filters from a Conditions trigger, but instead deprecate them and migrate to a new Player Location trigger.
We could then show an AuraWarning to notify players to update their aura before it breaks:
image

Boneshockx:
The latest commit will not remove the instance filters from the Conditions trigger.
Instead, on migration it will create new Player Location trigger(s) if instance filters are found in the aura.
An AuraWarning is shown that the aura has migrated instance filters to the Player Location trigger.

image
image

mrbuds:
I don't want this merge with this migration

Boneshockx:
I can drop the migration if you'd rather not have it. I thought of it as a convenience before you mentioned texts, conditions and trigger activation. 
Though I do prefer to have the instance filters in the player location trigger, and eventually remove those from the conditions trigger after a while.

WeakAuras/Prototypes.lua
```diff
@@ -10702,6 +10701,183 @@ Private.event_prototypes = {
...
+        tinsert(events, "WA_DELAYED_PLAYER_ENTERING_WORLD")
+      end
+      return events;
+    end,
+    force_events = "INSTANCE_LOCATION_CHECK",
+    name = WeakAuras.newFeatureString..L["Location"],
+    init = function(trigger)
+      local ret = [=[
+        local recursive = %s
+        local zoneChecker = Private.ExecEnv.ParseZoneCheck(%q, recursive)
```
InfusOnWoW, line 10747: Put that into a preamble for zoneIds.

Putting it here makes the trigger parse the zone check string each time it's trigered, whereas the preamble runs the code just once.

For the recursive thing, imho recursive should always be true for the 'c'-prefix.

WeakAuras/GenericTrigger.lua
```diff
@@ -1781,6 +1789,21 @@ function GenericTrigger.Add(data, region)
   else
     Private.AuraWarnings.UpdateWarning(data.uid, "spammy_event_warning")
   end
+
+  if warnAboutDeprecatedInstanceFilters then
+    Private.AuraWarnings.UpdateWarning(data.uid, "deprecated_instance_filter", "warning",
```
InfusOnWoW, line 1794: I don't think we should immediately put an aura warning out for this. From experience those create quite bit of friction

WeakAuras/Prototypes.lua
```diff
@@ -10170,15 +10169,15 @@ Private.event_prototypes = {
       },
       {
         name = "instance_size",
-        display = L["Instance Size Type"],
+        display = L["Instance Size Type"].." "..L["|cffff0000deprecated|r"],
```
InfusOnWoW, line 10172: I think adding a desc that directs to the location trigger here is needed, if we don't add an AuraWarning.

WeakAuras/Prototypes.lua
```diff
@@ -10702,6 +10701,183 @@ Private.event_prototypes = {
...
+        hidden = true,
+        test = "true",
+        noProgressSource = true,
+      },
+      {
+        name = "zone",
+        display = L["Zone Name"],
+        type = "string",
+        conditionType = "string",
+        store = true,
```
InfusOnWoW, line 10798: If you store the value in the state, then the trigger has to always watch the events that update it.

WeakAuras/Prototypes.lua
```diff
@@ -10702,6 +10701,183 @@ Private.event_prototypes = {
...
+        init = "minimapZoneText",
+        multiEntry = {
+          operator = "or",
+        },
+      },
+      {
+        name = "mapType",
+        display = L["Map Type"],
```
InfusOnWoW, line 10818: Do you have an use case in mind? Or did you add it just for completeness? To me it feels a bit too niche.

InfusOnWoW (approved):

Boneshockx:
@InfusOnWoW I think I tackled all your reviews now.
I added MapType initially to check for caves, though I added IsIndoors which is better at checking this and it makes the MapType option redundant, so I removed that.

I've removed the AuraWarning for having instance filters enabled in the conditions trigger. The red deprecated text remains with an additional description to guide users to the location trigger. Since the instance filters are multiselect options I had to make sure it uses arg.desc. This overwrites the multiselect description, but if you prefer to concatenate the two descriptions together then let me know and I will change it.

WeakAuras/Prototypes.lua
```diff
@@ -10702,6 +10705,163 @@ Private.event_prototypes = {
...
   },
+  ["Location"] = {
+    type = "unit",
+    events = {
+      ["events"] = {
+        "ZONE_CHANGED",
+        "ZONE_CHANGED_INDOORS",
+        "ZONE_CHANGED_NEW_AREA",
+        "VEHICLE_UPDATE",
+        "MINIMAP_UPDATE_ZOOM",
```
InfusOnWoW, line 10716: Are you sure about MINIMAP_UPDATE_ZOOM and VEHICLE_UPDATE?

Boneshockx: I am sure about MINIMAP_UPDATE_ZOOM, because ZONE_CHANGED_INDOORS is not sufficient enough to check IsIndoors(). 
I am not sure about VEHICLE_UPDATE, I think I copied this from https://github.com/WeakAuras/WeakAuras2/blob/19bdbb7165fd2812c8e83f727f04c334c694357e/WeakAuras/Prototypes.lua#L1772
I don't know what the reason is for that being there, I have never found this event to be required for anything location related.

InfusOnWoW: Why isn't ZONE_CHANGED_INDOORS not enough? Can you explain?

Boneshockx: https://github.com/WeakAuras/WeakAuras2/assets/6916881/88970e86-0855-448e-9b48-aa20029d9b1a

https://github.com/WeakAuras/WeakAuras2/assets/6916881/e30c20f4-3775-4d43-9234-6daa7393e4ae

`MINIMAP_UPDATE_ZOOM` fires when you move from outdoor to indoor (and your minimap changes)
`ZONE_CHANGED_INDOORS` fires when you are already indoors and you change subzones.
Both events are required.

Boneshockx:
I wasn't able to find anything about `VEHICLE_UPDATE` being required. I'm not sure why it's being used for the load option.   
I've tried pet battling, went to the darkmoon faire rides, did a quest somewhere in stonetalon mountains that forces you in a vehicle.

InfusOnWoW:
The vehicle update comes from [b57179727940d1fb539bccd5f690c9ceb6847f09](https://github.com/WeakAuras/WeakAuras2/commit/b57179727940d1fb539bccd5f690c9ceb6847f09). The bosses there are from the Eternal Palace raid in Bfa. So no idea whether this is still a problem worth considering. I'd say leaving it out is fine. 

I looked at what events are fired on moving inside and I don't see a better event. My conclusion from that is though, that we shouldn't have an inside trigger option. I don't want to build a trigger on an seemingly unrelated event without a good reason. 

(The good reason for using VEHICLE_UPDATE for zone id/name was that it previously worked that way and users had auras relying on us using the event.)

The inside check is not something I have ever seen anyone ask for nor do I see that as crucial. so given that there's no useable event for it, I'd say leave it out.

InfusOnWoW:
I've added a small commit fixing an lua error I encountered while testing and making the parsing function easier to understand (at least I think so).

jjholleman:
@InfusOnWoW I've removed the Indoors option.

Now, something a bit unrelated. While developing this trigger, I wasn't using the Player Location ID filter at first. I had them all split up in their own trigger options. So that made a big list of filters and I had the idea of adding a divider to split the map related filters and the instance filters. Just like in an Aura Trigger where you have categories for Spell Selection Filters, Affected Unit Filters and such. However, it seems like the header type from BuffTrigger is not available in Prototypes. Is that something I can convince you to add? Because it looks nicer than what I have currently.


image
image

InfusOnWoW:
The BuffTrigger 2 options directly use AceOptions, whereas the GenericTrigger doesn't directly use AceOptions, but instead is converted to AceOptions via "ConstructOptions".

ConstructOptions currently doesn't support a "header" type, but adding that is trivial, by just adding code e.g. similar to 
```
elseif (arg.type == "header") then
        options["header_"..name] = {
          type = "header",
          name = arg.display,
          order = order,
          hidden = hidden,
        }
        order = order + 1
```

InfusOnWoW (approved):

PR #4944, opened 2024-03-22
Smallpatches

InfusOnWoW:


WeakAuras/GenericTrigger.lua
```diff
@@ -2946,6 +2946,9 @@ do
   function Private.CheckItemCooldowns()
     for id, _ in pairs(items) do
       local startTime, duration, enabled = GetItemCooldown(id);
+       -- In 10.26. the apis return values changed from 1,0 for enabled to true, false
+       -- We should adjust once its on all versions
+      enabled = enabled and 1 or 0
```
mrbuds, line 2951: this break behavior on not-retail, if enabled was 0 it's now 1

InfusOnWoW: Oh indeed.

InfusOnWoW: Fixed.

WeakAuras/WeakAuras.lua
```diff
@@ -4513,6 +4513,10 @@ local function startStopTimers(id, cloneId, triggernum, state)
       stopAutoHideTimer(id, triggernum, cloneId)
       return
     else
+      if state.expirationTime == nil then
+        -- Set the expiration time, because users rely on that, even though it's wrong to do
+        state.expirationTime = GetTime() + state.duration
+      end
```
mrbuds, line 4519:
```suggestion
      if state.expirationTime == nil and type(state.duration) == "number" then
        -- Set the expiration time, because users rely on that, even though it's wrong to do
        state.expirationTime = GetTime() + state.duration
      end
```
I'm not sure if this test is needed but it's more safe

mrbuds (approved):

PR #4947, opened 2024-03-23
Smallpatches

InfusOnWoW:


WeakAuras/Animations.lua
```diff
@@ -382,7 +382,9 @@ function Private.Animate(namespace, uid, type, anim, region, inverse, onFinished
       last_update = GetTime()
       updatingAnimations = true;
     end
-    RunAnimation(key, animation, 0, GetTime())
+    C_Timer.After(0, function()
+      RunAnimation(key, animation, 0, GetTime())
+    end)
```
Stanzilla, line 387: lovely

InfusOnWoW: Yes indeed... No way that'll create problems in the future.

InfusOnWoW: Actually I have an idea...

InfusOnWoW: Nah, I won't tinkering with it for now.

mrbuds (approved):

PR #4968, opened 2024-03-30
Faction Reputation trigger rework

### LINKED ISSUE #4966: Renown Support Show by Rank (maxed)
Currently made an weakaura as i would normaly for (Old Factions) as Example: When Reached EXALTED
- Weakaura will "Load"

If the player in Dragonflight (now using RENOWN) can go from all between 1 > 30 Rank.
Being as Example (Valdrakken Accord) Renown 30 is maxed, Means that in current state weakauras dont support the new Renown system
It Can show the Rank.

However it cannot show when or IF a player reached MAX.

Request: Renown requirement (< or >=) 1 to XX-number
Request: Renown MAX (True / False)

618e1c3e6521bd341dfbe17edba27881

Currently only support
Vanilla System: > Revered / Exalted
Pandaria System: Friendship Rank

Missing: RENOWN*

Boneshockx:
# Description

Slightly reworks the Faction Reputation trigger with full support for friendship, renown, and paragon reputation.
Fixes #4966

## Details

- All existing trigger fields have been kept as is, no migration is needed.
- Progress value priority is now: Paragon > Renown > Friendship > Basic. So if a reputation includes a paragon level after exalted, the progress values will be updated to track the paragon progress instead once the player starts progressing in the paragon level.
- Trigger filters for Current Reputation, Total Reputation, and Reputation (%) have been made available.
- Added a Capped filter that returns true when a player gathers the maximum amount of reputation.
- Added a At War With filter
- Added a Bonus Reputation Gain filter, used mainly for Mists of Pandaria factions where the player can purchase commendations to unlock bonus rep
- Added  a Is Paragon Reputation filter to check if the faction has a paragon level and the player is currently progressing in paragon.
- Friendship rank and max friendship rank now only show when the selected faction is a friendship.
- Renown level and max renown level filters have been added. Only show when the selected faction is a major faction (renown).
- Added a Is Weekly Renown Capped filter for major factions that have seasonal/weekly caps. This is separated from the "Capped" filter on purpose.
- Available text replacements and conditions: `%name, %value, %total, %percentRep, %standing, %standingId, %capped, %atWar, %isParagon, %friendshipRank, %friendshipMaxRank, %renownLevel, %maxRenownLevel, %isWeeklyRenownCapped`.
- Standing text for friendships and renown now outputs the text as displayed on the Blizzard Reputation Bar.
- Friendship rank and Renown level are available as a progress source.

https://i.imgur.com/Xt9sJHs.png
https://i.imgur.com/yjg2dJP.png
https://i.imgur.com/F1NsbMy.png
https://i.imgur.com/pKKTiwH.png


WeakAuras/Prototypes.lua
```diff
@@ -2467,37 +2480,68 @@ Private.event_prototypes = {
...
         init = "name",
         test = "true"
       },
+      {
+        name = "value",
+        display = L["Reputation"],
+        type = "number",
+        store = true,
+        init = [[currentValue - minValue]],
+        test = "true",
```
mrbuds, line 2501: filter doesn't work if test = "true"

WeakAuras/Prototypes.lua
```diff
@@ -2510,51 +2554,112 @@ Private.event_prototypes = {
...
+        type = "tristate",
+        init = "isCapped",
+        conditionType = "bool",
+        store = true,
+      },
+      {
+        name = "atWarWith",
+        display = L["At War With"],
+        type = "tristate",
+        init = "atWarWith",
```
mrbuds, line 2573: could be called simply "At War"

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,80 @@ Private.event_prototypes = {
...
+        local minValue, maxValue, currentValue
+        local name, description, standingID, barMin, barMax, barValue, atWarWith, canToggleAtWar, isHeader, isCollapsed, hasRep, isWatched, isChild, factionID, hasBonusRepGain, canSetInactive = GetFactionInfoByID(factionID)
+        minValue, maxValue, currentValue = barMin, barMax, barValue
         local standing
-        if tonumber(standingId) then
-           standing = GetText("FACTION_STANDING_LABEL"..standingId, UnitSex("player"))
+        if tonumber(standingID) then
+           standing = GetText("FACTION_STANDING_LABEL"..standingID, UnitSex("player"))
         end
+        local isCapped = standingID == MAX_REPUTATION_REACTION
```
mrbuds, line 2389: MAX_REPUTATION_REACTION is nil on Wotlk and Cata, it's 8 on classic_era

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,86 @@ Private.event_prototypes = {
...
-            elseif C_Reputation.IsMajorFaction(factionID) then
-              local majorFactionData = C_MajorFactions.GetMajorFactionData(factionID)
-              local isCapped = C_MajorFactions.HasMaximumRenown(factionID)
-              bottomValue, topValue = 0, majorFactionData.renownLevelThreshold
-              earnedValue = isCapped and majorFactionData.renownLevelThreshold or majorFactionData.renownReputationEarned or 0
-              standing = RENOWN_LEVEL_LABEL .. majorFactionData.renownLevel
+          local repInfo = factionID and C_GossipInfo.GetFriendshipReputation(factionID);
+          if (repInfo and repInfo.friendshipFactionID > 0) then
+            standing = repInfo.reaction
+            if ( repInfo.nextThreshold ) then
```
InfusOnWoW, line 2400: we tend to use no () for if.

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,86 @@ Private.event_prototypes = {
...
+
+          -- check if this is a faction with a paragon track
+          local isParagon = C_Reputation.IsFactionParagon(factionID)
+          local paragonCurrentValue, paragonBarMax, rewardQuestID, hasRewardPending, tooLowLevelForParagon, realParagonValue
+          if ( isParagon ) then
+            paragonCurrentValue, paragonBarMax, rewardQuestID, hasRewardPending, tooLowLevelForParagon = C_Reputation.GetFactionParagonInfo(factionID)
+            local level = math.floor(paragonCurrentValue/paragonBarMax)-(hasRewardPending and 1 or 0)
+            if level > 0 then
+              realParagonValue = tonumber(string.sub(paragonCurrentValue, string.len( level ) + 1 ))
+           else
```
InfusOnWoW, line 2438: the indetnation is wrong here

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,86 @@ Private.event_prototypes = {
...
+            standing = RENOWN_LEVEL_LABEL .. majorFactionData.renownLevel
+            renownLevel = majorFactionData.renownLevel
+            local renownLevels = C_MajorFactions.GetRenownLevels(factionID)
+            maxRenownLevel = renownLevels and renownLevels[#renownLevels].level
+            isWeeklyRenownCapped = C_MajorFactions.IsWeeklyRenownCapped(factionID)
+          end
+
+          -- check if this is a faction with a paragon track
+          local isParagon = C_Reputation.IsFactionParagon(factionID)
+          local paragonCurrentValue, paragonBarMax, rewardQuestID, hasRewardPending, tooLowLevelForParagon, realParagonValue
```
InfusOnWoW, line 2432: these can be moved inside the isParagon scope

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,86 @@ Private.event_prototypes = {
...
+          if ( isParagon ) then
+            paragonCurrentValue, paragonBarMax, rewardQuestID, hasRewardPending, tooLowLevelForParagon = C_Reputation.GetFactionParagonInfo(factionID)
+            local level = math.floor(paragonCurrentValue/paragonBarMax)-(hasRewardPending and 1 or 0)
+            if level > 0 then
+              realParagonValue = tonumber(string.sub(paragonCurrentValue, string.len( level ) + 1 ))
```
InfusOnWoW, line 2437: I don't understand the logic here, string.len on a number?

WeakAuras/Prototypes.lua
InfusOnWoW: What aws wrong with the old logic?

Boneshockx:
@InfusOnWoW there wasn't anything wrong with the previous logic. Why this part is so different from the old is because I rebuild the trigger from scratch to get a better understanding of the reputation system. Most of the code actually turned out quite similar, except for the paragon block. I think at some point I wanted to show the amount of paragon rewards you received so far, but then scrapped it though the code remained.

I like the old logic more so I reverted to that.

WeakAuras/Prototypes.lua
```diff
@@ -2373,67 +2373,87 @@ Private.event_prototypes = {
...
+          local isParagon = factionID and C_Reputation.IsFactionParagon(factionID)
+          if isParagon then
+            local paragonCurrentValue, paragonBarMax, rewardQuestID, hasRewardPending, tooLowLevelForParagon = C_Reputation.GetFactionParagonInfo(factionID)
+            if paragonCurrentValue then
+              minValue, maxValue = 0, paragonBarMax
+              currentValue = paragonCurrentValue %% paragonBarMax
+              if hasRewardPending then
+                currentValue = currentValue + paragonBarMax
               end
+              isCapped = currentValue >= maxValue
```
InfusOnWoW, line 2440: This is different from before, previously capped was simply HasMaximumRenown()

InfusOnWoW: Ah storing isCapped is new, so that's not acutally new.

Anyway, what made you decide for this behaviour for capped?

Boneshockx: I'm not sure if you mean HasMaximumRenown, I think you mean hasRewardPending? Since you're referring to the paragon block. 
`isCapped = currentValue >= maxValue`  is the same as setting `isCapped = hasRewardPending`, though it makes more sense to use hasRewardPending.

InfusOnWoW: The point is what does "is capped" mean if you have a paragon reputation?

I think there are multiple ways to treat that:
* is capped == exalted || max renown, so ignoring paragon levels for is capped
* is capped == can't earn additional reputation, so false for paragon reputations
* is capped == hasRewardPending

Boneshockx: @InfusOnWoW my idea was is capped == hasRewardPending
But do you propose I split capped and hasRewardPending into seperate options? In that case, capped for a paragon rep would always be true, since you are max rep/renown/friendship.

InfusOnWoW: I frankly have no firm opinion on what is best. Someone that cares deeply about reputations should decide that. As in, you got to decide that.

And yes one option is to use different names

The things you have to consider are:
* How does the api behave. In general we keep close to what the apis view of the world is. As that's the most likely to keep working when blizzard introduces new concepts.
* What does the user expect when they read "is capped".
* What's the most useful, what are the use cases where a user would enable the "is capped" check.

Boneshockx: Paragon is quite a special case :\\
What I've done now:
* Consider `isCapped = true` when player is max exalted/friendship rank/renown level.
* If there is a paragon and player is in paragon: `isCapped = true` following the rule above.
* Paragon is an endless reputation and repeats itself every 10000 points. Previously we would display a player with a paragon reward pending and 74 points as `current: 10074` `total: 10000`. 
* This is now changed to `current: 74` `total: 10000`
* Added a Paragon Reward Pending option.

mrbuds (approved):

InfusOnWoW (approved):

PR #4969, opened 2024-03-30
add a label to bool property changes

emptyrivers:
especially useful if WA is skinned by e.g. elvui, which causes the toggle to almost vanish when switched off

Before:
image

After: 
image

### DIFF
--- a/WeakAurasOptions/ConditionOptions.lua
+++ b/WeakAurasOptions/ConditionOptions.lua
@@ -489,7 +489,6 @@ local function addControlsForChange(args, order, data, conditionVariable, totalA
     args["condition" .. i .. "value" .. j] = {
       type = "toggle",
       width = WeakAuras.normalWidth,
-      name = blueIfNoValue(data, conditions[i].changes[j], "value", L["Differences"]),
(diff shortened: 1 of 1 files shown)


WeakAurasOptions/ConditionOptions.lua
```diff
@@ -489,7 +489,16 @@ local function addControlsForChange(args, order, data, conditionVariable, totalA
     args["condition" .. i .. "value" .. j] = {
       type = "toggle",
       width = WeakAuras.normalWidth,
-      name = blueIfNoValue(data, conditions[i].changes[j], "value", L["Differences"]),
+      name = function()
+        local value = conditions[i].changes[j].value
+        local normalString = ""
```
emptyrivers, line 494: numbery properties (which enter this same branch of condition options) will retain a label of empty string if the blue path isn't taken. I think that's fine, the problem here is specifically with elvui's approach to styling checkbox buttons (black square against a slightly less black background is very hard to see!)

InfusOnWoW: But change.value can be nil for bool properties, which should show as off.

InfusOnWoW (approved):

PR #4970, opened 2024-03-30
Hide boolean types in text replacements tooltips

Boneshockx:
# Description

Trigger boolean types stored in the state are needed to make the option available for conditions, however this also shows them as an available text replacement. When a trigger features many of these boolean options, the text replacements tooltip becomes cluttered with options that are probably never used as output.

Before:
image

After:
image
`%capped, %isParagon, %isWeeklyRenownCapped` are removed from the tooltip, though they are still available.

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -4049,7 +4049,7 @@ function GenericTrigger.GetAdditionalProperties(data, triggernum)
       elseif type(v.enable) == "boolean" then
         enable = v.enable
       end
-      if (enable and v.store and v.name and v.display) then
+      if (enable and v.store and v.name and v.display and v.conditionType ~= "bool") then
         found = true;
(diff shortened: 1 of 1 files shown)


Boneshockx:
And/Or adding a `hideTextReplacement = true` as a trigger arg would give more control to hide things like spellids

WeakAuras/GenericTrigger.lua
```diff
@@ -4049,7 +4049,7 @@ function GenericTrigger.GetAdditionalProperties(data, triggernum)
       elseif type(v.enable) == "boolean" then
         enable = v.enable
       end
-      if (enable and v.store and v.name and v.display) then
+      if (enable and v.store and v.name and v.display and v.conditionType ~= "bool") then
```
InfusOnWoW, line 4052: the "type" is actually the "display-gui widget" type, whereas the "conditionType" is actually closer to what's stored in the state, so using the conditionType is better.

PR #4973, opened 2024-03-31
Smallpatches

InfusOnWoW:


Stanzilla (commented):
need minor bumps

WeakAuras/RegionTypes/RegionPrototype.lua
```diff
@@ -383,8 +383,10 @@ local function UpdateProgressFromState(self, minMaxConfig, state, progressSource
   local remainingProperty = progressSource[8]
 
   if progressType == "number" then
-    local value = state[property] or 0
-    local total = totalProperty and state[totalProperty] or 0
+    local value = state[property]
+    if type(value) ~= "number" then value = 0 end
+    local total = totalProperty and state[totalProperty]
+    if type(total) ~= "number" then total = 0 end
```
Stanzilla, line 389: one of the rare cases where a comment might be good to explain why we are doing that?

WeakAuras/Animations.lua
```diff
@@ -160,6 +158,9 @@ end
 local updatingAnimations;
 local last_update = GetTime();
 local function UpdateAnimations()
+  if not updatingAnimations then
```
mrbuds, line 161: I'd prefer if instead of using `updatingAnimations` the visibility state of `frame` was changed to stop running OnUpdate, but since it needs a later refactor, up to you.

mrbuds (approved):

PR #4975, opened 2024-03-31
Smoll change

mrbuds:


WeakAuras/Types.lua
```diff
@@ -1450,6 +1450,14 @@ elseif WeakAuras.IsCataClassic() then
   Private.power_types[17] = nil
   Private.power_types[18] = nil
   Private.power_types[26] = L["Eclipse"] -- couldn't find a localised global
+
+  ---@class Private
+  ---@field eclipseDirection_types table<string, string>
+  Private.eclipseDirection_types = {
+    none = L["None"],
```
emptyrivers, line 1457: Does this make sense? I see that we only use this option when the trigger power type is 26. Is there ever a time when:
✅ user has powertype (== is a chicken?)
✅ Eclipse direction isn't defined?

Or, am i speaking nonsense because cata was before my time?

mrbuds: when i login the direction is not defined and function return none

mrbuds: it's defined when my power reach -100 or 100, then power doesn't update if i cast wrong spell, but before it's define it can still change in both direction

InfusOnWoW (approved):

emptyrivers (approved):
there might be some strangeness with eclipse direction when tracking units *other* than player, but i think we can wait for a ticket before caring about that (if it's even a problem)

mrbuds:
> there might be some strangeness with eclipse direction when tracking units _other_ than player, but i think we can wait for a ticket before caring about that (if it's even a problem)

option is only enable with trigger.unit == "player"

PR #4982, opened 2024-04-02
LuaLS but it actually works now

emptyrivers:
# Description

After watching LuaLS choke on Types.lua for the 1000th time, I decided to try moving everything in there into a sub-table with a generic type signature, and removing all the redeclarations of `---@class Private`, and then adjusting the rest of the code base appropriately.

It seemed to work!

## Type of change

👏 developer 👏 experience 👏 performance 👏 improvement 👏

## How Has This Been Tested

I followed error messages until my normal savedvariables stopped whining at me. There are probably things I missed.


emmericp:
fwiw: visibility checks being slow is also something that I encountered, I just ended up disabling them in DBM: https://github.com/DeadlyBossMods/LuaLS-Config/blob/main/Check-Config.lua#L81

Stanzilla (approved):

emptyrivers:
@emmericp thanks for the tip! `invisible` definitely was a huge time waster, wow.

PR #4983, opened 2024-04-02
Api Search in text editor

mrbuds:
# Description

Use Blizzard_APIDocumentation to make a side panel to help find documented wow functions (not all of them are in there sadly) or complete as you type.

## Type of change

There is no known bug i want to fix, but i'm not in a hurry to merge, maybe there's some cool stuff to add i'v not think about yet


emptyrivers:
click snippets & search API at the same time, you get a very wide window, may or may not be desirable to ensure exactly one side panel is open at a time 
image

emptyrivers:
consider doing some debouncing to only perform search after the user has stopped typing? and/or spread across frames

emptyrivers:
some events are documented. Maybe add them? IDK how to make that make sense in the UI but you're smart :)
image

mrbuds:
> some events are documented. Maybe add them?



https://github.com/WeakAuras/WeakAuras2/assets/189656/27df1964-824b-4f59-95de-a034593e6463

Stanzilla:
to pile on, can we add teh WA APIs that we offer to use from auras?

mrbuds:
> to pile on, can we add teh WA APIs that we offer to use from auras?

That would ideally be a script to generate documentation data like what's in Blizzard_APIDocumentationGenerated to feed APIDocumentation:AddDocumentationTable

Mhmm maybe .. that would help if you know someone at blizzard willing to share their own script (if they have one)

Stanzilla:
> > to pile on, can we add teh WA APIs that we offer to use from auras?
> 
> That would ideally be a script to generate documentation data like what's in Blizzard_APIDocumentationGenerated to feed APIDocumentation:AddDocumentationTable
> 
> Mhmm maybe .. that would help if you know someone at blizzard willing to share their own script (if they have one)

So kinda like https://github.com/DeadlyBossMods/LuaLS-Config/tree/main but not as intense

mrbuds:
> So kinda like https://github.com/DeadlyBossMods/LuaLS-Config/tree/main but not as intense

I don't see how that help generate files like https://github.com/Gethe/wow-ui-source/blob/ddfbf3288df2e897818d8a8cdf76f1309bb20481/Interface/AddOns/Blizzard_APIDocumentationGenerated/AdventureJournalDocumentation.lua

mrbuds:
More toughs on this project

In this state it's trying to do too much at once, and i'm not pleased with everything

## intelisense widget & LibAPIAutoComplete-1.0 
I have strong doubts that the *intelisense* widget from LibAPIAutoComplete-1.0 is good enough

It happens too often that it's shown in bad context
It's big enough that often you do unwanted clicks, that mistakenly replace a word with a function, and editbox doesn't have an undo / ctrl-z function. This can make the widget more an annoyance than an improvement.
It's data is incomplete, which feels less problematic with the side search panel, but when navigating your code it's too obvious

Also LibAPIAutoComplete-1.0 as a lib isn't working, as it use a frame template defined in an xml file, it's necessary for WowScrollBoxList with DataProvider, but that doesn't fit in a lib as it generate a name conflicts when xml file load in a 2nd addon using same lib (i know river's github action for edit namespace in xml files could solve this, but it doesn't feel right)

**I think it's best to drop that idea**, for now at least

## WeakaurasAPI (weakauras function added to apidoc)
It could be good to move that to a separate loadondemand addon?

Next, i'm not 100% pleased with the types generated, luals --doc resolve all aliases into their basic type

for example with `---@alias UnitToken "player" | " target" | "focus"` and a `---@param UnitToken unit`, in generated json file it won't show `UnitToken` but `"player" | " target" | "focus"` :(


**It's for the best to hold this** until 11.0 WeakAuras namespace cleanup, and take time to find a solution for this type problem

## Search API side panel
I'd need to decouple it from LibAPIAutoComplete-1.0
Maybe make it hide other panels when shown instead of display multiple panels
I also want change background like with <https://github.com/WeakAuras/WeakAuras2/pull/5102>


That would reduce the scope of this PR to just the side panel

InfusOnWoW:
@mrbuds  I hacked on the library for a bit and got keyboard handling to mostly work. The change is quite hacky, as I had to experiment a lot to get it to mostly work and I didn't even try to understand how to cleanly pass various widgets, data to various methods that need the data.


I changed:
* While the popup is open: Up/Down/Enter/Escape are handled by the popup and do the right thing
* Tweaked in which circumstances the popup shows up, so that it only shows up if the left side looks like a word and the right side is a space.

To me that feels pretty useable imho, and to me it looks like this is worth pursuing, even with the noted drawbacks you mentioned.

mrbuds:
PR split
panel https://github.com/WeakAuras/WeakAuras2/pull/5170
autocomplete widget https://github.com/WeakAuras/WeakAuras2/pull/5201
WeakAurasAPI.lua & wiki https://github.com/WeakAuras/WeakAuras2/pull/5216

PR #4992, opened 2024-04-04
fix some more type annotations

emptyrivers:


WeakAuras/Types.lua
```diff
@@ -1495,24 +1504,23 @@ for id, str in pairs(Private.combatlog_spell_school_types) do
...
   Private.GetCurrencyIDFromLink = C_CurrencyInfo.GetCurrencyIDFromLink
   Private.ExpandCurrencyList = C_CurrencyInfo.ExpandCurrencyList
   Private.GetCurrencyListInfo = C_CurrencyInfo.GetCurrencyListInfo
 elseif WeakAuras.IsWrathOrCata() then
   Private.GetCurrencyListSize = GetCurrencyListSize
+  ---@type fun(currencyLink: string): number?
   Private.GetCurrencyIDFromLink = function(currencyLink)
     local currencyID = string.match(currencyLink, "|Hcurrency:(%d+):")
-    return currencyID
+    return tonumber(currencyID)
```
emptyrivers, line 1517: @Boneshockx can you verify this is correct? It looks like it's correct (C_CurrencyInfo.GetCurrencyIDFromLink returns number, not string)? You touched this code last, so you own it :).

WeakAuras/Types.lua
```diff
@@ -266,12 +266,21 @@ local simpleFormatters = {
...
+---@alias optionGetter fun(name: string): any
+---@alias formatter fun(input: string): string, nil | boolean
+
+---@class FormatType
+---@field display string
+---@field AddOptions fun(symbol: string, hidden: boolean, addOption: optionAdder, get: optionGetter)
+---@field CreateFormatter fun(symbol: string, get: optionGetter, withoutColor: boolean, data: table): formatter | nil
+
+---@type table<string, FormatType>
 Private.format_types = {
```
emptyrivers, line 279: If you checkout this branch, you might notice 2 diagnostics for redundant-return-value, related to this type annotation. They're on these lines, which I can't make sense of: https://github.com/WeakAuras/WeakAuras2/blob/b314b97d439c5b80f7926af4d9c3da76fb565688/WeakAuras/Types.lua#L492-L494

InfusOnWoW (approved):

WeakAuras/Types.lua
InfusOnWoW: Yeah tonumber should be correct here.

Stanzilla (approved):

PR #4997, opened 2024-04-07
Smallpatches

InfusOnWoW:


WeakAuras/Modernize.lua
```diff
@@ -2022,6 +2022,16 @@ function Private.Modernize(data)
     end
   end
 
+  if data.internalVersion < 73 then
```
mrbuds, line 2025: internalVersion is still set to 72

mrbuds: or is it intended?

InfusOnWoW: no, unintended

mrbuds (approved):

PR #5008, opened 2024-04-09
Cataclysm

mrbuds:


InfusOnWoW (approved):
That code is a 1:1 copy?

mrbuds:
> That code is a 1:1 copy?

Yes it is same code as in retail without any change

PR #5013, opened 2024-04-12
Types

mrbuds:


WeakAuras/WeakAuras.lua
```diff
@@ -1607,14 +1629,18 @@ local function GetInstanceTypeAndSize()
   return "none", "none", nil, nil, nil
 end
 
+---@return string instanceType
+---@return nil
```
InfusOnWoW, line 1633: I guess the cleaner implementation of InstanceType would be:
``` 
return (GetInstanceTypeAndSize())
```

InfusOnWoW (approved):

WeakAuras/WeakAuras.lua
mrbuds: done

PR #5014, opened 2024-04-14
Prepare for the removal of deprecated functions

mrbuds:
10.2.7 PTR added a CVar to disable Blizzard_Deprecated* addons

`/run C_CVar.SetCVar("loadDeprecationFallbacks", 0)`

The goal of this PR is to make WeakAuras work for all versions of the game, including incoming The War Within (TWW) alpha

This was tested (not extensively) on all versions
It work fine on Dragonflight, Cataclysm beta and Classic SoD
It doesn't work on Classic WoTLK where C_AddOns and C_Item are missing too much functions, Classic WoTLK is gone starting 30th april (cataclysm prepatch release)

LibRangeCheck-3.0 will also need an update

C_Container.GetItemCooldown => C_Item.GetItemCooldown also exists, but was triggering error on classic sod


InfusOnWoW:
I looked through the current iteration and everything looked good.

InfusOnWoW:
I think we can merge this now, can't we?

InfusOnWoW (approved):

PR #5027, opened 2024-04-23
Add headers to categorize filters in large triggers

Boneshockx:
This adds headers to triggers with a large amount of filters, similarly to the headers in BuffTrigger2.

Headers were only added to triggers where filters could be grouped distinctively, and with at least 2 or more filters to group together. In some cases, filters were shifted up or down in the list to order and group together in a better way.

Triggers with newly added headers:
* Combat Log
* Cast
* Character Stats
* Faction Reputation
* Health
* Location
* Player Experience
* Power
* Threat Situation

Examples:
image
image
image


mrbuds:
That nice!

For Combat Log trigger `Prefix` & `Suffix` are wrong and will make confusion with `Message Prefix` & `Message Suffix` options (btw I don't know why they are called `Message` this seems wrong)

I'm not sure about the word `Filters` on most headers, i'd try shorten without that word

Boneshockx:
@mrbuds `Message Prefix` & `Message Suffix` wording is indeed odd. Api and Wiki refers to it as `subevent`. If we want to match that I'd say rename `Message Prefix` > `Subevent` and rename `Message Suffix` > `Subevent suffix`. This would make more sense for the special subevents like `damage_shield_missed` which are in the prefix list but don't have a suffix.  
  
For the headers in that case; I'd merge prefix and suffix headers and call it `Subevent info`

I agree with the removal of the `Filters` word.

InfusOnWoW:
I agree with Buds, this looks pretty nice. I don't have an opinion on whether the inclusion of "Filters" is better or not. You guys can decide that. Otherwise it looks imho ready.

InfusOnWoW (approved):

Stanzilla:
image

Boneshockx:
I removed "Filters" from the header title. In the CLEU trigger that looked a bit meh as it would be just "Source". So I have appended "Info" where the header would otherwise be short, and I think it has a nicer ring to it than "Filters". (This is a bit nitpicky, but I like it more)

Second, I merged prefix and suffix headers into "Subevent Info".

PR #5029, opened 2024-04-25
WeakAuras.WatchUnitChange & UnitIsUnit

mrbuds:


WeakAuras/GenericTrigger.lua
```diff
@@ -3392,8 +3393,57 @@ function WeakAuras.WatchUnitChange(unit)
     watchUnitChange:RegisterEvent("UNIT_PET")
     watchUnitChange:RegisterEvent("RAID_TARGET_UPDATE")
 
+    local function testMultiUnitIsUnit(unitA, unitB)
```
InfusOnWoW, line 3396: This does a lot of checks, quite a lot more than I would want to add. Fortunately I think there's an easier way to handle that if we have two datastructures:

unitIdToGUID map (which we alreaedy have in watchUnitChange.unitChangeGUIDS
and GUIDToUnitIds map

And we rely on UnitGUID returning the same value if two unit ids refer to the same unit.

Then e.g. the PLAYER_TARGET_CHANGED code becomes:
1)
local eventsToSend = {}

local oldguid = unitIdToGUID["target"]
if oldguid then
  for unitId in pairs(GUIDToUnitIds[oldguid]) do
    if unitId ~= "target" then
      -- Check for multi units here? Unsure
     eventsToSend["UNIT_IS_UNIT_CHANGED_target_" .. unitId] = target
     eventsToSend["UNIT_IS_UNIT_CHANGED_" .. unitID .. "_target"] = unitId
   end
  end
end

local newguid = UnitGUID("target")
if newguid then
  if GUIDToUnitIds["target"] then
    for unitId in pairs(GUIDToUnitIds[newguid]) do
      -- Check for multi units here? Unsure
      eventsToSend["UNIT_IS_UNIT_CHANGED_target_" .. unitId] = target
      eventsToSend["UNIT_IS_UNIT_CHANGED_" .. unitId .. "_target"] = unitId
    end
  end
end

2) Update data structures
Update watchUnitChange.unitChangeGUIDS
GUIDToUnitIds[oldGUID]["target"] = nil
GUIDToUnitIds[newGUID]["target"] = true

3) Send events
for event, unit in pairs(eventsToSend) do
  WeakAuras.ScanEvents(event, unit)
end


Btw, the code in watchUnitChange OnEvent handler looks a bit strange.
The events:
PLAYER_SOFT_ENEMY_CHANGED, PLAYER_SOFT_FRIEND_CHANGED, INSTANCE_ENCOUNTER_ENGAGE_UNIT, GROUP_ROSTER_UPDATE all end up in the last block, which looks wrong to me.

InfusOnWoW (changes requested):

WeakAuras/GenericTrigger.lua
mrbuds: I refactored the whole function

- UnitIsUnit between nameplates & target still work
- Markers on nameplates work
- i don't remember why UNIT_TARGET is register and what to do with it

More tests needs to be done to be sure everything is ok

WeakAuras/GenericTrigger.lua
```diff
@@ -3391,85 +3393,164 @@ function WeakAuras.WatchUnitChange(unit)
...
+          markerUpdate(unit, eventsToSend)
+          reactionUpdate(unit, eventsToSend)
         end
-        if event == "NAME_PLATE_UNIT_ADDED" then
-          watchUnitChange.nameplateFaction[unit] = WeakAuras.GetPlayerReaction(unit)
+      elseif event == "NAME_PLATE_UNIT_ADDED" then
+        unitUpdate(unit, eventsToSend)
+        markerUpdate(unit, eventsToSend)
+        reactionUpdate(unit, eventsToSend)
+      elseif event == "NAME_PLATE_UNIT_REMOVED" then
```
InfusOnWoW, line 3478: TODO: Might be wrong...

WeakAuras/GenericTrigger.lua
```diff
@@ -3365,6 +3365,7 @@ function WeakAuras.WatchUnitChange(unit)
   if not watchUnitChange then
     ---@class UnitChangeFrame: FrameScriptObject
     watchUnitChange = CreateFrame("Frame");
+    watchUnitChange.trackedUnits = {}
```
InfusOnWoW, line 3368: I've added this to quickly check whether we are even interested in this specific unit. This avoids e.g. tracking all nameplates if there's no unit characteristics trigger for nameplates.

WeakAuras/GenericTrigger.lua
```diff
@@ -3434,6 +3435,14 @@ function WeakAuras.WatchUnitChange(unit)
       end
     end
 
+    local function markerInit(unit)
+      watchUnitChange.raidmark[unit] = GetRaidTargetIndex(unit) or 0
+    end
+
+    local function markerClear(unit)
+      watchUnitChange.raidmark[unit] = nil
+    end
```
InfusOnWoW, line 3444: For almost every event, e.g. PLAYER_TARGET_CHANGED, we don't need to actually check whether the mark has changed, because we if the guid is different we'll send a UNIT_CHANGED anyway, and if it isn't different (e.g. a bogus event) then presumably the mark hasn't either.

WeakAuras/GenericTrigger.lua
```diff
@@ -3472,66 +3492,95 @@ function WeakAuras.WatchUnitChange(unit)
...
-        reactionUpdate("softfriend", eventsToSend)
+        markerInit("softfriend")
+        reactionInit("softfriend")
       elseif event == "RAID_TARGET_UPDATE" then
         for unit in pairs(watchUnitChange.raidmark) do
           markerUpdate(unit, eventsToSend)
         end
       elseif event == "UNIT_FACTION" then
-        if unit:sub(1, 9) == "nameplate" then
+        if watchUnitChange.trackedUnits[unit] then
```
InfusOnWoW, line 3550: When the reaction was initially introduced it was only supposed to be only available on the nameplate unit, though it turns out that the option was not disabled/hidden for other unit types in the Unit Characteristics trigger. 

Thus, fix that by removing this check, and making the reaction option available in all triggers.

WeakAuras/GenericTrigger.lua
```diff
@@ -3472,66 +3492,95 @@ function WeakAuras.WatchUnitChange(unit)
...
           eventsToSend["UNIT_CHANGED_" .. pet] = pet
         end
       elseif event == "PLAYER_ROLES_ASSIGNED" then
+        if not watchUnitChange.trackedUnits[unit] then
+          return
+        end
         roleUpdate(unit, eventsToSend)
       elseif event == "UNIT_TARGET" then
-        -- i dont know why this event is register
+        local unitTarget = unit .. "target"
```
InfusOnWoW, line 3564: This allows e.g. checking whether boss1target == player.

WeakAuras/GenericTrigger.lua
```diff
@@ -3472,66 +3492,95 @@ function WeakAuras.WatchUnitChange(unit)
...
+        if not watchUnitChange.trackedUnits[unitTarget] then
+          return
+        end
+        unitUpdate(unitTarget, eventsToSend)
+        markerInit(unitTarget)
+        reactionInit(unitTarget)
       elseif event == "GROUP_ROSTER_UPDATE" then
-        for unit in pairs(watchUnitChange.unitIdToGUID) do
-          if Private.multiUnitUnits.raid[unit] or Private.multiUnitUnits.group[unit] then
+        for unit in pairs(Private.multiUnitUnits.group) do
```
InfusOnWoW, line 3572: We can restrict ourselves to checking just Private.multiUnitUnits.group here (which is already a lot, but that can't be helped)

WeakAuras/Prototypes.lua
```diff
@@ -2195,8 +2200,6 @@ Private.event_prototypes = {
           end
         end,
         test = "unitisunit",
-        store = true,
-        conditionType = "bool",
```
InfusOnWoW, line 2192: This condition did nothing and was particular useless. I looked at whether it's fixable, since it does sound useful, but unfortunately that appears to be not easy.

WeakAuras/GenericTrigger.lua
```diff
@@ -3472,66 +3492,95 @@ function WeakAuras.WatchUnitChange(unit)
...
           reactionUpdate(unit, eventsToSend)
         end
       elseif event == "UNIT_PET" then
         local pet = WeakAuras.unitToPetUnit[unit]
-        if pet then
+        if pet and watchUnitChange.trackedUnits[pet] then
           eventsToSend["UNIT_CHANGED_" .. pet] = pet
         end
       elseif event == "PLAYER_ROLES_ASSIGNED" then
+        if not watchUnitChange.trackedUnits[unit] then
```
mrbuds, line 3559: PLAYER_ROLES_ASSIGNED has no payload, unit is nil

WeakAuras/GenericTrigger.lua
```diff
@@ -3547,7 +3596,13 @@ function WeakAuras.WatchUnitChange(unit)
       Private.StopProfileSystem("generictrigger unit change");
     end)
   end
-  watchUnitChange.unitIdToGUID[unit] = UnitGUID(unit) or ""
+  local guid = UnitGUID(unit) or ""
```
mrbuds, line 3599: i think the `or ""` can be remove thanks to `trackedUnits`

mrbuds:
WeakAuras.WatchUnitChange could exit sooner if watchUnitChange.trackedUnits[unit] ~= nil (as this function is called more often now)

and `IsWrath` needs to be go for rebase

InfusOnWoW (approved):

PR #5030, opened 2024-04-26
Cataclysm tiny change

mrbuds:
read commit for pepeHands


WeakAuras/Types.lua
```diff
@@ -1754,26 +1754,25 @@ WeakAuras.spec_types_specific = {}
 ---@type table<number, string>
 Private.spec_types_all = {}
 local function update_specs()
+  local cataFix = WeakAuras.IsCataClassic() and -1 or 0
```
InfusOnWoW, line 1757: We should report that as bug, and add a comment here what the bug number is.

InfusOnWoW (approved):

WeakAuras/Types.lua
mrbuds: Bug report is https://github.com/Stanzilla/WoWUIBugs/issues/559

InfusOnWoW (approved):

PR #5037, opened 2024-04-29
advertise a link to an experimental build, if one exists

emptyrivers:
try opening a pull request (any kind!) over on my [fork](https://github.com/emptyrivers/WeakAuras2/pulls), or see emptyrivers/WeakAuras2#12 for an example.

Will only post links for pull requests with the feature preview label.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,8 @@
 name: CI-PR
 
-on: [pull_request]
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
 
 jobs:
   build:
@@ -32,9 +34,22 @@ jobs:
           args: -d -z
 
       - uses: actions/upload-artifact@v4
+        id: upload-artifact
         with:
           name: WeakAuras-PR#${{ github.event.number }}-${{ github.sha }}
           path: .release/
+
+      - uses: exercism/pr-commenter-action@v1.5.1
+        if: ${{ contains(github.event.pull_request.labels.*.name, '🆕 Feature Preview') }}
+        with:
+          github-token: ${{ github.token }}
+          config-file: ".github/pr-commenter.yml"
+          template-variables: |
+            {
+              "repository": "${{ github.repository }}",
+              "build_id": "${{ steps.upload-artifact.outputs.artifact-id }}",
+              "run_id": "${{ github.run_id }}"
+            }
 
       - name: Send Status to Discord
         uses: nebularg/actions-discord-webhook@v1
--- a/.github/pr-commenter.yml
+++ b/.github/pr-commenter.yml
@@ -1,0 +1,12 @@
+comment:
+  on-update: recreate
+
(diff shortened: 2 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -1,6 +1,8 @@
 name: CI-PR
 
-on: [pull_request]
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled]
```
emptyrivers, line 5: slight weirdness here: a build is retriggered if someone edits the labels. I wanted to keep this to just 'feature preview' prs, so there's some discernment about which changes we actually expect someone to install & give feedback on. But we could also ditch that if frivolous retriggers are considered annoying

.github/workflows/pull_request.yml
```diff
@@ -32,10 +34,26 @@ jobs:
           args: -d -z
 
       - uses: actions/upload-artifact@v4
+        id: upload-artifact
         with:
           name: WeakAuras-PR#${{ github.event.number }}-${{ github.sha }}
           path: .release/
 
+      - name: test
```
Stanzilla, line 42: what does this do exactly?

emptyrivers: anything you want, so long as that isn't very useful

Stanzilla: -
ConfusedConfusedFaceGIF

Jodsderechte: It obviously tests duh

Boneshockx:
What if this posted to #wow-addon and pings @WAalpha

.github/workflows/pull_request.yml
```diff
@@ -32,10 +34,23 @@ jobs:
           args: -d -z
 
       - uses: actions/upload-artifact@v4
+        id: upload-artifact
         with:
           name: WeakAuras-PR#${{ github.event.number }}-${{ github.sha }}
           path: .release/
 
+      - uses: exercism/pr-commenter-action@v1.5.1
+        if: ${{ contains(github.event.pull_request.labels.*.name, '🆕 Feature Preview') }}
```
InfusOnWoW, line 43: relying on 🆕 being part of the string sounds like a time bomb

emptyrivers: Hmm, how do you reckon? It's just a unicode codepoint (that either I or @Stanzilla wrote into the label's name a million years ago). We could drop the emojis from the name, certainly, but (because the expression syntax is fairly limited) trying to preserve them & also not write emojis into our source files would involve IMO worse hackery:

```yml
      - uses: exercism/pr-commenter-action@v1.5.1
      	if: ${{ contains(join(github.event.pull_request.labels.*.name, ', '), 'Feature Preveiew') }}
```

InfusOnWoW: Ah okay, then don't bother.

InfusOnWoW (approved):

PR #5038, opened 2024-04-30
Cata

mrbuds:


InfusOnWoW:
I don't like that. Tooltips are used quite a lot for various purposes, and I don't think we can break that for adopting the new api.

PR #5052, opened 2024-05-03
only upload & advertise build artifact if triggering actor has a write bit

emptyrivers:


.github/workflows/pull_request.yml
```diff
@@ -1,14 +1,20 @@
 name: CI-PR
 
 on:
-  pull_request:
+  pull_request_target:
```
emptyrivers, line 4: pull_request_target creates a trusted environment for the workflow run, so some care is needed with outsider submissions (namely, the pull request changes).

we can check if the build succeeds just fine, since those steps don't actually execute the code. But just in case some joker tries to make a version of WeakAuras that steals all of your gold, lets not upload the artifact from such a build unless someone with write access triggered the workflow run.

PR #5055, opened 2024-05-05
Smallfixes

### LINKED ISSUE #5012: Fade out slider for sounds
Hi,

I was recently making a WeakAura where I needed to use custom code to play sounds. I noticed that the StopSound API has an option to fade out the sound effect (https://warcraft.wiki.gg/wiki/API_StopSound). 

Please consider adding a fade out time slider or something similar to the WeakAuras sound GUI for conditions and actions, I think a lot of people would enjoy making use of it.

Many thanks!

### LINKED ISSUE #5042: Mouse Clicks are incorrectly enabled when using "Tooltip on Mouseover"

- [X] I have searched the existing open and closed issues.

### Description

Normally auras don't register mouse clicks but when you enable "Tooltip on Mouseover" in the Display tab of an icon aura, it starts registering mouse clicks. For example I use my mouse to move my camera around, if my mouse happens to land on an aura with tooltip enabled then it eats the mouse click and prevents me from moving the camera.

Putting `aura_env.region.tooltipFrame:SetMouseClickEnabled(false)` somewhere in my aura solves this issue while retaining a working tooltip. I disabled all addons & removed all of my auras in case an aura of mine was screwing something up, but it still happens so can be assumed to be an issue with the base addon.

### WeakAuras Version

5.12.9

### World of Warcraft Flavor

Classic

### World of Warcraft Region

EU

### Lua Error

_No response_

### Reproduction Steps

1. Create an Icon Aura
2. Enable "Tooltip on Mouseover"
3. Aura eats your mouse clicks and prevents you from moving your camera

### Last Good Version

_No response_

### Screenshots

_No response_

### Export String

_No response_

### Bisector Report

_No response_

### LINKED ISSUE #5047: Instance Type load condition is empty on Cata Classic

- [X] I have searched the existing open and closed issues.

### Description

load

I selected WoW flavor below as Wrath Classic but this is for Cata Classic, issue template is outdated.

### WeakAuras Version

5.12.9

### World of Warcraft Flavor

Wrath of the Lich King Classic

### World of Warcraft Region

EU

### Lua Error

```Text
No bugsack error
```

### Reproduction Steps

1. Create an aura
2. Go into Load tab
3. Select "Instance Type"
4. Nothing there

### Last Good Version

_No response_

### Screenshots

_No response_

### Export String

_No response_

### Bisector Report

_No response_

Boneshockx:
- In the CLEU trigger, "Message Prefix" and "Message Suffix" have been renamed to "Subevent" and "Subevent Suffix".  _Previously some special events like `DAMAGE_SHIELD_MISSED` (and other events without a suffix) would be found in the Message Prefix list, though this event does not have a prefix or suffix; it's simply the "Subevent"._ 
- Enable clickthrough when tooltips on mouseover is enabled. Fixes #5042
- Fill instance type options on Cata Classic. Fixes #5047
- Add sound fadeout options. Fixes #5012 

image


WeakAuras/RegionTypes/RegionPrototype.lua
```diff
@@ -203,7 +203,8 @@ local function SoundPlay(self, options)
     return
   end
 
-  self:SoundStop();
+  local fadeoutTime = options.sound_fade and options.sound_fade * 1000 or 0
+  self:SoundStop(fadeoutTime);
```
InfusOnWoW, line 207: This should imho not use the fadeoutTime at all, but immediately stop.

Boneshockx: In what situation do you want the sound to stop immediately even if the user has set a sound fade in conditions?
Or are you saying there shouldn't be a sound fadeout in conditions?

InfusOnWoW: That isn't the code path from a SoundStop via conditions, this is when the user uses "Sound/Play" and we automatically stop the sound, in that case the sound should stop immediately and not be affected by an unrelated setting.

InfusOnWoW: Or actually you might be right that it is the code path called from options. Still the code here is wrong to assume that the settings apply. So there's a check for options.sound_type == "Stop" missing here.

Boneshockx: I added the check for sound_type == "Stop" in 76896a6 and changed the code a bit so that `soundFadeout` is set to `self` / `region` instead of a local variable.

WeakAuras/RegionTypes/RegionPrototype.lua
```diff
@@ -203,8 +203,8 @@ local function SoundPlay(self, options)
     return
   end
 
-  local fadeoutTime = options.sound_fade and options.sound_fade * 1000 or 0
-  self:SoundStop(fadeoutTime);
+  self.soundFadeout = options.sound_type == "Stop" and options.sound_fade and options.sound_fade * 1000 or 0
```
InfusOnWoW, line 206: Imho it was fine as a parameter to SoundStop()

Boneshockx: I thought it was nice if someone required it for custom aura things.
Though I missed that there already is `self.soundOptions = options`, so this change was indeed not necessary.

Boneshockx:
Oh buds did an oopsie :)

InfusOnWoW (approved):

PR #5056, opened 2024-05-05
AuraBar: add some alias textures, and option to use texture picker

mrbuds:
## Screenshots
image
image

classic sod and cataclysm:
image

dragonflight:
image

## What could be done differently

An other possibility is to not add these aliases to LSM dropdown, and in the Texture Picker widget make a curated list of usable textures, but i don't like to add a list to maintain manually, and i don't believe average user explore the texture picker much (as i don't see much users take advantage of the model subregion)

## Notes

- WeakAurasStatusbarAtlasWidget.lua is almost a 1:1 copy of LSM30_Statusbar widget, only difference is the Atlas test for using SetAtlas in 2 places

- Funkeh raised concern on discord about mixing 2 types in the values, but we do that already with Private.SetTextureOrAtlas

- The texture picker option make auraBar intersect some functionalities with progressTexture, bonus is i think users naturally go to auraBar when they want a progress bar, it support spark and ticks

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -3151,6 +3151,30 @@ LSM:Register("statusbar", "Thick Stripes", [[Interface\AddOns\WeakAuras\Media\Te
 LSM:Register("statusbar", "Thin Stripes", [[Interface\AddOns\WeakAuras\Media\Textures\Statusbar_Stripes_Thin]])
 LSM:Register("border", "Drop Shadow", [[Interface\AddOns\WeakAuras\Media\Textures\Border_DropShadow]])
 
+if PowerBarColor then
+  local function capitalizeFirstLetter(str)
+    -- Split the string into words separated by underscores
+    local words = {}
+    for word in string.gmatch(str, "[^_]+") do
+      table.insert(words, word)
+    end
+    -- Capitalize the first letter of each word
+    for i, word in ipairs(words) do
+      words[i] = word:sub(1, 1):upper() .. word:sub(2):lower()
+    end
+    return table.concat(words, " ")
+  end
+
+  for power, data in pairs(PowerBarColor) do
+    if type(power) == "string" and data.atlas then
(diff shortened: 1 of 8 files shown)


WeakAuras/Types.lua
```diff
@@ -3151,6 +3151,28 @@ LSM:Register("statusbar", "Thick Stripes", [[Interface\AddOns\WeakAuras\Media\Te
...
+    end
+    -- Capitalize the first letter of each word
+    for i, word in ipairs(words) do
+      words[i] = word:sub(1, 1):upper() .. word:sub(2):lower()
+    end
+    return table.concat(words, " ")
+  end
+
+  for power, data in pairs(PowerBarColor) do
+    if type(power) == "string" and data.atlas then
```
InfusOnWoW, line 3169: There are apparently more files here which can be deduced from the atlasElementName. There are lots of variants of those files for various settings, bit I think including:
"UI-HUD-UnitFrame-Player-PortraitOff-Bar" .. data.atlasElementName would probably be a good idea. All the other variants aren't imho that useful.

WeakAurasOptions/RegionOptions/AuraBar.lua
```diff
@@ -9,18 +9,73 @@ local L = WeakAuras.L;
...
+      func = function()
+        local path = {}
+        local paths = {}
+        for child in OptionsPrivate.Private.TraverseLeafsOrAura(data) do
+          paths[child.id] = path
+        end
+        OptionsPrivate.OpenTexturePicker(data, paths, {
+          texture = "textureInput",
+          color = "color",
+          mirror = "mirror",
```
InfusOnWoW, line 68: there are no mirror or blendMode settings on AuraBars, so these should be removed here.

InfusOnWoW:
I still think LibSharedMedia should have added code for those, but reading through the discussion, it seems there wasn't much interest in that. So yeah, this is it.

PR #5063, opened 2024-05-07
fix

mrbuds:


emptyrivers (approved):
I ran into the same thing with bisector, the return value is an [enum](https://warcraft.wiki.gg/wiki/API_C_AddOns.GetAddOnEnableState#Returns) if you feel like being 'idiomatic' about it.

mrbuds:
> I ran into the same thing with bisector, the return value is an [enum](https://warcraft.wiki.gg/wiki/API_C_AddOns.GetAddOnEnableState#Returns) if you feel like being 'idiomatic' about it.

Sure i'll add it, it makes the code more readable and the lookup time is not worth caring here

PR #5079, opened 2024-05-13
Optimization for running trigger function less often

mrbuds:


WeakAuras/Prototypes.lua
```diff
@@ -4975,9 +4975,15 @@ Private.event_prototypes = {
...
     },
-    internal_events = {
-      "WA_UPDATE_OVERLAY_GLOW"
-    },
+    internal_events = function(trigger)
+      local name
+      if (trigger.use_exact_spellName) then
+        name = tonumber(trigger.spellName)
+      else
+        name = type(trigger.spellName) == "number" and GetSpellInfo(trigger.spellName) or trigger.spellName
```
InfusOnWoW, line 4983: you are implicitly relying on GetSpellInfo returning the same name in both internal_events and loadFunc. I'm not sure whether that's guranteed with overloaded spells, that change depending on e.g. talent selections. If I recall correctly those functions do 

IIRC, correctly internal_events is called in Add, whereas loadFunc should be called on loading.

mrbuds: So a solution would to have a new `loadInternalEventFunc` field that is run in `LoadEvent` function, next to where `data.loadFunc(data.trigger)` is called

mrbuds: Implemented and pushed that solution
also fixed missing change on `Cooldown Ready (Spell)`

InfusOnWoW (approved):

PR #5091, opened 2024-05-22
LibSpecialization support on Cataclysm

mrbuds:


mrbuds:
Tested solo with unit = smart group for aura and power trigger
Load condition seems to also work correctly

mrbuds:
For testing before lib is actually release on CF:

[message(1).txt](https://github.com/WeakAuras/WeakAuras2/files/15403627/message.1.txt)

InfusOnWoW (approved):

PR #5100, opened 2024-05-27
BuffTrigger2: remove old UnitAura api

mrbuds:
Both classic_era and cataclysm has their UNIT_AURA event work like in Retail

Remaining difference is they don't have new GameTooltip.SetUnitBuffByAuraInstanceID functions, and C_TooltipInfo..GetUnitBuffByAuraInstanceID functions, and AuraUtil.ForEachAura is missing

For tooltips function we need to keep using previous functions, which require to find aura index from auraInstanceID

I have concerns about `GetIndexFromInstanceID` cost called from `UpdateToolTipDataInMatchData`

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -52,9 +52,6 @@ local Private = select(2, ...)
 -- Lua APIs
 local tinsert, wipe = table.insert, wipe
 local pairs, next, type = pairs, next, type
-local UnitAura = UnitAura
-
-local newAPI = WeakAuras.IsRetail()
 
 ---@class WeakAuras
 local WeakAuras = WeakAuras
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
Can you pester blizzard about the missing GetUnitBuffByAuraInstanceID? I wouldn't want to search for the index.

mrbuds:
> Can you pester blizzard about the missing GetUnitBuffByAuraInstanceID? I wouldn't want to search for the index.

There is https://github.com/Stanzilla/WoWUIBugs/issues/561

InfusOnWoW:
I think that full iteration to get the tooltip is not something we should do. It's probably not that bad, since fetching the tooltip is probably not something people do that often. But if a user has that enabled on a aura that shows a lot of buffs, we do lots of duplicated iterations. 

And I don't see that much of a benefit in getting rid of the old code atm, even though that probably means we have to mainain it until mist of pandaria classic, when hopefully it's synced again.

PR #5122, opened 2024-06-07
Reputation trigger fixes

Boneshockx:
Ensures we collect all player factions regardlesss of display settings.

`ExpandAllFactionHeaders()` does not correctly expand all faction headers. If subheaders are present it would remove all factions under the subheader which gives us an incomplete list. This PR fixes that by expanding headers individually.

### DIFF
--- a/WeakAuras/Types.lua
+++ b/WeakAuras/Types.lua
@@ -1635,15 +1635,34 @@ local function InitializeReputations()
   ---@type table<string, boolean>
   Private.reputations_headers = {}
 
+  -- Ensure all factions are shown by adjusting filters
+  local showLegacy = true
+  if not Private.ExecEnv.AreLegacyReputationsShown() then
+    showLegacy = false
+    C_Reputation.SetLegacyReputationsShown(true)
+  end
+  local sortType = 0
+  if Private.ExecEnv.GetReputationSortType() > 0 then
+    sortType = Private.ExecEnv.GetReputationSortType()
+    C_Reputation.SetReputationSortType(0)
+  end
+
+  -- Dynamic expansion of all collapsed headers
   local collapsed = {}
-  for i = 1, Private.ExecEnv.GetNumFactions() do
-    local factionData = Private.ExecEnv.GetFactionDataByIndex(i)
-    if factionData.isCollapsed then
-      collapsed[factionData.name] = true
+  local keepExpanding = true
+  while keepExpanding do
+    keepExpanding = false
+    for i = Private.ExecEnv.GetNumFactions(), 1, -1 do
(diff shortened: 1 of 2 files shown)


WeakAuras/Types.lua
```diff
@@ -1635,15 +1635,34 @@ local function InitializeReputations()
...
+  -- Dynamic expansion of all collapsed headers
   local collapsed = {}
-  for i = 1, Private.ExecEnv.GetNumFactions() do
-    local factionData = Private.ExecEnv.GetFactionDataByIndex(i)
-    if factionData.isCollapsed then
-      collapsed[factionData.name] = true
+  local keepExpanding = true
+  while keepExpanding do
+    keepExpanding = false
+    for i = Private.ExecEnv.GetNumFactions(), 1, -1 do
```
InfusOnWoW, line 1655: Can't you iterate forward and then don't need the outer loop?

Boneshockx: I've tried this:
```lua
  local collapsed = {}
  local index = 1
  local keepExpanding = true
  while keepExpanding do
    keepExpanding = false
    local factionData = Private.ExecEnv.GetFactionDataByIndex(index)
    if factionData.isHeader and factionData.isCollapsed then
      Private.ExecEnv.ExpandFactionHeader(index)
      collapsed[factionData.name] = true
    end
    if factionData.name then
      keepExpanding = true
    end
    index = index + 1
  end
  ```
  
  And although that collects all factions, it doesn't correctly restore collapsed headers later on.
  I think that is because of even more weird interactions in this API:
nEpCp2b

Boneshockx: I guess not doing a reverse loop somehow breaks index

InfusOnWoW: I pushed something to your branch which iterates forward (and thus only once). I can't reproduce the issue you show in that gif.

Boneshockx:
@InfusOnWoW with your commit:
I set everything to collapsed:
image

Then /reload
image

Seems to be only happening on Retail but not on Beta. I'm not annoyed by it personaly, it's just weird behaviour.

InfusOnWoW:
Hmm, that doesn't happen for me. If I collapse all headers, they stay collapsed. Are you sure it's caused by our code?

Boneshockx:
DataStore_Reputation addon snuck itself into my dev addon profile and that is conflicting. 
Its working fine with that disabled.
Excuse me for wasting time 🙃

InfusOnWoW (approved):

InfusOnWoW:
No worries, it's unusual that other addons interfer with WA development.

PR #5128, opened 2024-06-09
Smallpatches

InfusOnWoW:


mrbuds:
Moving `iconFunc` & `nameFunc` fields of trigger.args has the side effect of showing a "?" icon on aura's button
An example is an aura with a spell cooldown trigger, aura's button won't show icon of tracked spell anymore

For this `GetNameAndIcon` was added for the `Cast` trigger, and later for the `Currency` trigger

Is it something that would be address with the refactor you have in mind?

InfusOnWoW:
Ah, nah not immediate refactorings, more of a long termin thing:
a) Getting rid of these functions means I can remove a bunch of support code in GenericTrigger, and maybe rewrite the custom trigger handling in a way where a compability layer translates the old style to a tsu style trigger
b) Refactoring the Prototypes.lua options so that it's less arcane is easier if they are gone.

So yeah I'll have to add a bunch of GetNameAndIcon functions.

mrbuds (approved):

PR #5150, opened 2024-06-13
use table.concat to build various function strings

emptyrivers:
with 1600 auras (~8.9 MB SV file), this saves about 600MiB (40%) in memory allocation, and about 1 second (20%) in CPU time on login process
baseline-cpu
tconcat-cpu
baseline-mem
tconcat-mem

### DIFF
--- a/WeakAuras/Conditions.lua
+++ b/WeakAuras/Conditions.lua
@@ -451,19 +451,19 @@ local function CreateCheckCondition(data, ret, condition, conditionNumber, allCo
     check = "false"
   end
   if condition.linked and conditionNumber > 1 then
-    ret = ret .. "      elseif (" .. check .. ") then\n";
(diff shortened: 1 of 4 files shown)


WeakAuras/Conditions.lua
```diff
@@ -451,19 +451,19 @@ local function CreateCheckCondition(data, ret, condition, conditionNumber, allCo
...
-    ret = ret .. "      end\n";
+    table.insert(ret, "      end\n")
   end
 
   if (check) then
-    ret = ret .. "\n";
+    table.insert(ret, "\n")
   end
-  return ret, recheckCode;
+  return recheckCode;
```
InfusOnWoW, line 466: This should return table.concat(ret), recheckCode

InfusOnWoW: Or maybe not

InfusOnWoW (approved):

PR #5158, opened 2024-06-15
fix

mrbuds:


WeakAuras/Prototypes.lua
```diff
@@ -6904,19 +6903,67 @@ Private.event_prototypes = {
...
+          ]])
+          for index, value in pairs(trigger.herotalent.multi) do
+            table.insert(ret, ([[
+              local talentId = %s
+              local shouldBeActive = %s
+              if talentId then
+                activeName, activeIcon, _, rank = WeakAuras.GetTalentById(talentId)
+                if activeName ~= nil then
+                  local hasTalent = rank > 0
+                  if hasTalent then
```
InfusOnWoW, line 6932: I think this can all be simplified to:

Before the loop:
local heroActive = true

if (rank > 0) ~= shouldBeActive then
  heroActive = false
  break
end

Now, that changes behaviour if no herotalent is selected at all, if you want to preserve that behaviour, the easiest is imho to simply do that at the very top level, thus e.g.:

if (next(trigger.herotalent.multi) == nil) then
  table.insert(ret, [[
    local heroActive = false
 ]])
else 
  [[ loop over talents ]]
end

mrbuds: done

InfusOnWoW:
Ah the break didn't actually work. Since we aren't actually looping in the function, but when generating.

I've also noticed that the inverse check code had a minor bug where trigger.inverse was used even on retail.

And I ported this simplified version to the talent check.

InfusOnWoW (approved):

PR #5195, opened 2024-06-27
Allstates helper methods

mrbuds:
Add methods to the states/allstates table that helps with creating, updating or remove states in an optimized way

## Advantage of using this function instead of doing a states[key] = { ... }

- If already created, update existing state, and return true if any value was changed, this can help reduce amount of resources an aura use
- Automatically add some fields: ~~`progressType`, `expirationTime`, `icon`,~~ `show`, `changed`
- Automatically `return true` when using these functions and any change was made

## Examples

```Lua
function(states, event, ...)
    if event == "PLAYER_TARGET_CHANGED" then
        if UnitExists("target") then
            -- if state exists it's updated, not replaced
            -- show & changed fields can be skipped
            states:Update("", {
                    name = UnitName("target"),
                    duration = 5,
                    expirationTime = GetTime() + 5,
                    progressType = "timed",
                    autoHide = true
            })
        else
            -- wipe
            states:RemoveAll()
        end
    end
    -- no need to return true
end
```

with clones

```Lua
function(states, event, ...)
    local currentEssence = UnitPower("player", Enum.PowerType.Essence)
    local maxEssence = UnitPowerMax("player", Enum.PowerType.Essence)
    for i = 1, 6 do
        if i > maxEssence then
            states:Remove(i) -- wipe allstates[6]
        else
            local value = currentEssence >= i and 1 or 0
            local newState = {
                progressType = "static",
                value = value,
                total = 1
            }
            states:Update(i, newState)
        end
    end
    -- no need to return true
end
```

~~maintaining value of the variable `changed` when editing multiple clones is not great :(~~

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -670,9 +670,10 @@ local function RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2
       else
         ok, returnValue = xpcall(data.triggerFunc, errorHandler, allStates, event, arg1, arg2, ...);
       end
-      if (ok and returnValue) then
+      if (ok and (returnValue or (returnValue ~= false and allStates.__changed))) then
         updateTriggerState = true;
       end
+      allStates.__changed = nil
       for key, state in pairs(allStates) do
         if (type(state) ~= "table") then
           errorHandler(string.format(L["All States table contains a non table at key: '%s'."], key))
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -4470,7 +4470,7 @@ function GenericTrigger.GetOverlayInfo(data, triggernum)
           count = variables.additionalProgress;
         end
       else
-        local allStates = {};
+        local allStates = setmetatable({}, Private.allstatesMetatable)
         Private.ActivateAuraEnvironment(data.id);
         RunTriggerFunc(allStates, events[data.id][triggernum], data.id, triggernum, "OPTIONS");
         Private.ActivateAuraEnvironment(nil);
--- a/WeakAuras/TSUHelpers.lua
+++ b/WeakAuras/TSUHelpers.lua
@@ -56,8 +56,8 @@ local function recurseUpdate(t1, t2)
   return changed
 end
 
----@type fun(states: states, newState: state, key: key): boolean
-local update = function(states, newState, key)
+---@type fun(states: states, key: key, newState: state): boolean
+local update = function(states, key, newState)
   local changed = false
   local state = states[key]
   if state then
@@ -70,21 +70,21 @@ local update = function(states, newState, key)
   return changed
 end
 
----@type fun(states: states, newState: state, key: key): boolean
-local create = function(states, newState, key)
+---@type fun(states: states, key: key, newState: state): boolean
+local create = function(states, key, newState)
   states[key] = newState
   states[key].changed = true
   fixMissingFields(states[key])
   return true
 end
 
----@type fun(states: states, newState: state, key: key?): boolean
-local createOrUpdate = function(states, newState, key)
+---@type fun(states: states, key: key?, newState: state): boolean
+local createOrUpdate = function(states, key, newState)
   key = key or ""
   if states[key] then
-    return update(states, newState, key)
+    return update(states, key, newState)
   else
-    return create(states, newState, key)
+    return create(states, key, newState)
   end
 end
 
--- a/WeakAuras/TSUHelpers.lua
+++ b/WeakAuras/TSUHelpers.lua
@@ -1,0 +1,85 @@
+if not WeakAuras.IsLibsOK() then return end
+---@type string
+local AddonName = ...
+---@class Private
+local Private = select(2, ...)
+
+---@alias key string | integer
+---@alias states table<key, state>
+
+---@type fun(state: state)
+local function fixMissingFields(state)
+  if type(state) ~= "table" then return end
+  -- set show
+  if state.show == nil then
+    state.show = true
+  end
+end
+
+---@type fun(states: states, key: key): boolean
+local remove = function(states, key)
+  local changed = false
+  local state = states[key]
+  if state then
+    state.show = false
+    state.changed = true
+    changed = true
+  end
+  return changed
+end
+
+---@type fun(states: states): boolean
+local removeAll = function(states)
+  local changed = false
+  for _, state in pairs(states) do
+    state.show = false
+    state.changed = true
+    changed = true
+  end
+  return changed
+end
+
+---@type fun(states: states, newState: state, key: key): boolean
+local update = function(states, newState, key)
+  local changed = false
+  local state = states[key]
+  if state then
+    fixMissingFields(newState)
+    for k, v in pairs(newState) do
+      if state[k] ~= v then
+        state[k] = v
+        changed = true
+      end
+    end
+    if changed then
+      state.changed = true
+    end
+  end
+  return changed
+end
+
+---@type fun(states: states, newState: state, key: key): boolean
+local create = function(states, newState, key)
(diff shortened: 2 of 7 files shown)


WeakAuras/AuraEnvironment.lua
```diff
@@ -520,7 +520,8 @@ local overridden = {
   WA_Utf8Sub = WA_Utf8Sub,
   ActionButton_ShowOverlayGlow = WeakAuras.ShowOverlayGlow,
   ActionButton_HideOverlayGlow = WeakAuras.HideOverlayGlow,
-  WeakAuras = FakeWeakAuras
+  WeakAuras = FakeWeakAuras,
+  tsu = Private.TSU.createOrUpdateOrWipe
```
emptyrivers, line 524: I think we should try to avoid creating any more magic globals. I don't think there's any good reason to not just have this be in WeakAuras global?

emptyrivers (commented):
I like the idea, in general.

have you considered "just" making those functions be part of a mixin so that they're magically part of tsu states param? or is that maybe going too far 🤔

WeakAuras/AuraEnvironment.lua
InfusOnWoW: I think the cleanest is to put it into the allstates table via a metatable, that means that it's almost perfectly backwards compatible, as iterating allstates doesn't see the metatable.

So allstates:Update(key, { .... }) (which does CreateOrUpdate, but that's a too long name)
allstates:Remove(key) and maybe allstates:RemoveAll() would be my suggestion.

mrbuds: updated PR & examples to do it this way

WeakAuras/TSUHelpers.lua
```diff
@@ -0,0 +1,85 @@
...
+  local changed = false
+  for _, state in pairs(states) do
+    state.show = false
+    state.changed = true
+    changed = true
+  end
+  return changed
+end
+
+---@type fun(states: states, newState: state, key: key): boolean
```
InfusOnWoW, line 42: That feels like the wrong order to me, "key, state" feels a lot more natural to me.

I know you did that for key to be optional, but I still feel that its the wrong order.

mrbuds: swapped them, and made update function check differences recursively (eg. for overlays)

emptyrivers:
> maintaining value of the variable changed when editing multiple clones is not great :(

It's not impossible that we could magic that inconvenience away, though it might be considered breaking.

Currently, we apply new state to its region if:
- state.changed is truthy, and
- state updater function returns truthy

But consider this:
apply new state to region if:
- state.changed is truthy, and
- **state updater function does not return false**

... so after writing it out, it's definitely a breaking change (though, possibly low impact - how many people actually use the TSU return value as designed?), but it does afford us the nice advantage of not having to manually track if any state is changed.

main downsides:
- it *is* still breaking
- potential performance problems could happen in TSU with many states. Not impossible for author to solve (the same risks exist with current behavior). But there are potential magics we could do to improve that too.

emptyrivers:
well, that is all probably beyond the scope of this ticket anyways.

mrbuds:
What if states:Update/Remove/RemoveAll set a hidden flag `allStates.__changed`, to make auras using these new functions not having to return anything without breaking (afaik __fields are ignored from pairs(allStates) right?)

RunTriggerFunc :

```diff
--    if( (ok and returnValue) or optionsEvent) then
++    if( (ok and (returnValue or (returnValue ~= false and allStates.__changed))) or optionsEvent) then
++      allStates.__changed = nil
        for id, state in pairs(allStates) do
          if (state.changed) then
            if (Private.ActivateEvent(id, triggernum, data, state)) then
              updateTriggerState = true;
```

mrbuds:
ha shit `All States table contains a non table at key: '__changed'.
[string "@Interface/AddOns/WeakAuras/GenericTrigger.lua"]:647: in `

mrbuds:
@emptyrivers with https://github.com/WeakAuras/WeakAuras2/pull/5195/commits/3b04d7aedfc93f71f8e3683ed34c3e14caa48613 there is no need to return true when Update|Remove|RemoveAll functions made a change to allStates table.
It doesn't change behavior of code that don't use these functions

WeakAuras/GenericTrigger.lua
```diff
@@ -639,9 +639,10 @@ local function RunTriggerFunc(allStates, data, id, triggernum, event, arg1, arg2
       else
         ok, returnValue = xpcall(data.triggerFunc, errorHandler, allStates, event, arg1, arg2, ...);
       end
-      if (ok and returnValue) then
+      if (ok and (returnValue or (returnValue ~= false and allStates.__changed))) then
```
emptyrivers, line 642: it is perhaps a tiny bit weird that code which doesn't use the utility functions can set `__changed = true` but on balance i think it's fine

emptyrivers (approved):

emptyrivers:
technically a breaking change for any code that used `states["__changed"]` state

InfusOnWoW:
So, I have lots of thoughts about this. When I designed the new trigger state mechanism a long tiem ago, I did consider various ways in which triggers communicate what has changed with the rest of WA. The ideal interface is that the user just writes to the states table and each state, and WA handles the syncing of that to the display/conditions/text replacement/dynamic groups and so on.

I rejected that, because this makes the case that nothing has actually changed pretty expensive, and that's not an uncommon case at all. That's why the user and triggers have to return true, and why the have to set .changed to true. Similarly the only purpose of .show = false, is that it's easier to check for than a missing cloneId.

The second tricky question back then, how granular should the change tracking be. I quickly settled on one changed flag.

I've been thinking about a new trigger state framework, which rexamines those two things and maybe gets us that ideal interface and a finer change tracking and even some further optimization potential, though it's been on the always long future ideas list. (It's not too unsimilar to what rivers did to the dynamic group,)

Now, so the question for me is, is that an idea I'll ever come around to, and if so, how does this PR interact with that. I'll try to do a bit of experimentation

mrbuds:
> I rejected that, because this makes the case that nothing has actually changed pretty expensive, and that's not an uncommon case at all. That's why the user and triggers have to return true, and why the have to set .changed to true

In practice people are lazy and i see all the time auras that overwrite allstates table without caring if anything was changed, and return true in any case, even if allstates table wasn't touched (like CLEU:SPELL_AURA_APPLIED => guid ~= myGUID => return true)

This keep old behavior, while giving access to simplified method to make code that doesn't fall in these traps (and you can still return false)

I was also thinking this would help with idea to convert current generic triggers to small/efficient/readable enough TSU

InfusOnWoW:
Yes I agree that in pratice people don't use changed very much, so in that sense the API is less than ideal and certainly your changes are an improvement as that makes editing states easier.

The concern I have is that introducing such a convenient API might clash with future changes on how TSU triggers work. That is if we add a TSUv2, where the api is just setting up states and WA handles the rest automatically, without any requirement to set .changed or even return true, then this might be a dead-end API, and adding it will just confuse users.

I've started a bit of research if that "perfect" api is possible and so far so good. My small prototype does behave like I want it to behave, and I want to push it a bit further to decide whether that's going to be the next "big" project I want to tackle. It's something I have been considering for a while and it would unlock a bunch of potential features. In any case, I don't intend to take too long to figure out what the right direction is.

mrbuds:
> I don't intend to take too long to figure out what the right direction is.

I'm glad this PR is giving you motivation for greater changes, but take the time you need

InfusOnWoW:
So I experimented with some bigger changes to our state handling, which would probably have let to a TSU v2.  But that experiment is on hold for some time and I'm likely not going to pursue that further in the nearer future.

As such, I think this is fine to merge. I'm not entirely happy about the magic __changed variable, but it's probably an acceptable hack for now.

InfusOnWoW (approved):

PR #5212, opened 2024-07-06
Smallpatches

InfusOnWoW:


WeakAuras/Types.lua
```diff
@@ -2757,6 +2757,7 @@ end
 
 ---@type table
 Private.instance_difficulty_types = {}
+WeakAuras.instance_difficulty_types = Private.instance_difficulty_types
```
mrbuds, line 2760: Is this intentional?

InfusOnWoW: oh no that was debugging code

mrbuds (approved):

PR #5213, opened 2024-07-06
spellCache

mrbuds:
I don't get why spellIds / iconIds were concatenated as strings instead of table of numbers, this should help with the known issue when storing a lot of strings in SV, and should run faster

comment with unused block of code is left to find new "holes" in spellIds for next classic versions

i tried it on classic_era/classic/retail/tww, retail & tww don't have such spellids holes

### DIFF
--- a/WeakAurasOptions/Cache.lua
+++ b/WeakAurasOptions/Cache.lua
@@ -31,6 +31,24 @@ function spellCache.Build()
     return
   end
 
+  local holes
+  if WeakAuras.IsClassicEra() then
+    holes = {}
+    holes[63707] = 81743
+    holes[81748] = 219002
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
The reason for the strings is found in https://github.com/WeakAuras/WeakAuras2/commit/829a4e37cd6650e585c2fd08c3311708f1db6cb5. 

I can still reproduce that, after forcing a spell cache regeneration by deleting the WeakAurasOptions.lua file, on next reload I have:
"3x Error loading WTF/Account/TESKE/SavedVariables/WeakAurasOptions.lua: constant table overflow" in the bugsack.

InfusOnWoW (approved):

PR #5220, opened 2024-07-11
Text editor: support ctrl z, ctrl y

mrbuds:


WeakAurasOptions/OptionsFrames/TextEditor.lua
```diff
@@ -720,13 +720,57 @@ local function ConstructTextEditor(frame)
...
+      end
+      local cursorPosition = self:GetCursorPosition()
+      local text = originalGetText(self)
+      if IndentationLib then
+        text, cursorPosition = IndentationLib.stripWowColorsWithPos(text, cursorPosition)
+      end
+      if self.timeMachine[1] and text == self.timeMachine[1][1] then
+        return
+      end
+      table.insert(self.timeMachine, 1, {text, cursorPosition - 1})
```
InfusOnWoW, line 769: There's a bug here, you need to clear the "redo" part of the history. That is if the you need to remove all entries from 1 .. self.timeMachinePos -1.

The bug can be demonstrated by entering:
ABC + Ctrl+Z
=> Text show in edit: AB
=> timeMachinePos = 2
=> timeMachine = { ABC, AB, A }

Now enter
D
=> Text show in edit: ABD
=> timeMachinePos = 1
=> timeMachine = { ABD, ABC, AB, A }
Ctrl +Z
=> Text shown in edit: ABC instead of the correct AB


I btw, would reverse the timeMachine, so that new entries are added to the end, not at the first pos. But that might be just personal preference.

mrbuds: bug fixed and added 2 unrelated commits

mrbuds:
about the chat trigger change, i checked that RegisterEvent doesn't complain with "CHAT_MSG_COMMUNITIES_CHANNEL" and it doesn't error for both

InfusOnWoW:
I have no idea what CHAT_MSG_COMMUNITIES_CHANNEL is either, but adding it should be fine.

InfusOnWoW (approved):

Issue #5191, opened 2024-06-26
unable to find mixin (ColorMixin) - TWW Bug


Peter-Day:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues.

### Description

Found someone else who had a similar issue in the pull request thread for the initial TWW alpha version, but I realized this bug might have gotten lost there, so posting a separate bug report for it.

Some wow api functions, when called inside of a weakaura script, will give the error in the title: "unable to find mixin (ColorMixin)"

I found this bug trying to use C_ChallengeMode.GetKeystoneLevelRarityColor. The other comment was using C_TooltipInfo.GetHyperlink. See: https://github.com/WeakAuras/WeakAuras2/pull/5071#issuecomment-2133352503

Of course, there are likely many others.

### WeakAuras Version

WeakAuras 5.14.1

### World of Warcraft Flavor

Beta (WoW 11.0)

### World of Warcraft Region

US/NA

### Tested with only WeakAuras

I got this issue with only WeakAuras enabled

### Lua Error

```Text
12x Lua error in aura 'New 2': Custom Text Function
WeakAuras Version: 5.14.1
[string "return "]:4: unable to find mixin (ColorMixin)
[string "=[C]"]: in function `GetKeystoneLevelRarityColor'
[string "return "]:4: in function <[string "return "]:3>
[string "=[C]"]: in function `xpcall'
[string "@WeakAuras/WeakAuras.lua"]:4862: in function `RunCustomTextFunc'
[string "@WeakAuras/SubRegionTypes/SubText.lua"]:330: in function `?'
[string "@WeakAuras/SubscribableObject.lua"]:94: in function `Notify'
[string "@WeakAuras/WeakAuras.lua"]:4600: in function <WeakAuras/WeakAuras.lua:4595>
[string "@WeakAuras/WeakAuras.lua"]:4687: in function <WeakAuras/WeakAuras.lua:4653>
[string "@WeakAuras/WeakAuras.lua"]:4797: in function `UpdatedTriggerState'
[string "@WeakAuras/WeakAuras.lua"]:4517: in function `UpdateFakeStatesFor'
[string "@WeakAuras/WeakAuras.lua"]:4490: in function `FakeStatesFor'
[string "@WeakAurasOptions/AceGUI-Widgets/AceGUIWidget-WeakAurasDisplayButton.lua"]:1539: in function `SyncVisibility'
[...]

PR #5221, opened 2024-07-11
add more mixins to exec_env

emptyrivers:
found these by grepping through blizzard interface code with `/Mixin = ".*Mixin"/`

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -126,6 +126,13 @@ globals = {
 	"TableHasAnyEntries",
 	"tIndexOf",
 	"UISpecialFrames",
+	"Vector2DMixin",
+	"Vector3DMixin",
+	"ItemLocationMixin",
+	"ItemTransmogInfoMixin",
+	"TransmogPendingInfoMixin",
+	"TransmogLocationMixin",
+	"PlayerLocationMixin",
 
 
 	-- FrameXML frames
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
did you check that these are problematic? I think the problem is specific to how blizzard implemented some functions and not actualy to Mixins.

emptyrivers:
Vector2DMixin, at least, has appeared in the wild with a similar issue to #5191, in the [wowuidev discord](https://discord.com/channels/327414731654692866/327414731654692866/1261063378382880919)

InfusOnWoW:
Right so my best explanation on what is happening is this, and why this is probably a good idea:

They have written themselves a framework where they can automatically map certain C++ types to Lua Mixins, the C++ type for e.g. ColorMixin is presumably "colorRGB". (No idea whether that's per type or per function information, doesn't really matter.)

The mixing of the ColorMixin table is crucially not done on the lua side, but done via C++ code. And that C++ code is buggy and looks only at the direct function environment, instead of upwards.

We can, I think, observe where they have done such a thing, by looking at the generated api documentation, whenever there's a Returns = { Type = "foo", Mixin = "bar" } line. That is, I assume, the framework both is responsible for the generated documentation and the wrapping code for looking up the mixin.

InfusOnWoW (approved):

PR #5226, opened 2024-07-15
Auto-hide dynamic text window

Boneshockx:
Automatically hide the window when input loses focus.

This PR also adds custom variables from TSU that have no display value to the list of available text replacements. In the case of no display value it falls back to the var's name

### DIFF
--- a/WeakAurasOptions/ActionOptions.lua
+++ b/WeakAurasOptions/ActionOptions.lua
@@ -160,7 +160,13 @@ function OptionsPrivate.GetActionOptions(data)
         callbacks = {
           OnEditFocusGained = function(self)
             local widget = dynamicTextInputs["start_message_dest"]
-            OptionsPrivate.ToggleTextReplacements(data, true, widget)
+            OptionsPrivate.ToggleTextReplacements(data, widget, "OnEditFocusGained")
+          end,
+          OnEditFocusLost = function(self)
+            OptionsPrivate.ToggleTextReplacements(nil, "OnEditFocusLost")
+          end,
+          OnEnterPressed = function(self)
+            OptionsPrivate.ToggleTextReplacements(nil, "OnEnterPressed")
           end,
           OnShow = function(self)
             dynamicTextInputs["start_message_dest"] = self
@@ -177,7 +183,7 @@ function OptionsPrivate.GetActionOptions(data)
         hidden = function() return data.actions.start.message_type ~= "WHISPER" end,
         func = function()
           local widget = dynamicTextInputs["start_message_dest"]
-          OptionsPrivate.ToggleTextReplacements(data, nil, widget)
+          OptionsPrivate.ToggleTextReplacements(data, widget, "ToggleButton")
         end,
         imageWidth = 24,
         imageHeight = 24,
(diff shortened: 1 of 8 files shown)


WeakAurasOptions/ActionOptions.lua
```diff
@@ -716,7 +734,13 @@ function OptionsPrivate.GetActionOptions(data)
         callbacks = {
           OnEditFocusGained = function(self)
             local widget = dynamicTextInputs["finish_message"]
-            OptionsPrivate.ToggleTextReplacements(data, true, widget)
+            OptionsPrivate.ToggleTextReplacements(data, widget, "OnEditFocusGained")
+          end,
+          OnEditFocusLost = function(self)
+            OptionsPrivate.ToggleTextReplacements(nil, "OnEditFocusLost")
```
InfusOnWoW, line 740: the second argument to ToggleTextReplacement is the widget, you have multiple places where you don't pass in the right amount of parameters.

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -897,14 +897,32 @@ function OptionsPrivate.CreateFrame()
...
-  function OptionsPrivate.ToggleTextReplacements(data, show, widget)
-    if show or not dynamicTextCodesFrame:IsShown() then
+  function OptionsPrivate.ToggleTextReplacements(data, widget, event)
+    -- This throttle prevents double toggling as a result from OnEditFocusLost
+    local currentTime = GetTime()
+    if event ~= "OnEditFocusGained"
+    and dynamicTextCodesFrame.lastCaller
+    and dynamicTextCodesFrame.lastCaller.event == "OnEditFocusLost"
+    and currentTime - dynamicTextCodesFrame.lastCaller.time < 0.1
+    and not (event == "OnEnterPressed" and dynamicTextCodesFrame:IsMouseOver())
```
InfusOnWoW, line 907: I don't understand what this is guarding against.

Boneshockx: When the frame is opened and then the user clicks the ToggleButton to close the frame, that action would first fire OnEditFocusLost and thus hide the frame first before the toggle button's callback runs and show the frame again. 
The throttle prevents that by checking if the frame was recently hidden from OnEditFocusLost.

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -898,13 +898,16 @@ function OptionsPrivate.CreateFrame()
   dynamicTextCodesFrame:Hide()
 
   function OptionsPrivate.ToggleTextReplacements(data, widget, event)
-    -- This throttle prevents double toggling as a result from OnEditFocusLost
+    -- If the text edit has focus when the user clicks on the button, we'll get two events:
+    -- a) The OnEditFocusLost
+    -- b) The ToggleButton OnClick event
+    -- Since we want to hide the text replacement window in that case,
+    -- ignore the ToggleButton if it is directly after the  OnEditFocusLost
```
InfusOnWoW, line 905: The code will be read and even modified multiple times in the future, so if we introduce hacks like this we to should ensure that whoever touches the code in the future can understand which case lead to the hack.

Also the hack should be as specific as possible, which means I changed to exactly check for OnEditFocusLost -> ToggleButton.

Looking at the time difference between the events, they were pretty close to 0.1 on my machine, 0.1 didn't feel safe enough.

InfusOnWoW (approved):

PR #5229, opened 2024-07-16
Add warband currency support

Boneshockx:
# Description

TWW brings warband currencies that can be transferred over to other characters within your warband. 
The C_CurrencyInfo API contains currencyData for all warband characters, and so we're able to create a cloning trigger that shows currency info for each character in the warband.

This is a full rework of the currency trigger prototype as this needs to use `statesParameter = "full"`.

Several state values for account wide currencies have been added:
- accountQuantity (sum of all account characters' quantity)
- realAccountQuantity (total account quantity after transferring everything to your current character and paying the transfer cost)
- realCharacterQuantity (remaining quantity if all this character's currency was transfered and paid the transfer cost)
- transferPercentage (transfer cost %)
- unitName (comes from currency API)
- unitGUID (comes from currency API)
- mainCharacter (bool value to check for the character you are currently playing on)

## Example
image
```
!WA:2!vI1(ZTXv1FRikDy7dSLBstAsBvCE0MKwxhx7usN4o41psCX12zLSDzOG2v7EL2BYQDV9URSJDBhAfbW88lJPq4nfXB(2cfbL3qbZJodZKzoJg(damu6WRbM8paCU3DLLT1QqW)G0Q7EUN3NpNZXjMlLF)JswivPuwPSw5(vlY9wuD0PNFQ1Z7XTiCvdZZBX9ySovDOlVSb3kDwppNakJFHPluWNeCM3CLx7F)0D0rh8LACWt(YYdQuKtTYUeJ0J2Ovlu2Xzek30HOR5Zmmjxh3WmG656Fcn)adEGs(cuxQVTIk(vGsLaoTyrc3)2omp6X3tvls(Yfkiyj)mJn5mJp7KQb4p0mkZn6VIpJ44mHLVY6(LZtwG4gKbPMEHA5gz4mzZLj7WAzvlJmppZXyjcxZ1OeXxrtsA(ZqmCcS34QZWj4v1YmZyto5QLDJubLAMEoLl5MrybjZZnSOL9V4nUQHlTKHWAobCJJVgXWNKjGtClgyF9q3QUEUK6wOokOiNqJ5(etpxl)kcsf2dCq1sgu3XHdGxaoiCi4WWDHFFtB)endhArx1jhB8S1kWrdafKrGrso6efAyIQOfOrkk8Sk5994bt7YnTnClsSuD8mSgpFGHdAG7sRuzmoQOIUnZDb9ROz6y47lEs1NUmr8qvZY(bELYGCH9Qfk7kdy3TX9Ko)rusJ)rlK2OxUuy9IYpG0l1k9qdLUNrkZrZ3CPSe)G0dBA6HEW0Zz4uM43tA5v94PZ))8vL3mWM4kFq8hNeuM7MUGHJprEiX1sztVyBQ3cc(K(udTDzlpxbVB98rj8J454XF4KjtotY(6DWv9jofMXJI5bzNEgHZNhgYj21clucPprIej18nrhCYQMOtMWL5)tQLpKk7QHFpMvrcRRmpEzdoj94yPr65TPbKv9diSHDl6qUZicZGrIRtZHwIgCdRAWnNuMuDXDv3OCGThFAwqyGgZMkqlQun0QecDnRLW8BQjwrxMTo(E0sDdghl68451gE0jMntLY(KjfS2ETqPfwa)gWCmMTrYQbEMlGPQid3DIx4f7HAX6CEdEEdxR0nIsvq8IWIHvfv7ZtTcSVHRy4AIk34I0tHQKpZiAJn2uxHkCiUgoZfY0l1FzQvnsXjM5CNK)wkmXSrUYjCrLiPQi3nFycyTqgkda5hzSPYoMwvr9dn06RrDl4Xdlavqnkq(uIuMhDLJuLIeMXRm3KSVoQzyDoKHeRhX4c7O5pOU7aU(eWBdES6Mi4ML3IUzwKYi6R14NI4LnCdy1TcCtkWnRaVXB(WqNRCywxOpmxrIlHtnZy7T40U6SUfNzeM6E2YOxNgSKnRBoIWm8wpnfMzjqTMc9ujUI4ALmUqZRKsYirvSeE6r8Si6WUGBLDlBISCEiEIbgvt9qWER0iWSQ4Qi0G551RkE00bbH81LhlZ2TxVjZLmMDhcfCeBdbWmH3c77zOHGDd7HLoYYAiPwj8HgILwWCejYxOcbrwTOOXVGOwXpqNT3n5L2oV0z3ABEt)3FF9PjnG(pEF9X292CZBrB7csb3McIEUB2TkewmbaB42H7O2M8M93pY1DgRNqDGbp(Bs61mJuQjS0xxQlnf7PgQ(wYgw3x(1irjs1L(LgHCDyNkIIWCIUs6vBY2r5S90qJBrpSz7lglPzAWPUY2CkANOVh4K9b3PcRlrp3fizd7Mjc7pFhWSx6kNNG4oIgbbAIAhBOhrJST3m6U)V3EcBr7BeiA1H1l37dVlO76ywGauvGd5ATY9XsfqUqqoFBd0F8OHiojy7rEO8JWs5C(rpahC1nEx9d9ehVxHV60ZoXOpf7(A5sTl1SbRy90YvAYWOtulwwG0TjTSj2olDl3)X3UmUs4BPxGech(mXCPJ3llxaTej63jyhSfsIiWHu0WezUJNh3MT)RM6BkutyGQsaqswKi1PMEQXQkj3Sr)SKBja8wJca1LNTiI)op3Gvz(OhwtE8cuFAECWn2UJrlJKA9OtXHUWww7NDK2zqrTLYfyZjyTHJ1sSd3E)Zwn)9edtJC7pBSCzdFJr(8CYcurA56HMFJg6QdpB2PzDlpeBQ6jAJykdCQdJ)m0diAuqlSe8yXebyBI5c47Uy3v7YlBnF82VQSZMDG27AyCIjv0hnzm(fZneb0nhRafVlgHfcETH2CvDHcmNs5eZmR5IJiZz7nwDl8rnHkAfB5wK6xYZkNaMqF1nYCyDooNUCA0lzjMPnD2SX6FAQiWlfLmhoFquY8ozhR9InM0V93Uu1ME46Hcyd4Qo3KuLzr1MEwCOKCAtC6ZKLDhTJHnQ2p61wQAy2u62XTn8G13I5NOzPOCtjuZMCIPgBLbH1sa)FycXhHT)dH(J03Tann2omp1rGvrc)OWZcFmjUh8XXFEPNb(ejypym6E8doKNwmNB5s5j8ikzDRIuE(zztAWlsMs(kF4tbdaFA4ZaFwj6e85saFE45GVGESzJcj1sve8HtaFX9dFPLGFd8kX0vODky4RxvLwmuxGpPn8LTHxob8vql)xPd)C4xaFn4RdFd4)h(zjHVjkQFDc4f6cEXKTb0QfBg(TS7jgkdIgfAgcxSNGrXMvHW3bReV29YwyoAjdNnLSc)y4LWEY1GFi8JGFGo8tWE8)0LGVlE6Zdpg89GldFBumFlB4RAd)YUGQWa)56fD8wCCo5Xll6BoZRRVE7FWqKdXlwdNf50orncTRiosKAvpF5Gap3PrWaCH6QIJd3pzN1ep3iFC1InUQmcVM4NJgTnCsjLnQRKArwBQ55Xzv9tkVyg52uQIhTLhmjfFzxsPPgUuvEBcTODWzH(MtoFReEYgoUyfwya5N3V8ZbfFY6ePz4Oz)LRcydzGSB)yQRnCgvX2dWFshEhkW4kWjXga3sJDcenyhL6BGDgTSH5Qn4PN8bgP)hD5PMFe1L98kLaE9jGPBS50TVL9A3imMEe5C5WXsctLuBrrBNZkh7VXaJ4mAAIAC40kCQRyDmITwZM5rBOl0L(vuuG3UsLg6N96mUxreLZpC9NJUVo2Xkdd67RduXQzr9f)xqexKrBSvNCn7hmnIoSTgwkdxqOSnYAtBItWlPlMbrdXvAnb)Po0Hos0AwS0Bs8TbJ8(IHKyLxJcNlhTF2nlwrJ1t7LWg4MWhWwesHpOTQaTeRw6Bo4p0(eNtgRkftPCmOa13YaVjH)Oi7suFdVQDywcMaGb5QnhC8I3ioYCGdMUdVcKFh13YeByu0uN1zZW)Sml0MG9U(whIbrUcxv9aaAWfSHf7FW(6dStauB4CYTjHZl2CeCGs6aM0Jt)Z0HNbQi2DkqhkdlalBlxMAf49j2Mcwk1PGlQdftnOyJqWxhEAKfVtD4jGNexPrStf8U0H39OC49IVqSriU9b8(F(oyhlwFyBMp6YSd3(W4wgne(xI07O1we7JGRLSP1w22j3ulNCV3gUQc8HGF)zH1TR0ym29WoX1K6gt8g(R6W)0M9GxZzXX154yT36BDog4Vh2a1yhW)yhiAf7aT)2BQzbcUn7LQ0yYF454nM19OWF5ScKR9Ev(pYHWElEUjk5FeT5OdmGafdrsHxlbR3R1sLnkE)B2S9gRchrX3V22xgdb7e4Yw2Pw439O)Nd
```


WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,150 @@ Private.event_prototypes = {
...
+        state.discovered = currencyInfo.discovered
+        state.transferPercentage = currencyInfo.transferPercentage
+        state.unitName = UnitName("player")
+        state.unitGUID = UnitGUID("player")
+        state.mainCharacter = true
+
+        if clone and not primaryCheckFailed then
+          for i, currencyData in ipairs(accountCurrencyData) do
+            local cloneId = currencyData.characterGUID or tostring(i)
+            states[cloneId] = CopyTable(states[""])
```
mrbuds, line 11382: are you sure about copying data from current character on warband characters?

WeakAuras/Prototypes.lua
```diff
@@ -11224,18 +11224,62 @@ Private.event_prototypes = {
       ["events"] = {
         "CHAT_MSG_CURRENCY",
         "CURRENCY_DISPLAY_UPDATE",
+        "ACCOUNT_CHARACTER_CURRENCY_DATA_RECEIVED",
+        "CURRENCY_TRANSFER_LOG_UPDATE",
```
mrbuds, line 11228: i was expecting this to error on cataclysm, but it didn't

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,150 @@ Private.event_prototypes = {
...
+
+        if clone and not primaryCheckFailed then
+          for i, currencyData in ipairs(accountCurrencyData) do
+            local cloneId = currencyData.characterGUID or tostring(i)
+            states[cloneId] = CopyTable(states[""])
+            local s = states[cloneId]
+            s.value = currencyData.quantity
+            s.realCharacterQuantity = currencyData.realQuantity
+            s.unitName = currencyData.characterName
+            s.unitGUID = currencyData.characterGUID
```
mrbuds, line 11387: there is no `unit` involved in this trigger, so i would rename these to `characterName` and `GUID`

WeakAuras/Prototypes.lua
Boneshockx: I'm 50/50 about it. I had to choose between copying the main state and overwriting character specific values, or using a fresh state and re-adding base values like name, icon, quality and more.

I think the more important question here is which way of the two is heavier resource wise? And I have no clue on that if I'm honest.

WeakAuras/Prototypes.lua
Boneshockx: non-retail flavors don't have `currencyInfo.isAccountTransferable` which is what is gatekeeping any retail code from running on cata.

WeakAuras/Prototypes.lua
mrbuds: You could store common values in local variables instead of directly on states[""]

Boneshockx:
- Added `sharedStateValues` which are applied to main state and clone states.
- Removed the CopyTable from main state to clones as per result of the above
- `unitName` and `unitGUID` renamed to `characterName` and `characterGUID` which is what the WoW API calls them too.
- hide existing clones if primary checks fail (bugfix)

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
...
+          totalEarned = currencyInfo.totalEarned,
+          changed = true,
+        }
+
+        states[""] = states[""] or {}
+        local state = states[""]
+        state.show = active
+        state.value = currencyInfo.quantity
+        state.realCharacterQuantity = currencyInfo.realCharacterQuantity
+        state.capped = currencyInfo.maxQuantity and currencyInfo.maxQuantity > 0 and currencyInfo.quantity >= currencyInfo.maxQuantity
```
InfusOnWoW, line 11375: These capped calculation is duplicated above in the checkTriState calls, the code above would be much easier to read if you calculated it just once.

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
           end
         ]=]
       end
-      return ret:format(trigger.currencyId or 1)
+
+      ret = ret .. [=[
+        -- Gather currency data for account characters
+        if currencyInfo and currencyInfo.isAccountTransferable then
```
InfusOnWoW, line 11294: this whole code probably should check for "clone" somewhere

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
...
+        trigger.use_quantityEarnedThisWeek and tonumber(trigger.quantityEarnedThisWeek) or "nil",
+        trigger.use_totalEarned and tonumber(trigger.totalEarned) or "nil",
+        trigger.use_value and trigger.value_operator or "<",
+        trigger.use_realCharacterQuantity and trigger.realCharacterQuantity_operator or "<",
+        trigger.use_accountQuantity and trigger.accountQuantity_operator or "<",
+        trigger.use_realAccountQuantity and trigger.realAccountQuantity_operator or "<",
+        trigger.use_maxQuantity and trigger.maxQuantity_operator or "<",
+        trigger.use_quantityEarnedThisWeek and trigger.quantityEarnedThisWeek_operator or "<",
+        trigger.use_totalEarned and trigger.totalEarned_operator or "<",
+        trigger.use_discovered ~= nil and (trigger.use_discovered and "true" or "false") or "nil",
```
InfusOnWoW, line 11437: Using something like "trigger.use_discovered and "true" or "false" " is fine, if the check is not set at all, we can treat it as "false".

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
...
+        state.realCharacterQuantity = currencyInfo.realCharacterQuantity
+        state.capped = currencyInfo.maxQuantity and currencyInfo.maxQuantity > 0 and currencyInfo.quantity >= currencyInfo.maxQuantity
+        state.seasonCapped = currencyInfo.useTotalEarnedForMaxQty and currencyInfo.maxQuantity and currencyInfo.maxQuantity > 0 and currencyInfo.totalEarned >= currencyInfo.maxQuantity
+        state.weeklyCapped = currencyInfo.maxWeeklyQuantity and currencyInfo.maxWeeklyQuantity > 0 and currencyInfo.quantityEarnedThisweek >= currencyInfo.maxWeeklyQuantity
+        state.characterName = UnitName("player")
+        state.characterGUID = UnitGUID("player")
+        state.mainCharacter = true
+
+        if clone then
+          for i, currencyData in ipairs(accountCurrencyData) do
```
InfusOnWoW, line 11383: Transfering **everything** from a character doesn't correctly update the trigger, the clone for the character who know has 0 is still displayed.

I think this because this loop here runs over accountCurrencyData which no longer contains any data for that character, so the existing state is kept unmodified even though it should have been dropped.

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
...
+          else
+            C_CurrencyInfo.RequestCurrencyDataForAccountCharacters()
+          end
+        end
+
+        local active = true
+        local primaryCheckFailed = false
+
+        -- Check quantity-related values
+        for attribute, triggerValue in pairs(triggerQuantities) do
```
InfusOnWoW, line 11327: This should be rewritten so that checking the triggerValue happens at function creation time, not every time the trigger runs.

So the looping should be outside of the string, and the string should only include a check if the triggerValue is not nil. This makes the resulting function much smaller in the usual case that there's no or one check.

WeakAuras/Prototypes.lua
```diff
@@ -11244,9 +11288,160 @@ Private.event_prototypes = {
...
+              primaryCheckFailed = primary and true
+            end
+          end
+        end
+
+        -- Check tristate values
+        checkTristate(discovered, currencyInfo.discovered, true, true)
+        checkTristate(capped, currencyInfo.quantity >= (currencyInfo.maxQuantity or 0), currencyInfo.maxQuantity and currencyInfo.maxQuantity > 0, false)
+        checkTristate(seasonCapped, currencyInfo.totalEarned >= (currencyInfo.maxQuantity or 0), currencyInfo.useTotalEarnedForMaxQty and currencyInfo.maxQuantity and currencyInfo.maxQuantity > 0, true)
+        checkTristate(weeklyCapped, currencyInfo.quantityEarnedThisWeek >= (currencyInfo.maxWeeklyQuantity or 0), currencyInfo.maxWeeklyQuantity and currencyInfo.maxWeeklyQuantity > 0, true)
```
InfusOnWoW, line 11351: Same as above, the checking whether to include the check for discovered/capped/seasonCapped/weeklyCapped should probably be done at function creation time.

InfusOnWoW (changes requested):

WeakAuras/Prototypes.lua
InfusOnWoW: Actually you are right here, this is a tristate value, which is the stupid kind of unreadable code we sometimes create.

Boneshockx:
I still plan on updating this with the feedback from code review, but I'm just enjoying playing the game a bit too much right now :D

InfusOnWoW:
That's not a problem. I can also take a look at some of the changes I suggested, but similarly not now..

WeakAuras/GenericTrigger.lua
```diff
@@ -4936,6 +4936,16 @@ Private.GetCurrencyInfoForTrigger = function(trigger)
   end
 end
 
+Private.ExecEnv.FetchCurrencyDataFromAccountCharacters = function(currencyId, refresh)
+  Private.AccountCurrencyData = Private.AccountCurrencyData or {}
+  if Private.AccountCurrencyData[currencyId] and not refresh then
+    return Private.AccountCurrencyData[currencyId]
+  end
+
+  Private.AccountCurrencyData[currencyId] = C_CurrencyInfo.FetchCurrencyDataFromAccountCharacters(currencyId)
```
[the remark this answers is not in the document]
Boneshockx: Results after testing was a 0.05ms difference per check with caching. So I chose to remove the caching.

WeakAuras/Prototypes.lua
Boneshockx: Even if clones are disabled, the trigger will still calculate values like `accountQuantity` and `realAccountQuantity`. 
I could add a "Fetch account data" option to the trigger aside from "Clone per character", but I don't think that is necessary.

Boneshockx:
I finaly had some time to come around to this PR again. 
All checks are now built in the triggerfunc instead, and then inserted in the ret string.

InfusOnWoW:
Thanks, I'll take a look at it asap.

WeakAuras/GenericTrigger.lua
```diff
@@ -5126,6 +5126,57 @@ Private.GetCurrencyInfoForTrigger = function(trigger)
...
+
+      currencyInfo.accountCurrencyData = C_CurrencyInfo.FetchCurrencyDataFromAccountCharacters(currencyId) or {}
+
+      for _, currencyData in ipairs(currencyInfo.accountCurrencyData) do
+        if currencyData.quantity then
+          local realQuantity = currencyData.quantity
+          if currencyInfo.transferPercentage and currencyInfo.transferPercentage > 0 and currencyInfo.transferPercentage < 100 then
+            realQuantity = math.floor(realQuantity * (currencyInfo.transferPercentage/100))
+          end
+          currencyData.realCharacterQuantity = realQuantity
```
InfusOnWoW, line 5149: This setting of realCharacterQuantity for each accountCurrencyData looks wrong.

Boneshockx: The logic there was correct, but also contained unneeded sanity checks. I've simplified this bit.

WeakAuras/Prototypes.lua
```diff
@@ -11188,33 +11188,197 @@ Private.event_prototypes = {
...
+          progressType = "static",
+          accountQuantity = currencyInfo.accountQuantity,
+          realAccountQuantity = currencyInfo.realAccountQuantity,
+          description = currencyInfo.description,
+          quality = currencyInfo.quality,
+          discovered = currencyInfo.discovered,
+          transferPercentage = currencyInfo.transferPercentage,
+          quantityEarnedThisWeek = currencyInfo.quantityEarnedThisWeek,
+          totalEarned = currencyInfo.totalEarned,
+          changed = true,
```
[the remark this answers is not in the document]
Boneshockx: almost every state value that can change during a session is connected to quantity, thus changing this was rather easy. It should now only set changed to true when required.

WeakAuras/Prototypes.lua
```diff
@@ -11188,33 +11188,197 @@ Private.event_prototypes = {
...
+          table.insert(cloneChecks, ([[
+            if cloneActive and not check(currencyData.%s or 0, %s, %q) then cloneActive = false end
+          ]]):format(check.attribute, tonumber(triggerValue), operator))
+        elseif check.check_type == "tristate" and trigger["use_"..check.trigger] ~= nil then
+          local tristateValue = trigger["use_"..check.trigger] and "true" or "false"
+          table.insert(cloneChecks, ([[
+            if cloneActive and tostring(%s) ~= %q then cloneActive = false end
+          ]]):format(check.condition, tristateValue))
+        end
+      end
```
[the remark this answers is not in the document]
Boneshockx: I've reworked the creation of checkstrings in a seperate commit, see https://github.com/WeakAuras/WeakAuras2/pull/5229/commits/bb422f60363eedd7ecad107802be5560a6ab01c3

WeakAuras/Prototypes.lua
```diff
@@ -11188,33 +11188,192 @@ Private.event_prototypes = {
...
+          totalEarned = currencyInfo.totalEarned,
+        }
+
+        if active then
+          activeCharacterGUIDs[""] = true
+          states[""] = states[""] or {}
+          local state = states[""]
+
+          if state.value == currencyInfo.quantity and state.discovered == currencyInfo.discovered and state.realCharacterQuantity == currencyInfo.realCharacterQuantity then
+            state.changed = false
```
Boneshockx, line 11305: I don't think I need to set state.changed to false :|

InfusOnWoW: yeah you don't need to do anything if you aren't changing the state at all.

WeakAuras/Prototypes.lua
```diff
@@ -11188,33 +11188,192 @@ Private.event_prototypes = {
...
+        -- Remove inactive states or apply shared values
+          for cloneId, state in pairs(states) do
+            if not activeCharacterGUIDs[cloneId] then
+              state.show = false
+              state.changed = true
+              changed = true
+            else
+              for key, value in pairs(sharedStateValues) do
+                state[key] = value
+              end
```
InfusOnWoW, line 11360: since can change the state, this should, if it does, also set ```state.changed = true``` and set ```changed = true```

Boneshockx:
The code is so much more readable now compared to a couple days ago 😄
I have done as many tests as I could, but I'd like others to check it too, to be sure. There are some things I couldn't test, like season/weekly caps as I don't have any currencies with those caps.

InfusOnWoW (approved):

Issue #5253, opened 2024-07-25
Model selector input text shows the wrong model and the model viewer opens a completely different one.


Pablocgdev:
### Is there an existing issue for this?

- [X] I have searched the existing open and closed issues.

### Description

Whenever you import a WeakAuras or manually create one, the Model input text always shows the texture "spells/arcanepower_state_chest.m2" no matter what model is loaded on the aura. When you click "Choose" icon, the model viewers shows the correct one which is completely different to the one in the input text, also the Model Viewer cant properly open the correct model in the tree structure. When you select a different model text updates but after reloading the ui, it shows again the incorrect model "spells/arcanepower_state_chest.m2". 

https://github.com/user-attachments/assets/3538684a-5edf-4b99-b5ea-07035f03b04e

https://github.com/user-attachments/assets/6400e685-f695-46df-badf-3c3197ee4246

### WeakAuras Version

WeakAuras 5.15.3

### World of Warcraft Flavor

Retail (Default)

### World of Warcraft Region

EU

### Tested with only WeakAuras

I got this issue with only WeakAuras enabled

### Reproduction Steps

1. Import or create a weakauras
2. Go to "Display" tab and "Model" section inside
3. Check the current loaded model according to the WeakAura "spells/arcanepower_state_chest.m2"
4. Open the model viewer and check that the model loaded is the correct one but its not open on the file tree and also is different at the one shown in the input text "spells/arcanepower_state_chest.m2"
5. Click on any different model and reload the ui
6. Open weakauras again, its showing "spells/arcanepower_state_chest.m2" once again incorrectly.

InfusOnWoW:
I cannot reproduce the issue. But from your videos it's clear that you have other addons enabled. Please test without any other addon enabled.

And if the issue still reproduces, export the exact aura that shows the problem and add it to the ticket.

Pablocgdev:
> I cannot reproduce the issue. But from your videos it's clear that you have other addons enabled. Please test without any other addon enabled.
> 
> And if the issue still reproduces, export the exact aura that shows the problem and add it to the ticket.

Ive tested multiple times without any addons except weakauras and still happens, more proof attached.

It happens with any aura even the ones you create through the addon, this is the exact one im using on the video:
https://wago.io/JR9ZXmqbY

PR #5260, opened 2024-07-26
Tinyfix

mrbuds:
https://github.com/WeakAuras/WeakAuras2/commit/a9365a0dafd77c5b89dae10290359e81565f80c5 doesn't set `subRegion:SetClipsChildren(true)` on TWW, while trying to fix https://github.com/WeakAuras/WeakAuras2/issues/5253 it seems restoring this is the fix, but you did it for a reason so i'm unsure it's correct.


mrbuds:
Btw there is non-zero chance that SetClipsChildren work differently between 10.0.0 and 10.0.2, i didn't test on beta server

PR #5264, opened 2024-07-26
Add percentage support for GenericTrigger trigger.remaining

potat0nerd:
Dynamic remaining time for all GenericTrigger functions based on duration. Makes rotational auras easier to maintain and reduces complexity for auras tracking spells with cooldowns that varies depending on talents. This is something I use for a ton of auras with custom triggers and would appreciate if it was  available via standard triggers :)

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -211,7 +211,7 @@ function TestForMultiSelect(trigger, arg)
 end
 
 local function singleTest(arg, trigger, name, value, operator, use_exact)
-  local number = tonumber(value)
+  local number, numberType = Private.ParseNumber(value)
   if(arg.type == "tristate") then
     return TestForTriState(trigger, arg);
   elseif(arg.type == "multiselect") then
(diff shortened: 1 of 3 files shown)


WeakAuras/GenericTrigger.lua
```diff
@@ -235,7 +235,11 @@ local function singleTest(arg, trigger, name, value, operator, use_exact)
   elseif (arg.type == "string" or arg.type == "select") then
     return "(".. name .." and "..name.."==" ..(number or ("\""..(tostring(value) or "").."\""))..")";
   elseif (arg.type == "number") then
-    return "(".. name .." and "..name..(operator or "==")..(number or 0) ..")";
+    if numberType == "percent" then
+      return "(duration == 0 or ("..name.." and "..name..(operator or "==")..(number or 0).."* duration".."))";
```
InfusOnWoW, line 239: This conceptually doesn't work. You are relying on triggers hardcoding a "duration" local, which is not a given, and also means only a single duration can be used per trigger.

WeakAurasOptions/LoadOptions.lua
```diff
@@ -461,7 +462,7 @@ function OptionsPrivate.ConstructOptions(prototype, data, startorder, triggernum
           options[name..suffix] = {
             type = "input",
             width = arg.noOperator and WeakAuras.normalWidth or WeakAuras.halfWidth,
-            validate = ValidateNumeric,
+            validate = (name == "remaining" and ValidateNumericOrPercent) or ValidateNumeric,
```
InfusOnWoW, line 465: you are changing all inputs of type number here, without any regard to whether they should or shouldn't support percents

InfusOnWoW:
Writing a complex rotation helper in WA is a non-goal. WA is simply not the right tool for that, nor has the right abstractions for that.

This does somewhat useful even without the rotation helper background, so I wouldn't be against it. But you have a hardcoded "duration" name in there, which is not generic enough and I don't see a easy way to make that generic.

WeakAuras/GenericTrigger.lua
potat0nerd: ah, I see. thanks for reading though!

PR #5267, opened 2024-07-28
Cooldown Progress (Spell), Cooldown Ready: Change spell id logic

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -5186,13 +5166,7 @@ Private.event_prototypes = {
...
     loadInternalEventFunc = function(trigger, untrigger)
-      trigger.spellName = type(trigger.spellName) ~= "table" and trigger.spellName or 0;
-      local spellName;
-      if (trigger.use_exact_spellName) then
-        spellName = trigger.spellName;
-      else
-        spellName = type(trigger.spellName) == "number" and Private.ExecEnv.GetSpellName(trigger.spellName) or trigger.spellName;
-      end
+      local spellName = type(trigger.spellName) ~= "table" and trigger.spellName or 0;
       if spellName == nil then return {} end
```
mrbuds, line 5170: Now that spellName is more simple, i wonder if these data integrity tests belongs in `loadInternalEventFunc`, `loadFunc`, or `init`
I could see a validate function running before these functions

WeakAuras/Prototypes.lua
```diff
@@ -11121,7 +11024,7 @@ Private.event_prototypes = {
       if trigger.use_exact_spellName then
         spellName = trigger.spellName
       else
-        spellName = type(trigger.spellName) == "number" and Private.ExecEnv.GetSpellInfo(trigger.spellName) or trigger.spellName
+        spellName = type(trigger.spellName) == "number" and Private.ExecEnv.GetSpellInfo(trigger.spellName) or trigger
```
mrbuds, line 11027: this is wrong

mrbuds (changes requested):

WeakAuras/Prototypes.lua
InfusOnWoW: I didn't want to write to the trigger data in there, because that makes the undo/redo work @emptyrivers is doing harder. If we do such a validation, that should happen in WeakAuras.Add and do it explictly so that it can be undone.

PR #5296, opened 2024-08-07
Options: Add a Thanks button

InfusOnWoW:
This is still a bit of WIP, although I think it's ready for a review:

The next steps would be to add all the relevant people to the channel and see whether the name sanitazion needs more work or not.

With the current reactions, that's how a PR would look like:
https://github.com/InfusOnWoW/WeakAuras2/pull/8

The commit message isn't a diff from the existing data, but what the PR rejected/cleaned.


InfusOnWoW:
Wow_7BGJw4e0Nw

InfusOnWoW:
Wow_apD8quovt6

mrbuds (approved):

PR #5355, opened 2024-08-26
Hide countdown numbers from NDui

siweia:
Hide the countdown numbers from ndui


InfusOnWoW:
NDUI has a far too low user count to warrant adding code supporting it.

PR #5359, opened 2024-08-28
improvements

emptyrivers:


WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -1546,43 +1546,31 @@ local function modify(parent, region, data)
...
         width = width > 0 and width or 16
         height = height > 0 and height or 16
 
-        self:SetWidth(width)
-        self:SetHeight(height)
+        self:SetSize(width, height)
         self.currentWidth = width
         self.currentHeight = height
 
         if data.border and not data.useAnchorPerUnit then
```
emptyrivers, line 1567: i don't think this border calculation is correct but i keep confusing myself when i try to fix it. Too many symbols to chase

WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -1546,43 +1546,31 @@ local function modify(parent, region, data)
...
-          local regionLeft = SafeGetPos(region, region.GetLeft) or minX
-          local regionBottom = SafeGetPos(region, region.GetBottom) or minY
+          local regionLeft = SafeGetPos(region, region.GetLeft) or left
+          local regionBottom = SafeGetPos(region, region.GetBottom) or bottom
```
mrbuds, line 1569: Do you think we still need to use region.GetLeft/GetBottom instead of left & bottom?

WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -1546,43 +1546,31 @@ local function modify(parent, region, data)
...
       end
 
-      if numVisible > 0 then
+      if anyVisible then
         self:Show()
-        minX, maxX, minY, maxY = (minX or 0), (maxX or 0), (minY or 0), (maxY or 0)
-
-        local width, height = maxX - minX, maxY - minY
+        self:SetSize(1, 1)
+        local left, bottom, width, height = self:GetBoundsRect()
```
InfusOnWoW, line 1559: Well.... Sorf of, there's a philosophical question here, what's the bounded rect of an icon aura with a text that that is below it. GetBoundsRect() returns something different thatn the previous code.

Now, the size calculation here is imho wrong anyway, since it takes the current actual region position at the time of update into account. If the aura is animating at the point, it uses the region's dynamic position at that moment, but doesn't update later as the animation continues.

That has lead to glorious fixes like https://github.com/InfusOnWoW/WeakAuras2/commit/206e179447b636ba4e42e3d385a328633df6d7f2, and the follow up https://github.com/InfusOnWoW/WeakAuras2/commit/3ae2a2f50cbc77073e3b3eec9ef81201b69e5dc1

Using GetBoundsRect() also has that problem, that we don't know when it changes, and thus don't reliable can update it.

So I'd rather not touch that code, unless we refactor it entirely and make it have clean semantics and robust.

emptyrivers:
dropped the GetBoundsRect change

InfusOnWoW (approved):

PR #5362, opened 2024-08-29
fix

mrbuds:


WeakAurasOptions/LoadOptions.lua
```diff
@@ -99,7 +99,7 @@ local function setValue(trigger, field, value, multiEntry, entryNumber)
         trigger[field] = { trigger[field] }
       end
     end
-    if value == "" then
+    if value == "" or value == nil then
```
InfusOnWoW, line 102: Do we need to fixup triggers with holes?

mrbuds: migration to fix that added

WeakAuras/Modernize.lua
```diff
@@ -2074,6 +2074,39 @@ function Private.Modernize(data, oldSnapshot)
     end
   end
 
+  if data.internalVersion < 76 then
+    local function removeHoles(t)
+      local keys = {}
+      for key in pairs(t) do
+        table.insert(keys, key)
+      end
+      table.sort(keys)
```
emptyrivers, line 2083: avoid unnecessary sorts by checking if #keys == #t first

WeakAuras/Modernize.lua
```diff
@@ -2074,6 +2074,39 @@ function Private.Modernize(data, oldSnapshot)
...
+        end
+        return newTable
+      else
+        return t
+      end
+    end
+
+    for _, triggerData in ipairs(data.triggers) do
+      local trigger = triggerData.trigger
+      if trigger.type and trigger.event and Private.category_event_prototype[trigger.type] then
```
InfusOnWoW, line 2097: We can't use category_event_prototype/event_prototype in Modernize, because they will change as triggers evolve, but the Modernize should not change.

mrbuds: fixed both comments

InfusOnWoW (approved):

InfusOnWoW:
you have to battle it out with rivers on the spell formatting. I'd be fine with both waiting until there's more organic support for that, and for merging it.

mrbuds:
For the one time i worked with a rwf guild, one of the task i had was, for each cast bar, to define a easy to read text instead of original spell name, then edit both bossmod's lua and weakauras to sync the labels.

This change and the functions added to bossmods to edit spell labels would have save me some time

I think it's something that our resident raid auras makers would be interested in, but i have not yet see any of them signal their interest

PR #5381, opened 2024-09-03
Fix percent progress validation and parsing

NighthawkSLO:
# Description

### Current implementation

The progress settings for Set Minimum Progress and Set Maximum Progress have a special validation and parsing function which is too basic and quite broken.

Current implementation of the parser is the regex `"(%d+)%%"` - digits followed by a literal %, first such match in the string.

### Examples of broken parsing

- `-12%` parsed as 12
- `12.34%` parsed as 34
- `12.34 %` not parsed
- `abcd 12% efgh` parsed as 12
- `12%34` parsed as 12

### New implementation

~~Find and replace first instance of a literal `%` with space (See 5th example for why not empty string). The rest of the work is offloaded to tonumber (which trims leading and trailing whitespace)~~

Find the `%` at the end of the string with regex `"%% *$"`. If it is found, the parsing is offloaded to tonumber on the substring before the find.

### Fixed examples

- `-12%` parsed as -12
- `12.34%` parsed as 12.34
- `12.34 %` parsed as 12.34
- `abcd 12% efgh` not parsed
- `12%34` not parsed
- `%12` not parsed
- `12% abcd` not parsed

## Type of change

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -984,8 +984,8 @@ end
 ---@return boolean result
 function WeakAuras.ValidateNumericOrPercent(info, val)
   if val ~= nil and val ~= "" then
-    local percent = string.match(val, "(%d+)%%")
-    local number = percent and tonumber(percent) or tonumber(val)
+    local percent, count = val:gsub("%%", " ", 1)
+    local number = count == 1 and tonumber(percent) or tonumber(val)
     if(not number or number >= 2^31) then
       return false;
     end
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -984,8 +984,8 @@ end
 ---@return boolean result
 function WeakAuras.ValidateNumericOrPercent(info, val)
   if val ~= nil and val ~= "" then
-    local percent, count = val:gsub("%%", " ", 1)
-    local number = count == 1 and tonumber(percent) or tonumber(val)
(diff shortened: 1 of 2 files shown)


WeakAuras/Prototypes.lua
```diff
@@ -984,8 +984,8 @@ end
 ---@return boolean result
 function WeakAuras.ValidateNumericOrPercent(info, val)
   if val ~= nil and val ~= "" then
-    local percent = string.match(val, "(%d+)%%")
-    local number = percent and tonumber(percent) or tonumber(val)
+    local percent, count = val:gsub("%%", " ", 1)
+    local number = count == 1 and tonumber(percent) or tonumber(val)
```
InfusOnWoW, line 988: According to your information, and I haven't tested it:
"%12" would be incorrectly parsed with your code?

Meaning the right thing to do is to search for the first %, and do a tonumber on the substring before that.

NighthawkSLO: Good catch, yeah

NighthawkSLO: > Search for the first %, and do a tonumber on the substring before that.

This would fix the case of "%12" but break "12% abcd"

The handling was fixed to find the last % that is followed only by spaces and end of string - regex `"%% *$"`. Then doing tonumber on the substring before that.

InfusOnWoW (approved):

InfusOnWoW:
thanks

PR #5388, opened 2024-09-06
be more careful about adding new effective spell ids 

emptyrivers:
the watched field contains a list of spell ids, and is consumed by SendEventsForSpell, invokes ScanEventsByID for each item in the array.

But we add an item to the array every time a new effective spell id is detected, so this array can grows very large & contain many duplicates if a spell is frequently replaced, causing a cooldown triggerFunc that tracks such a spell to be run many many times for a single SPELL_UPDATE_COOLDOWN event

So, check first if the array has the spell id before adding it. This still seems slightly wrong, but naievely swapping the array for a table<number, true> introduced problems with scheduled checks that proved elusive to fix for me as a relative outsider to this bit of code.

### DIFF
--- a/WeakAuras/GenericTrigger.lua
+++ b/WeakAuras/GenericTrigger.lua
@@ -2509,7 +2509,9 @@ do
               -- a) We are already tracking it, only add the wathed spell ids
               -- b) We aren't tracking it yet add it
               if self.data[newEffectiveSpellId] then
-                tinsert(self.data[newEffectiveSpellId].watched, userSpellId)
+                if not tContains(self.data[newEffectiveSpellId].watched, userSpellId) then
+                  tinsert(self.data[newEffectiveSpellId].watched, userSpellId)
+                end
               else
                 self:AddEffectiveSpellId(newEffectiveSpellId, userSpellId)
               end


WeakAuras/GenericTrigger.lua
```diff
@@ -2509,7 +2509,9 @@ do
               -- a) We are already tracking it, only add the wathed spell ids
               -- b) We aren't tracking it yet add it
               if self.data[newEffectiveSpellId] then
-                tinsert(self.data[newEffectiveSpellId].watched, userSpellId)
+                if not tContains(self.data[newEffectiveSpellId].watched, userSpellId) then
+                  tinsert(self.data[newEffectiveSpellId].watched, userSpellId)
+                end
```
emptyrivers, line 2514: I'm not really happy with this fix, though it does massively improve things - should this not be a set-like table instead of an array? But when I tried that, i got problems with scheduled timers, and when i tried to fix *that* i ran into a situation where the aura would vanish every time the base spell was replaced. So, I decided to instead push this up in the hopes @InfusOnWoW would understand the code better and decide on a better fix.

InfusOnWoW: I did at one point understand the code. Though it was a moment of clairvoyance.

emptyrivers:
#5389 is a better fix

PR #5410, opened 2024-09-13
Model picker search with multiple words

walruscow:
# Description

This change allows multiple filters to be applied in model picker search, by separating on spaces.

Before, it was difficult to find specific things like "warlock fel effects" or "runes with multiple pieces". Now, these filters can be accomplished by e.g. searching "warlock fel" or "rune 0". In my experience, this makes it a lot easier to find models I'm interested in.

- The filter string is split on space to form a filter list
  - This is safe because no model paths include a space today
- Must match all filters in the list
- Only create the filter list once

## Type of change

## How Has This Been Tested

- [ ] Searched in the model picker without spaces
- [ ] Searched in the model picker with spaces
- [ ] Paid attention to the search speed, it seems similar to before

## Checklist

## Screenshot
image

### DIFF
--- a/WeakAurasOptions/OptionsFrames/ModelPicker.lua
+++ b/WeakAurasOptions/OptionsFrames/ModelPicker.lua
@@ -52,7 +52,7 @@ local function ConstructModelPicker(frame)
       end
       return filterList
     end
-    
+  
     local function MatchesAllFilters(text, filters)
       for _, f in ipairs(filters) do
         if not text:find(f, 1, true) then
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
We recently added some more advanced filtering to the aura search: https://github.com/WeakAuras/WeakAuras2/commit/7569c2f73ba6b9f1f42b5dad1e1f0229985685d4 That adds "|" and " or " as or filters. 

Your change adds "and" filtering which conflicts with what we just added to the general aura search. 

These two searches shouldn't behave in widely different ways.

walruscow:
That's ok, feel free to take/change it as you like or not :)

Stanzilla:
> That's ok, feel free to take/change it as you like or not :)

Thank you!

PR #5436, opened 2024-09-23
consider wagoID as metadata

emptyrivers:


WeakAuras/Types.lua
```diff
@@ -3531,7 +3531,8 @@ Private.update_categories = {
       "url",
       "desc",
       "version",
-      "semver"
+      "semver",
+      "wagoID", -- i don't *love* that we're so closely tied to wago, but eh
```
emptyrivers, line 3535: without this, uploading to wago and then importing again causes a "change" in the display tab, which doesn't seem correct to me

InfusOnWoW (approved):

PR #5455, opened 2024-10-02
Use fonts from official sources

Nightwarden24:
# Description

Whenever possible, add/update fonts from official sources rather than third-party ones (use a repository on github, website of the studio that developed them, etc.)

Fira Sans
Type foundry: bBox Type GmbH
URL: https://bboxtype.com/
Font page: https://bboxtype.com/typefaces/FiraSans

PT Sans
Type foundry: Paratype
URL: https://www.paratype.com/
Font page: https://www.paratype.com/fonts/pt/pt-sans

Notes:

1. Fonts updated to latest versions (4.301, 2.006)

2. Font names have not been changed from Black to Heavy to ensure compatibility.
Black and Heavy are common names for the same font weight (900). 
See links:
https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass
https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping


InfusOnWoW (approved):

InfusOnWoW:
Thanks, also for adding PT Sans License.

PR #5456, opened 2024-10-02
fix

mrbuds:


WeakAuras/GenericTrigger.lua
```diff
@@ -876,39 +876,61 @@ function WeakAuras.ScanEvents(event, arg1, arg2, ...)
...
+    local unit_list = loaded_unit_events[unit]
+    if unit_list then
+      local event_list = unit_list[event]
+      if event_list then
+        for id, triggers in pairs(event_list) do
+          Private.StartProfileAura(id);
+          Private.ActivateAuraEnvironment(id);
+          local updateTriggerState = false;
+          for triggernum, data in pairs(triggers) do
+            local skipThisUnit = false
```
InfusOnWoW, line 907: I took a look at the code and I think it fits better inside RunTriggerFunc next to the already existing code for data.trigger.unit.

PR #5474, opened 2024-10-08
Smallpatches

InfusOnWoW:


WeakAuras/RegionTypes/DynamicGroup.lua
```diff
@@ -457,7 +457,7 @@ local centeredIndexerStart = {
     if maxIndex % 2 == 1 then
       return maxIndex
     else
-     return maxIndex - 1
+     return maxIndex > 0 and maxIndex - 1 or nil
```
emptyrivers, line 460: I would expect CLR to need the same surgery

emptyrivers (changes requested):

PR #5475, opened 2024-10-10
finally remove LibDeflate hard commit

emptyrivers:
since years ago the test data was finally excluded from packaging which was the whole reason we were doing that


InfusOnWoW:
Hmm, there's still a bit of unnecessary files in there, although I guess directly using upstream is probably worth that minor cost.

mrbuds (approved):

PR #5493, opened 2024-10-23
try moving history updates to background task

emptyrivers:
i've been meaning to try this for a while, i can't really tell if it improves anything or not


WeakAurasOptions/OptionsFrames/Update.lua
```diff
@@ -1807,6 +1829,14 @@ local methods = {
       OptionsPrivate.ClearPicks()
       WeakAuras.PickDisplay(pendingPickData.id, pendingPickData.tabToShow)
     end
+    OptionsPrivate.Private.Threads:Add("history_update", coroutine.create(function()
```
emptyrivers, line 1832: this is the meat of the change, I've a suspicion that doing the work to create the archive is weirdly taxing to the game (possibly due to extreme GC churn?).

InfusOnWoW (approved):

PR #5495, opened 2024-10-24
Include area ID in tooltip for zone identifier

m33shoq:
# Description

Currently WeakAuras allows to use areaID for zone identifier option, but finding the correct areaID is a manual process that users must handle themselves. Looking for IDs on wago.tools is not a great experience so why not just try to find the areaID ingame?

Approach I took for this PR is brute but it achieves the goal. It doesn't result in any lag (at least based on testing on my PC), so as long as `get_zoneId_list` only called when hovering field for zone identifier I think it is acceptable for such brute way.

Not all subzones have a corresponding entry in the AreaTable that matches the subzone name. In these cases we fallback to "not found". Maybe better approach would be to remove `Current Area` part from tooltip when there is no areaID found. And also do we need to change text at the bottom of the tooltip explaining where to find areaIDs?

## How Has This Been Tested
I flied over Dornogal, Isle of Dorn and Stormwind mouseovering zone identifier load condition. Works as I would expect it to.

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -203,7 +203,17 @@ local function get_zoneId_list()
                                          L["Parent Zone"], parentmap_name, currentmap_info.parentMapID)
   end
 
-  return ("%s|cffffd200%s|r\n%s: %d\n\n%s%s|cffffd200%s|r\n%s: i%d\n\n%s"):format(
+  local minimapZoneText = GetMinimapZoneText()
+  local currentAreaID = L["not found"]
+  for id=1,30000 do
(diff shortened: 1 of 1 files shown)


InfusOnWoW:
I would rather not merge a brute-force enumeration of all ids. We do have that kind of loop for buffs/debuffs and it's a constant ableit low amount of maintenance, which we always miss until someone points out that it is broken.

That cost isn't that high, but annoying, and on the other hand I don't see the benefit being relevant. The area option is there as a fallback when all the other options don't work. The other options are the preferred way to check the zone.

PR #5506, opened 2024-10-29
add support for "stealable glow"

mrbuds:


WeakAuras/Types.lua
```diff
@@ -3222,6 +3222,7 @@ LSM:Register("statusbar", "Stripes", [[Interface\AddOns\WeakAuras\Media\Textures
 LSM:Register("statusbar", "Thick Stripes", [[Interface\AddOns\WeakAuras\Media\Textures\Statusbar_Stripes_Thick]])
 LSM:Register("statusbar", "Thin Stripes", [[Interface\AddOns\WeakAuras\Media\Textures\Statusbar_Stripes_Thin]])
 LSM:Register("border", "Drop Shadow", [[Interface\AddOns\WeakAuras\Media\Textures\Border_DropShadow]])
+LSM:Register("border", "Stealable Glow", [[Interface\TargetingFrame\UI-TargetingFrame-Stealable]])
```
InfusOnWoW, line 3225: this doesn't appear to work. That is the resulting border looks wrong.

InfusOnWoW (commented):
I'm fine with adding this, though it's a rather minimal prefernce for adding it.

WeakAuras/Types.lua
mrbuds: ho right, i added it as a test and forget to remove later

mrbuds:
> I'm fine with adding this, though it's a rather minimal prefernce for adding it.

I'm not convinced it's good enough & generic to be added, as it look bad on anything that is not square & tiny

mrbuds:
i closed https://github.com/Stanzilla/LibCustomGlow/pull/28

PR #5525, opened 2024-11-07
Smallpatches

InfusOnWoW:


emptyrivers (approved):

emptyrivers:
of course it was only after i clicked merge that i had the thought that maybe we should teach BT2 to have an option to explicitly filter for/against "durationless" auras.

I suppose we'll see if that's something people actually want soon enough. But also perhaps `total duration == 0` is sufficient anyways.

PR #5549, opened 2024-11-23
Smallpatches

InfusOnWoW:


WeakAurasOptions/ConditionOptions.lua
```diff
@@ -422,7 +420,6 @@ local function addControlsForChange(args, order, data, conditionVariable, totalA
           auraData[conditionVariable][conditionIndex].changes[reference.changeIndex].value[property][2] = g;
           auraData[conditionVariable][conditionIndex].changes[reference.changeIndex].value[property][3] = b;
           auraData[conditionVariable][conditionIndex].changes[reference.changeIndex].value[property][4] = a;
-          WeakAuras.Add(auraData);
```
mrbuds, line 425: shouldn't this turn on a boolean, then after the "for" loop apply WeakAuras.Add if the boolean is true?

mrbuds: ignore previous comment, auraData is set on each loop

InfusOnWoW: Oh you are right taht is a wrong change there.

emptyrivers (approved):

PR #5561, opened 2024-11-30
Smallpatches

InfusOnWoW:


mrbuds:
Trying to add any subregion to an icon (on classic era)

```
2x ...aceWeakAuras/RegionTypes/RegionPrototype.lua:51: attempt to index field '?' (a nil value)
[string "@WeakAuras/RegionTypes/RegionPrototype.lua"]:51: in function `GetAnchorsForData'
[string "@WeakAurasOptions/SubRegionOptions/Glow.lua"]:13: in function `create'
[string "@WeakAurasOptions/DisplayOptions.lua"]:105: in function `?'
[string "@WeakAurasOptions/OptionsFrames/OptionsFrame.lua"]:1094: in function `EnsureOptions'
[string "@WeakAurasOptions/OptionsFrames/OptionsFrame.lua"]:1071: in function `UpdateOptions'
[string "@WeakAurasOptions/OptionsFrames/OptionsFrame.lua"]:1048: in function `ClearAndUpdateOptions'
[string "@WeakAurasOptions/WeakAurasOptions.lua"]:908: in function `ClearAndUpdateOptions'
[string "@WeakAurasOptions/DisplayOptions.lua"]:34: in function <WeakAurasOptions/DisplayOptions.lua:20>
[string "@WeakAurasOptions/DisplayOptions.lua"]:65: in function <WeakAurasOptions/DisplayOptions.lua:64>
[string "=[C]"]: ?
[string "@BetterBags/libs/AceConfig-3.0-3/AceConfigDialog-3.0-87/AceConfigDialog-3.0.lua"]:45: in function <...nfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua:43>
[string "@BetterBags/libs/AceConfig-3.0-3/AceConfigDialog-3.0-87/AceConfigDialog-3.0.lua"]:839: in function <...nfig-3.0/AceConfigDialog-3.0/AceConfigDialog-3.0.lua:661>
```

WeakAurasOptions/WeakAurasOptions.toc
```diff
@@ -38,6 +38,10 @@ SubRegionOptions\Border.lua
 SubRegionOptions\Glow.lua
 SubRegionOptions\Tick.lua
 SubRegionOptions\Model.lua
+SubRegionOptions\StopMotion.lua
+SubRegionOptions\Texture.lua
+SubRegionOptions\CircularProgressTexture.lua
+SubRegionOptions\LinearProgressTexture.lua
```
mrbuds, line 44: This needs to be added to WeakAurasOptions_Vanilla.toc and WeakAurasOptions_Cata.toc

WeakAurasOptions/SubRegionOptions/LinearProgressTexture.lua
```diff
@@ -0,0 +1,268 @@
...
+      width = WeakAuras.normalWidth,
+      control = "WeakAurasTwoColumnDropdown",
+      order = 20,
+      values = areaAnchors,
+      hidden = function()
+        return data.anchor_mode == "point"
+      end
+    },
+
+    self_point = {
```
mrbuds, line 210: could use a spacer to align anchors & size on same lines
image

btw, something for an other PR, it would be nice if self_point dropdown used WeakAurasAnchorButtons widget, and a new one would have to be created for anchor_point

WeakAurasOptions/SubRegionOptions/LinearProgressTexture.lua
```diff
@@ -0,0 +1,268 @@
...
+      min = 0,
+      softMax = 200,
+      step = 1,
+      order = 23,
+      hidden = function()
+        return data.anchor_mode == "area"
+      end
+    },
+
+    height = {
```
mrbuds, line 248: model subelement may need an update to match textures/stopmotion subelements size & positioning options

WeakAurasOptions/SubRegionOptions/LinearProgressTexture.lua
mrbuds: Also could you add x/y offsets?

InfusOnWoW: Yeah they make sense for point anchoring.

WeakAurasOptions/SubRegionOptions/LinearProgressTexture.lua
InfusOnWoW: I did look at that briefly and I agree, but it seems this requires a bit of tinkering to get right, I'll do that some other day.

InfusOnWoW: I think there are 3 cases for sub element position settings:
Area + Point: (Linear/Circle) Texture, StopMotion, **Border!, Glow!, Model!**
Only Point anchoring: Text
Special: Ticks

There's not a lot of value in having Border and Glow being abled to anchor at a point with a specific size atm, but I think they are in same category as e.g. linear progres textures. I'll put changing those 3 on my todo-list but I don't think it's that important for now.

While I made Ticks being abled to provide anchors for other sub elements, and thus the new sub element types can anchor to ticks, I didn't made these new sub elements themselves provide anchors. That's because doing that requires preventing cycles in the anchoring, which are currently simply impossible.

WeakAuras/WeakAuras_Cata.toc
```diff
@@ -84,6 +89,10 @@ SubRegionTypes\Border.lua
 SubRegionTypes\Glow.lua
 SubRegionTypes\Tick.lua
 SubRegionTypes\Model.lua
+SubRegionTypes\StopMotion.lua
+SubRegionTypes\Texture.lua
+SubRegionTypes\CircularTexture.lua
```
mrbuds, line 94: it's CircularProgressTexture.lua

WeakAuras/Prototypes.lua
```diff
@@ -1206,6 +1206,16 @@ function WeakAuras.GetEffectiveAttackPower()
...
 
+--- @type fun(): number
+function WeakAuras.GetEffectiveSpellPower()
+  -- Straight from the PaperDoll
+  local spellPower = GetSpellBonusDamage(2)
+  for i = 3, MAX_SPELL_SCHOOLS do
+    spellPower = min(spellPower, GetSpellBonusDamage(i))
+  end
+  return spellPower
+end
```
mrbuds, line 1217: It doesn't look right to return minimal value, if i have 400 spell power and 600 fire spell power i expect it to return 600
Changed to look more similar to `WeakAuras.GetCritChance` and work on classic_era

```suggestion
--- @type fun(): number
function WeakAuras.GetEffectiveSpellPower()
  -- Straight from the PaperDoll
  local spellPower = 0
  for i = 2, MAX_SPELL_SCHOOLS or 7 do
    spellPower = max(spellPower, GetSpellBonusDamage(i))
  end
  return spellPower
end
```

WeakAuras/Prototypes.lua
```diff
@@ -10367,6 +10377,20 @@ Private.event_prototypes = {
...
+        type = "number",
+        init = "WeakAuras.GetEffectiveSpellPower()",
+        store = true,
+        conditionType = "number",
+        multiEntry = {
+          operator = "and",
+          limit = 2
+        },
+        enable = WeakAuras.IsCataClassic(),
+        hidden = not WeakAuras.IsCataClassic(),
```
mrbuds, line 10392:
```suggestion
        enable = WeakAuras.IsClassicOrCata(),
        hidden = not WeakAuras.IsClassicOrCata(),
```

WeakAurasOptions/OptionsFrames/MoverSizer.lua
```diff
@@ -49,6 +49,10 @@ local function moveOnePxl(direction)
     local physicalWidth, physicalHeight = GetPhysicalScreenSize();
     local oneX = GetScreenWidth() / physicalWidth
     local oneY = GetScreenHeight() / physicalHeight
+    if IsShiftKeyDown() then
```
mrbuds, line 52: Problem with this feature is that when we press shift, it change moversizer alpha to zero, so we click on the void
Maybe that behavior needs to be changed when cursor is close to the buttons or something?

WeakAuras/Prototypes.lua
InfusOnWoW: Hmm I checked the ui sources for what blizzard does, and to me the code looked like it did a min, though max should be much more useful in pratice. So yeah I'll change that.

WeakAurasOptions/OptionsFrames/MoverSizer.lua
InfusOnWoW: Yeah, I tested the change in the last few weeks and I did find it annoying once already, so I think I'll drop the change.

PR #5586, opened 2024-12-17
Add Money Formatting Option and Add Player Money to Currency Trigger

Boneshockx:
## Currency Trigger: Add Player Money
I'm not sure what makes more sense for players, finding this in the currency trigger or as its own trigger.
Adding it to the currency trigger also means I had to add additional checks and hide everything related to currencies when player money was selected. From a coding PoV, this is ugly, and I would prefer having Player Money as a seperate trigger.

## Money Formatting Option
Accepts a number:wowmoney
Has two options: 
- Gold format (big number with a "disable" option)
- Coin Precision
image
image

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -11272,6 +11272,71 @@ Private.event_prototypes = {
     automaticrequired = true,
     progressType = "none"
   },
+  ["Money"] = {
+    type = "unit",
+    statesParameter = "one",
+    progressType = "none",
+    automaticrequired = true,
+    events = {
+      ["events"] = {"PLAYER_MONEY"}
+    },
+    internal_events = {"WA_DELAYED_PLAYER_ENTERING_WORLD"},
+    force_events = "WA_DELAYED_PLAYER_ENTERING_WORLD",
+    name = L["Player Money"],
+    init = function()
+      return [=[
+        local money = GetMoney()
+        local gold = floor(money / 1e4)
+        local silver = floor(money / 100 % 100)
+        local copper = money % 100
+      ]=]
+    end,
+    args = {
+      {
+        name = "money",
+        init = "money",
+        type = "number",
+        display = L["Money"],
+        store = true,
+        hidden = true,
+        test = "true",
+      },
+      {
+        name = "gold",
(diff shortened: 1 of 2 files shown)


InfusOnWoW:
Looking at it, I think I have a preference to it being a separate trigger.

WeakAuras/Prototypes.lua
```diff
@@ -11290,6 +11291,55 @@ Private.event_prototypes = {
...
+          return([[
+            if not (%s %s %s) then
+              active = false
+            end
+          ]]):format(property, operator or "<", tonumber(value) or 0)
+        end
+
+        table.insert(moneyChecks, createMoneyCheck("gold", trigger.gold_operator, trigger.gold, trigger.use_gold))
+        table.insert(moneyChecks, createMoneyCheck("silver", trigger.silver_operator, trigger.silver, trigger.use_silver))
+        table.insert(moneyChecks, createMoneyCheck("copper", trigger.copper_operator, trigger.copper, trigger.use_copper))
```
InfusOnWoW, line 11307: I don't think these are that useful, except maybe the gold one.

Boneshockx:
Reworked this into a seperate Player Money trigger.
I do think having `%copper` and `%silver` available is a plus for players who want to display this in their own way (without the format option). Having the additional checks/conditions for these coins is less useful, but there are probably players out there that require more precise denominations, for example: lower level players and roleplaying. Keeping these options avoids the need to reintroduce them later.

InfusOnWoW:
Well I think "> 1 silver" not being true when the user has 1g 0s 0c is very strange and I can't think of anyway that's useful. So nah, I'd rather not have them and wait for someone to complain. And then we can figure out what they actually want.

Arguably the cleaner solution would be if the "money" condition would allow configuring the "unit" of the right hand side via some method. E.g. a dropdown, or what we for time, where you can (iirc) input 2:00 for 120s.

Now, personally I think that's far beyond what is needed atm, even on classic "gold" is the "dollar" and "silver/copper" being the cent values. So having a gold check is imho 90% of the solution, with the rest to be figured out when someone has a problem.

InfusOnWoW (approved):

PR #5596, opened 2024-12-26
add sourceGUID to chat msg trigger state

emptyrivers:
inspired by [this post in discord](https://discord.com/channels/172440238717665280/1321645681580052554/1321645681580052554).
We could & maybe should add a shim for Ambiguate in our format options, but since the source GUID is, yknow, *right there* it seems silly to not make it available

### DIFF
--- a/WeakAuras/Prototypes.lua
+++ b/WeakAuras/Prototypes.lua
@@ -8009,6 +8009,32 @@ Private.event_prototypes = {
         conditionType = "string",
       },
       {
+        -- flags
+      },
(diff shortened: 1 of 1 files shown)


InfusOnWoW (approved):

mrbuds (approved):
Kinda related, in recent days someone asked how to filter by channel base name so they could only show Trade chat, whatever index it has. Problem is that this arg has a localized value and i couldn't find a proper way to make it work across different languages.

PR #5606, opened 2025-01-04
Smallpatches

InfusOnWoW:


WeakAurasOptions/RegionOptions/Empty.lua
```diff
@@ -0,0 +1,57 @@
...
+local function modifyThumbnail(parent, frame, data)
+
+end
+
+-- Register new region type options with WeakAuras
+OptionsPrivate.registerRegions = OptionsPrivate.registerRegions or {}
+table.insert(OptionsPrivate.registerRegions, function()
+  OptionsPrivate.Private.RegisterRegionOptions("empty", createOptions, createThumbnail, L["Empty Region"],
+                                               createThumbnail, modifyThumbnail,
+                                               L["Shows nothing"]);
```
mrbuds, line 56: Region's name and description seems a bit off to me, as after some subelement(s) are added it's not empty anymore. Also the description refer to the word "region" which is new in the UI.

What about "Empty base Aura" and "Shows only Sub Elements"

While at i wonder if it's time to make a "Group" section for Group & Dynamic Group and an "Aura Type" section?

mrbuds (approved):

PR #5616, opened 2025-01-13
Smallpatches

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -2431,7 +2431,7 @@ Private.event_prototypes = {
         store = true,
         conditionType = "select",
         enable = function(trigger)
-          return WeakAuras.IsCataOrRetail() and (trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party")
+          return trigger.unit == "group" or trigger.unit == "raid" or trigger.unit == "party"
```
mrbuds, line 2434: I think we can lift this unit restriction to make it work with player & target

mrbuds (commented):
I was looking into making the "Spec Role" load option work on classic_era, but we use the value from LibSpecialization for it on cata & retail, and it doesnt use UnitGroupRolesAssigned

mrbuds (approved):

WeakAuras/Prototypes.lua
InfusOnWoW: I agree, I'll put that into the next PR.

PR #5644, opened 2025-01-30
Smallpatches

InfusOnWoW:


WeakAuras/Prototypes.lua
```diff
@@ -6161,6 +6161,16 @@ Private.event_prototypes = {
...
+      {
+        name = "itemId",
+        display = L["ItemId"],
+        hidden = true,
+        init = "item",
+        test = "true",
+        store = true,
+        conditionType = "number",
+        operator_types = "only_equal",
+      },
```
mrbuds, line 6173: Most item triggers use property `hasItemID = true,`  that is only use for tooltips, i wonder if that property can be use for refactor more of these trigger filters / conditions

mrbuds (approved):

WeakAurasOptions/OptionsFrames/OptionsFrame.lua
```diff
@@ -518,8 +518,7 @@ function OptionsPrivate.CreateFrame()
   if OptionsPrivate.changelog then
     local changelog
     if OptionsPrivate.changelog.highlightText then
-      changelog = L["Highlights"] .. "\n" .. OptionsPrivate.changelog.highlightText .. "\n"
-      .. L["Commits"] .. "\n" ..OptionsPrivate.changelog.commitText
+      changelog = L["Highlights"] .. "\n" .. OptionsPrivate.changelog.highlightText
```
emptyrivers, line 521: i want to note that because the changelog is a lot more accessible now, we shouldn't be lazy about writing them.

not that we *have* been lazy recently. just sayin'

emptyrivers (approved):

PR #5683, opened 2025-02-20
deduplicate localization phrases

### LINKED ISSUE #5682: Reuse translation strings
There are many many many examples in the curseforge localization table of needing to translate strings that have already been translated. It is sloppy to not use the same translation multiple times if necessary.

Reuse strings instead of creating them 2, 3, or even 4 times

No good alternative

**Additional context**
Just one example when I filtered for the word "source"

Image

emptyrivers:
Fixes #5682

### DIFF
--- a/babelfish.lua
+++ b/babelfish.lua
@@ -40,23 +40,27 @@ local function parseFile(filename)
 end
 
 -- extract data from specified lua files
+local dedupe = {}
 for _, namespace in ipairs(ordered) do
     print(namespace)
     local ns_file = assert(io.open(namespace .. ".lua", "w"), "Error opening file")
+    local sorted = {}
     for _, file in ipairs(fileList[namespace]) do
+        local count = 0
         local strings = parseFile(file)
 
-        local sorted = {}
         for k in next, strings do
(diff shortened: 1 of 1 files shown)


babelfish.lua
```diff
@@ -40,23 +40,27 @@ local function parseFile(filename)
...
+                dedupe[k] = true
+                count = count + 1
             end
         end
-        print("  (" .. #sorted .. ") " .. file)
+        print("  (" .. count .. ") " .. file)
+    end
+    table.sort(sorted)
+    if #sorted > 0 then
+        ns_file:write("L[\"" .. (table.concat(sorted, "\"] = true\nL[\"") or "") .. "\"] = true\n")
```
emptyrivers, line 63: i suppose we could also write each line one at a time..

InfusOnWoW: yeah that looks too clever for me.

You could also simplify that by making sorted contain L=["XXX"] = true, instead of just XXX. Then the write becomes a simple table.concat.

InfusOnWoW (approved):

PR #5833, opened 2025-04-23
fixes

mrbuds:


WeakAuras/Types.lua
```diff
@@ -2499,7 +2499,17 @@ Private.texture_types = {
...
+if C_Texture and C_Texture.GetAtlasElements then
+  Private.texture_types["Blizzard Atlas"] = function()
+    local atlasList = C_Texture.GetAtlasElements()
+    table.sort(atlasList)
+    local atlasTable = {}
+    for _, atlas in ipairs(atlasList) do
+      atlasTable[atlas] = atlas
+    end
+    return atlasTable
+  end
```
InfusOnWoW, line 2511: I add something to this branch that checks if the change was ported to cata/classic, e.g.
```
if WeakAuras.buildType == "dev" and Private.AtlasList then
  WeakAuras.prettyPrint("Private.AtlasList can be removed now.")
end
```

PR #5841, opened 2025-04-28
Smallpatches

InfusOnWoW:


WeakAurasOptions/ActionOptions.lua
```diff
@@ -82,13 +82,25 @@ function OptionsPrivate.GetActionOptions(data)
...
       },
       init_do_custom = {
         type = "toggle",
-        name = L["Custom"],
-        order = 0.011,
+        name = L["Custom Init"],
+        order = 0.2,
+        width = WeakAuras.doubleWidth
+      },
+      init_do_custom_load = {
```
emptyrivers, line 94: custom code needs to be shown in import code review

WeakAurasOptions/ActionOptions.lua
```diff
@@ -82,13 +82,25 @@ function OptionsPrivate.GetActionOptions(data)
...
+        type = "toggle",
+        name = L["Custom Load"],
+        order = 0.3,
+        width = WeakAuras.doubleWidth
+      },
+      init_do_custom_unload = {
```
emptyrivers, line 100: same as above. also, cc @oratory, there's new custom code to decode in wago.

emptyrivers (changes requested):

Stanzilla:
I actually don''t like/think we should have the WA category in the addon list. it just looks weird with the group in there as well. you could argue that I could then also stick all the Wago stuff in there and people maybe their helper addons but I don''t think it improves the visual clarity at all

mrbuds (approved):
tiny typo on 1st commit message, otherwise good

PR #5852, opened 2025-05-04
Smallpatches

InfusOnWoW:


WeakAuras/BossMods.lua
```diff
@@ -1523,11 +1523,16 @@ Private.event_prototypes["Boss Mod Announce"] = {
...
+        return Private.ExecEnv.CreateTriggerCounter(input)
+      end,
       conditionTest = function(state, needle, op, preamble)
-        return preamble and preamble:Check(state.count)
+        preamble:SetCount(tonumber(state.count) or 0)
+        return preamble:Match()
       end,
       store = true,
-      conditionType = "number",
+      conditionType = "string",
```
mrbuds, line 1534: It should be the same type used for triggers like Combat Log's count, aka "number"

image

InfusOnWoW: hmm i made it work with the same syntax as the trigger, so you can use e.g. "2,4"

mrbuds: it's odd to have them work differently, but i'll let it be

PR #5891, opened 2025-06-03
fix

mrbuds:


mrbuds:
It would indeed be sane to not sort something you don't care about

PR #5929, opened 2025-07-06
fix

mrbuds:


WeakAuras/Types.lua
```diff
@@ -2582,6 +2582,18 @@ elseif WeakAuras.IsCataClassic() then
       runes[tostring(v)] = nil
     end
   end
+elseif WeakAuras.IsMists() then
+  local alerts = Private.texture_types["Blizzard Alerts"]
```
InfusOnWoW, line 2586: I'd prefer a list of MOP available ids, instead of "Retail - MOP", since retail is going to occasional change.

But I'm fine with this if you want.

mrbuds: done

PR #6004, opened 2025-08-27
fix

mrbuds:


WeakAuras/WeakAuras.lua
```diff
@@ -3545,24 +3545,19 @@ function Private.ReleaseClone(id, cloneId, regionType)
...
-        C_VoiceChat.SpeakText(
-          validVoice and voice or next(Private.tts_voices) or 0,
-          message,
-          1,
-          C_TTSSettings and C_TTSSettings.GetSpeechRate() or 0,
-          C_TTSSettings and C_TTSSettings.GetSpeechVolume() or 100
-        );
+        local voice = TextToSpeech_GetSelectedVoice(Enum.TtsVoiceType.Standard)
+        if not voice then return end
+        TextToSpeech_Speak(message, voice)
```
InfusOnWoW, line 3560: We can try that, but if this behaves too differently from what people know, we'll need to revert this.

PR #6016, opened 2025-09-06
Smallpatches

InfusOnWoW:


mrbuds:
for the model subregion foreground mode change, is it possible to test if it was fixed on both mop & anniversary
if it was not updated yet on this version, is there a test we can do to check if api support it, then gracefully add the support?

InfusOnWoW:
> for the model subregion foreground mode change, is it possible to test if it was fixed on both mop & anniversary if it was not updated yet on this version, is there a test we can do to check if api support it, then gracefully add the support?

I did test it, the clipping works there too.

PR #6046, opened 2025-09-26
Smallpatches

InfusOnWoW:


WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -33,6 +35,15 @@ local function getAuraMatchesList(name)
         end
       end
     end
+
+    if showSpellIdRecommendation then
+      if numMatches == 1 then
+        descText = descText .. "\n" .. L["WeakAuras recommend using spell ids instead of names. Click to replace this name with the spell id"]
```
emptyrivers, line 41: recommend**s**

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -33,6 +35,15 @@ local function getAuraMatchesList(name)
...
+      else
+        descText = descText .. "\n" .. L["WeakAuras recommend using spell ids instead of names."]
```
emptyrivers, line 43: same as above

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -160,6 +173,23 @@ local function CreateNameOptions(aura_options, data, trigger, size, isExactSpell
         end
         return icon and tostring(icon) or "", 18, 18
       end
+
+      aura_options[iconOption].func = function()
```
emptyrivers, line 177: i don't love that this is such a hidden feature, though i can't think of a way to improve that without tearing holes in how we layout these options.

emptyrivers:
A wrinkle with tracking totems by spell id: restoration shaman has cloudburst totem, which replaces the spell healing stream totem. But the totem trigger doesn't seem to be smart enough to figure out that if it's configured to track healing stream totem (spell id 5394) which is overridden by cloudburst (157153), then a totem with spell id 157153 should match. Also, multiple spell ids in one trigger is not supported, so the only solution is to have near duplicate triggers.

emptyrivers:
A similar thing exists for demonology warlock, where summon vilefiend creates a totem, and there is a choice node in the spec tree which overrides vilefiend into 2 different spells.

InfusOnWoW:
The totem trigger can now handle overriden spells.

mrbuds:
While testing https://github.com/WeakAuras/WeakAuras2/pull/6046/commits/05cb084cf68e75563916f90f7210e7a5d7069b08 i see custom text function in conditions is not handled the same for chat message (which show it's own code editbox) and for subtext's text editing (which use custom func defined in display tab), this is not news.. but i don't like it

WeakAurasOptions/BuffTrigger2.lua
```diff
@@ -33,7 +48,44 @@ local function getAuraMatchesList(name)
...
+        addSpellToDesc(id)
+      end
+    end
+
+    local bestSuggestion
+    if #playerSpells == 1 then
+      bestSuggestion = playerSpells[1]
+    elseif #playerSpells == 0 and #otherSpells == 1 then
+      bestSuggestion = otherSpells[1]
+    end
```
mrbuds, line 78: Is there any reason to make it work only when 1 spellId match instead of always returning the first of the list?

As an example, i type "power word: shield", it give me a list of ids, any of them would work

InfusOnWoW:
Well, I think it's good that %c in the custom chat message has its own custom function, so that you don't have to reuse the same one. (It's also even per condition.)

mrbuds:
> Well, I think it's good that %c in the custom chat message has its own custom function, so that you don't have to reuse the same one. (It's also even per condition.)

I think it's good too, maybe conditions that edit a text should work the same way
It's just food for though, not a request for changes

mrbuds (approved):

PR #6143, opened 2026-01-28
Update for Midnight compatibility

shrugal:
# Description

This is somewhat of a quick-fix for the Midnight changes. Basically all places that caused problems in combat or during M+ dungeons have been wrapped with the new secret value access APIs, and not being able to access values is treated as them being nil or false in most cases.

This does result in some auras starting to become a bit janky when the restrictions kick in (e.g. buffs/debuffs not updating anymore), so I also added an option to unload auras that would cause problems with addon restrictions.

I still get an error message when logging in about calling `coroutine.resume` and that somehow being prohibited, but I couldn't figure out a solution for this. It doesn't seem to cause any real problems though.

Changes in detail:

- Update Retail ToC version
- Wrap places where potentially secret values are read in new secret value access APIs
- Wrap calls to `WeakAuras.GetUnitFrame` in `pcall`, cause the lib behind it can trigger secret value access errors right now
- Add a load option to load/unload auras based on addon restriction state
- Replace the Midnight warning on login with one hinting at the new addon restrictions
- Remove the Midnight force breakage (libsAreOk) and the warning banner at the bottom

## Type of change

## How Has This Been Tested

- [x] Using my own auras on Retail, in open world combat and M+ dungeon runs

## Checklist

### DIFF
--- a/WeakAuras/BuffTrigger2.lua
+++ b/WeakAuras/BuffTrigger2.lua
@@ -85,6 +85,8 @@ end
 local tinsert, wipe = table.insert, wipe
 local pairs, next, type = pairs, next, type
 local UnitAura = UnitAura
+local canaccessvalue = canaccessvalue or function () return true end
+local getaccessiblevalue = function (a, b) if canaccessvalue(a) then return a else return b end end
 
 local newAPI = WeakAuras.IsRetail()
 
@@ -462,6 +464,8 @@ local function UpdateMatchData(time, matchDataChanged, unit, index, auraInstance
   end
   local key = index or auraInstanceID
   local debuffClassIcon = WeakAuras.EJIcons[debuffClass]
+  local casterName = unitCaster and getaccessiblevalue(GetUnitName(unitCaster, false)) or ""
+  local unitName = getaccessiblevalue(GetUnitName(unit, false)) or ""
   if not matchData[unit][filter][key] then
     matchData[unit][filter][key] = {
       name = name,
@@ -473,10 +477,10 @@ local function UpdateMatchData(time, matchDataChanged, unit, index, auraInstance
       expirationTime = expirationTime,
       modRate = modRate,
       unitCaster = unitCaster,
-      casterName = unitCaster and GetUnitName(unitCaster, false) or "",
+      casterName = casterName,
       spellId = spellId,
       unit = unit,
-      unitName = GetUnitName(unit, false) or "",
+      unitName = unitName,
       isStealable = isStealable,
       isBossDebuff = isBossDebuff,
       isCastByPlayer = isCastByPlayer,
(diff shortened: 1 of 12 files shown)


InfusOnWoW:
We have no interest in maintaining that jank.

shrugal:
There's nothing jank about it, it works just fine.

InfusOnWoW:
Your own words are: "This does result in some auras starting to become a bit **janky** when the restrictions kick in (e.g. buffs/debuffs not updating anymore), so I also added an option to unload auras that would cause problems with addon restrictions."

And I fully agree with you, that trigger is useless if behaves that way.

shrugal:
You should read the 2nd part, it states that I added a simple solution for that. Auras not quite working correctly when loaded in the wrong circumstances is nothing new. And ofc you could do a better job by e.g. just automatically unloading auras that try to read secret values, but I don't think I understand enough of the overall code to implement that.

> And I fully agree with you, that trigger is useless if behaves that way.

I never said anything like that. I have a lot of auras that work absolutely fine with this fix, e.g. custom resource displays and a castbar replacement.

I don't really want to argue with you about this though. If you don't want to merge it then ok, but imo it would restore quite a lot of functionallity that people have grown accustomed to.

InfusOnWoW:
That solution is pure jank, it requires the user to toggle an setting on a different page, while understanding the internals of WOW's API to figure out whether that trigger works or doesn't. Prediciting whether it works or not is simply impossible for users.

That WA is just a bait for most people, it pretends to work by providing all kind of pratically useless triggers. 

There's obviously some functionality of WA that can and still does work, but this is not the way to do that.

Stanzilla:
Thanks for your PR. And yeah I agree with you and Infus. Yes it would bring certain features back under certain conditions but yes it also is janky and would be very confusing for our users. That's the reason why we did not do these changes ourselves yet and had decided to not maintain WA for Midnight.

shrugal:
Honestly, I'm kinda baffled that you'd rather let the whole thing die than have users toggle a single checkbox for some auras to not stop updating for some time, in an addon that literally requires intimate knowledge of wow events and even lua coding for half its functionallity. But I'll leave you to it, sorry to interrupt. :/

InfusOnWoW:
Also I should note that WA is open source, you can absolutely fork WA and maintain a midnight compatible version.

If you would do that, we'd ask you to rename it, but otherwise we have zero objections to that.

shrugal:
Hmm ok, I might do that. Ty!

Stanzilla:
> Honestly, I'm kinda baffled that you'd rather let the whole thing die than have users toggle a single checkbox for some auras to not stop updating for some time, in an addon that literally requires intimate knowledge of wow events and even lua coding for half its functionallity. But I'll leave you to it, sorry to interrupt. :/

All good and thanks for your effort again, we just don't want to merge something right now and claim WA is supported for Midnight when we are not really ready to support it and it does not match our expectations.

PR #6170, opened 2026-04-05
Fix Circular Texture subregion disappearing with inverse at full progress

### LINKED ISSUE #5904: Inconsistent behavior on "Circular Texture" subRegion

### Description

Hey,

when using a Circular Texture subRegion (with "Inverse" checked), the texture disappears after filling up, even though it should remain visible.
The same setup with a Linear Texture keeps the subRegion texture visible after filling up.
Setting up a circular Progress Texture WeakAura (with no subRegions) also remains visible after filling up, leading me to believe that the behavior of the Circular Texture subRegion is not correct.

This works correctly in 5.19.9 and first broke in version 5.19.10.

I've attached a clip of a Progress Texture WA (in red) with a subRegion Texture (white).
The circular-version has the subRegion disappear when there is no progress while the parent WA remains visible.
The bar-version doesn't behave this way at all.

The WA from that clip is attached as well. May have to change the Trigger to a spell your character knows tho.

### WeakAuras Version

5.19.11 and 5.19.10

### World of Warcraft Flavor

Retail (Default)

### World of Warcraft Region

EU

### Lua Error

```Text

```

### Reproduction Steps

1. Install attached WA
2. Change Trigger to a spell you know
3. Observe it working correctly in the Preview
4. Circular Texture subRegion disappears when there is no progress left

### Last Good Version

5.19.9

### Screenshots

https://github.com/user-attachments/assets/97151c14-7c29-4f0b-bea9-47e05d6198ba

### Export String

```Text
!WA:2!DA16YTTXvybrRCb2jvKrr2Y5szts144jJgjflNXUnxaVftBkrAqkjxx1qUayjXAbcSzxaDRTtBz9KO0PjtJMo9bG)nz6p4JGM(aCgn9jWpc6jO7casksPoTJ)dWIDp75C2Z577BHYgP4SvwyP7SWslLQDkRuwh(2pZWJzHzzqMBBX8O0PZ4qo4aeZkDnpphFc9euGVThRm1N45Yvz7xUzto2xPhY1umFfpIRVr28RvlVEViFL1ZXJDFffLemKz4UUTo3hX8vnAsCjCB1mIx(QD8zKwTWm(1NNfp8R6AHncA2S2(um7E5lvPW6LY4l(qhfWql3HtXoofT4QpJhyG3b76xvynzVE1ZQvTw9Q100RP7IAJ5QzceXWG6G2hZ0dn14EyKJV9GTwHHfBvVAL8LkDuGBCkOEQ4aHzUiNnePMi7)HJ4yNMHNt4f7k2SoUvyPiJJhYQGHpYr4Sz1BhikxQ6MoioFw4DvZiswZObKdWYbNyexLJQqjsKOsIfxyfDUPWfjmIQE2DJEN3QfMMS6xgGy40fcCCsVPnXh3LfgDzbsVfZlGgBEvrmM64OXrDOxURVN5orNHRP8pN7Diw0Rpi9txH51IH5801W75hWW9AYe1TQ(mKpkXPrn3cYPKHYOAw985xth5qTrjwoGy17oU5y8hEJ85k5f33l6kIAcdtp3MKwQS9IbkDftyrIqp9iUn9yTrYVupc5sIgEBqTWXyexgESBlF7xesMX1ZfFILOTlTOUeeW4yPV4DKMktl4tZ0grClaFIydWNcFgObzeVV84ZKY8MhUIrahZQVVs079uGlP2dz9KaUp2AvI7KWu)Rx5ur6HL1vxR4MKagtNXcZr(YubljkrRB3JlRorvVgN1Mcd8Hn8kV68WREyYLz42PwKMue66TWUygXSQT3ULDBCYiF2LlEP5SlAFoCv4AWpbMgsQhc7HzGxhMJoxwbV0YB3Z0bVrv56V)rHMTMiPEPA3TNmsexz)hBFI8JbR2aEnvyw1EIuQSOOI89yP(e4nG38iPzcaG52nK1iDCBB4TuPjLS4DW1I4hR6zH)Xj6GDT0CB5GF6SWB)dpJgNkv9cyM4B(Mtmj8tLKfjRp0mLomp)WoPsMMEU(0PlWihK(HbiljGlDTAq6BolK8eba1yqj(W3IoTir0IBsI2eApB6SMeMzGdIfx73GWjgoIJ1nsmSFI2BYZzy8l6hvuYWBImXBPzzv2LV1My02AIEhFRvXwe0wXwY3sN42Q(sls3Bb)wOXtgIRDxP2qDAK4a91hlGzfC967x5slUWTwIo3ylwDyXjwlTEBrTnJGVJ6pHC8YALkzyJjTS91GF1n)GjM844vJdlKKMsu267E)4t5Bmw8KNp9(nb6mJNQ9LKsCUYwXiuud61gBHmcHpljEqptP8RLdUBPj03Ly5BRDotZ64zU9UePxU6ylniNGPknbuyYZvfxLWyEm7lU6U3)LQBUbSr7ZfX8dXUop9kWhlLFgxcj7)BrfdZZ0EHFHku3UJeBlLINdqtstoum55f6TCFOxYHSI((62J4lVr8vfVDXSWHrEnAw5Zp80yaYMmev28OZKTK2QvQvotjTSpqlxUI1kUrEJ2rvDqOH9bN2NFhVv429YHCqRwPWNVEWdpYOpoa4NomnJquH3Wn2d4dHBDe3b56l3KrX1QwmxEy5eWDuHxw0muHlRcxrTJPxBQmQ2Wliq5lgFjwNSXDt4UkWkjaFrl8NvaEh5nSW7f(8Nh(CE5ZJrNf2ZIveB0ZJre3Aho7rzlvo7d2Sy18rD0yif8lvPtp84eFB2HRa5uG8Ylpeyvykf4ZdVJaUNnuSbC)bk(WdsTiOBdRcRjeBJeYHYrI4qf4Hcn6qf(QHYXLAa1IeHH1BaBi1DHn)XjGhfQVkO9sj1NxGk8yf4xRaBb)Mirw4lo860uoex8fr)V2iluEyDQ79kRx8XLxRMwjHuBiBfle9Mze77doV7))a9O)XPU8FCc)fNX8yF5hWgiWtGTPV2ilVE0L5JnBSMHqACAH0y0s9bVHAO0zhX(mdHWJf((3UqV6itFw9Lrd9zf9bFTlkD3x5I2tFCNJgDUrwDeLBOT8IMrwFOQ(OZ3NHdDKsZHADNHEb)2ZQCb)oPKf87H)b8hG)i8Ne75phr(HNkO7SDx5JwFF0gbL9HVc4HesjX7q4BIjUF9ft1)l2sAAFURrPW8t(ZbsQRqhqsDN)Cu3iY8Fvb(2gW3b)D4VnGscFVsQD(3p6)8d
```

### Bisector Report

```Text

```

AKharytonchyk:
## Summary
- Fixes the Circular Texture subregion disappearing when "Inverse" is checked and duration reaches 0
- The inverse calculation was inside the `duration ~= 0` else block, so it was skipped at zero duration
- Gated behind a new compatibility option `circularTextureInverseLegacy` (auto-enabled for existing auras via Modernize at internalVersion < 90)
- Old auras keep the old behavior; new auras get the fix where inverse always applies, matching the static progress branch
- Bumps internalVersion to 90

Fixes #5904

## Test plan
- [ ] Create a **new** aura with a Circular Texture subregion with "Inverse" checked
- [ ] Verify the texture remains visible when timed progress completes (duration becomes 0)
- [ ] Verify old auras (with compat option enabled) keep the old behavior
- [ ] Toggling the compat option off on an old aura should apply the fix
- [ ] Verify normal (non-inverse) circular textures still work correctly

### DIFF
--- a/WeakAuras/Init.lua
+++ b/WeakAuras/Init.lua
@@ -256,6 +256,7 @@ Private.frames = {}
 --- @field ignoreOptionsEventErrors boolean|nil
 --- @field showNilIsFalse boolean|nil
 --- @field groupOffset boolean|nil
+--- @field circularTextureInverseLegacy boolean|nil
 
 --- @alias dynamicGroupCenterType
 --- | "LR"
--- a/WeakAuras/Modernize.lua
+++ b/WeakAuras/Modernize.lua
@@ -2493,6 +2493,10 @@ function Private.Modernize(data, oldSnapshot)
     data.information.showNilIsFalse = true
   end
 
+  if data.internalVersion < 90 then
+    data.information.circularTextureInverseLegacy = true
+  end
+
   data.internalVersion = max(data.internalVersion or 0, WeakAuras.InternalVersion())
 end
 
--- a/WeakAuras/SubRegionTypes/CircularProgressTexture.lua
+++ b/WeakAuras/SubRegionTypes/CircularProgressTexture.lua
@@ -211,6 +211,9 @@ local funcs = {
           local progress
           if duration == 0 then
             progress = 0
+            if self.inverse and not self.inverseLegacy then
+              progress = 1 - progress
+            end
           else
             progress = remaining / self.progressData.duration
             if self.inverse then
@@ -224,6 +227,9 @@ local funcs = {
           local progress
           if duration == 0 then
             progress = 0
+            if self.inverse and not self.inverseLegacy then
+              progress = 1 - progress
+            end
           else
             progress = remaining / self.progressData.duration
             if self.inverse then
@@ -296,6 +302,7 @@ local function modify(parent, region, parentData, data, first)
   end
 
   region.inverse = data.circularTextureInverse
+  region.inverseLegacy = parentData.information.circularTextureInverseLegacy
 
   Private.CircularProgressTextureBase.modify(region.circularTexture, {
     crop_x = 1 + data.circularTextureCrop_x,
--- a/WeakAuras/WeakAuras.lua
+++ b/WeakAuras/WeakAuras.lua
@@ -3,7 +3,7 @@ local AddonName = ...
 ---@class Private
 local Private = select(2, ...)
 
-local internalVersion = 89
+local internalVersion = 90
 
 -- Lua APIs
 local insert = table.insert
--- a/WeakAurasOptions/InformationOptions.lua
+++ b/WeakAurasOptions/InformationOptions.lua
@@ -216,6 +216,9 @@ function OptionsPrivate.GetInformationOptions(data)
       name = L["Offset by 1px"],
       onParent = true,
       regionType = "group"
+    },
+    circularTextureInverseLegacy = {
+      name = L["Circular Texture: Skip inverse when duration is zero"]
     }
   }
 
@@ -224,7 +227,8 @@ function OptionsPrivate.GetInformationOptions(data)
     ignoreOptionsEventErrors = true,
     showNilIsFalse = true,
     forceEvents = true,
-    groupOffset = true
+    groupOffset = true,
+    circularTextureInverseLegacy = true
   }
 
   --- @type table<string, boolean>


AKharytonchyk:
This is a bug in the inverse feature itself, not a behavior change for existing auras.

The inverse option for CircularProgressTexture was added in commit c7c82677. In that commit, the `if self.inverse` check was placed **inside** the `duration ~= 0` else branch for both timed paths (paused and non-paused). This means inverse is never applied when `duration == 0`.

The `static` progress branch (lines 198-206) correctly applies inverse **after** the `total == 0` check:
```lua
local progress = 0
if progressData.total ~= 0 then
  progress = progressData.value / progressData.total
end
if self.inverse then  -- outside the conditional
  progress = 1 - progress
end
```

The timed branches are inconsistent — inverse is scoped inside `else` and skipped at `duration == 0`. This fix simply makes the timed branches match the static branch.

No existing aura could be relying on this because the inverse feature on circular texture subregions has never worked correctly when duration reaches 0.

Rebased on latest main.

AKharytonchyk:
Reworked with a compatibility option as requested.

The fix is now gated behind `circularTextureInverseLegacy` — auto-enabled for existing auras (internalVersion < 90) via Modernize. Old auras keep the old behavior (inverse skipped at duration=0). New auras get the corrected behavior where inverse always applies, matching the static progress branch. Users can toggle the compat option to opt in/out.

Changes:
- `WeakAuras.lua`: bump internalVersion to 90
- `Modernize.lua`: set `circularTextureInverseLegacy = true` for < 90
- `Init.lua`: add field to information class
- `InformationOptions.lua`: add compat toggle
- `CircularProgressTexture.lua`: read flag in `modify()`, gate inverse logic with `not self.inverseLegacy or duration ~= 0`

InfusOnWoW:
After invesitgation some more, the circular progress texture works exactly the same way as the icon's swipe. Where inverse also is different for no progress auras. I think the reason for that is that "duration == 0" can be both a missing aura and an infinite duration aura.

While that behaviour is a bit odd, I think I'd rather leave the behaviour unchanged.

PR #6171, opened 2026-04-05
Fix showNilIsFalse compat converting nil to false instead of true

### LINKED ISSUE #6169: New compatibility option breaks older auras using TSU helper methods

### Description

In WA 5.21.3, a new compatibility option called "Hide auras on setting show to nil in custom triggers" was added and is automatically applied to all older auras. It has the unintentional side effect of breaking TSU helper methods that don't specify show in it, for example:
```lua
function(allstates)
    allstates:Update("",{
            progressType = "timed",
            duration = 5,
            expirationTime = 5 + GetTime(),
            autoHide = true
    })
end
```
This should show a 5 second icon, it worked before 5.21.3 and still works with the compatibility option turned off but fails when turned on. If I add `show = true` to the table it works but the majority of auras in the past year that use TSU helpers omit show/changed as that's how the github wiki & early discord examples from WA devs were written and so most of us followed suit. 

I very rarely see auras that actually set show to nil so currently the compatibility option breaks more auras than it saves. Related side note: I believe the compatibility option "Custom Trigger: Send fake events instead of STATUS event" has outlived its usefulness and should be removed. Auras relying on that must be ancient by now and the only thing it does is troll me when i'm troubleshooting auras - as in "why isn't my `if event == "STATUS"` check worki-oh ffs that setting is on again isn't it..."

### WeakAuras Version

WeakAuras 5.21.3

### World of Warcraft Flavor

Mists of Pandaria Classic

### World of Warcraft Region

EU

### Lua Error

```Text

```

### Reproduction Steps

Description describes it well enough.

### Last Good Version

WeakAuras 5.21.2

### Screenshots

_No response_

### Export String

```Text

```

### Bisector Report

```Text

```

AKharytonchyk:
## Summary
- The `fixUpShowNil` compat block (enabled for auras with internalVersion < 89) converts `state.show = nil` to `state.show = false`, which causes `UpdatedTriggerState` to remove the state from allstates — hiding it
- This is the **opposite** of the old `fixMissingFields()` behavior which converted `nil` to `true` on `create()` and `replaceOrUpdate()`
- TSU helpers (`allstates:Create()`, `allstates:Update()`) never set `show`, so the compat layer was hiding states that should be shown
- Fix: change `nil→false` to `nil→true` in the compat block, matching pre-v89 behavior
- Updated the compat option label to reflect the corrected behavior
- No changes to TSUHelpers.lua — this only affects old auras via the existing compat option

Fixes #6169

## Test plan
- [ ] Create a TSU aura using `allstates:Update("", { progressType = "timed", duration = 5, expirationTime = 5 + GetTime(), autoHide = true })` without `show = true`
- [ ] Verify the aura shows correctly with the compat option enabled (old auras)
- [ ] Verify auras that explicitly set `show = false` still hide correctly
- [ ] Verify new auras (internalVersion >= 89, no compat flag) work unchanged
- [ ] Verify deprecation warning still appears for states with `show = nil`

### DIFF
--- a/WeakAuras/TSUHelpers.lua
+++ b/WeakAuras/TSUHelpers.lua
@@ -88,6 +88,9 @@ end
 
 ---@type fun(states: states, key: key, newState: state): boolean
 local create = function(states, key, newState)
+  if newState.show == nil then
+    newState.show = true
+  end
   states[key] = newState
   states[key].changed = true
   states:SetChanged(true)


InfusOnWoW:
I'll will fix that, it's easier than to explain what's wrong with your code.

PR #6279, opened 2026-08-15
fix: Use GetMasteryEffect() to get Mastery percentages on Mists of Pandaria

Pewtro:
# Description

Use the modern function that returns both the values and the coefficient

## Type of change

## How Has This Been Tested

image
image

Before fix: 
image

After fix: 
image

## Checklist


InfusOnWoW (approved):

WeakAuras/Prototypes.lua
```diff
@@ -10664,7 +10664,7 @@ Private.event_prototypes = {
         name = "masterypercent",
         display = L["Mastery (%)"],
         type = "number",
-        init = "WeakAuras.IsCataOrMists() and GetMastery() or GetMasteryEffect()",
+        init = "GetMasteryEffect()",
         store = true,
         enable = WeakAuras.IsCataOrMistsOrRetail(),
```
InfusOnWoW, line 10669: GetMasteryEffect was added in MOP. This trigger is atm enabled for cata/mists/retail.

Now, the question is what's the most future proof for cata
* remove IsCata from enable/hidden, so that when they release cata this is simply hidden
* use GetMastery on cata
* assume that the API of any future cata release will include GetMasteryEffect and it'll be a sane function
* treat it as a future problem, to be found and revisted if they do another cata release.

I think the last two are perfectly reasonable, so i approved. But you might disagree.

Pewtro: I can swap it to GetMastery for Cata only so it _might_ work if they ever do Cata again, and then leave it as future problem if they choose to use GetMasteryEffect in a future Cata

InfusOnWoW (approved):

PR #6287, opened 2026-08-29
fix: make aura warning severity deterministic

### LINKED ISSUE #6286: Make aura warning severity formatting deterministic
## Problem

`FormatWarnings` updates the maximum severity when a higher-severity warning appears after a lower-severity warning. That branch does not mark the result as mixed. The formatted result can therefore omit per-message severity icons for one table iteration order.

## Expected behavior

Warnings with different severity levels must always be marked as mixed, independent of table iteration order.

## Scope

Set `mixedSeverity` when the maximum severity increases. Keep the existing severity metadata, formatted output order, and public interface unchanged.

## Validation

- mixed severities with the lower severity visited first
- mixed severities with the higher severity visited first
- Lua syntax check

Stanzilla:
## Summary

- mark warnings as mixed when a higher severity follows a lower severity
- keep the existing severity metadata and formatted output order unchanged

## Validation

- focused Lua checks for both mixed-severity iteration orders
- `luac -p WeakAuras/AuraWarnings.lua`
- `git diff --check`

Closes #6286

### DIFF
--- a/WeakAuras/AuraWarnings.lua
+++ b/WeakAuras/AuraWarnings.lua
@@ -68,32 +68,51 @@ function Private.AuraWarnings.UpdateWarning(uid, key, severity, message, printOn
   end
 end
 
---- @type table<AuraWarningSeverity, number>
-local severityLevel = {
-  info = 0,
-  sound = 1,
-  tts = 2,
-  warning = 3,
-  error = 4
+---@class AuraWarningSeverityDescriptor
+---@field id AuraWarningSeverity
+---@field level number
+---@field icon string
+---@field title string
+
+--- @type AuraWarningSeverityDescriptor[]
+local severityDescriptors = {
+  {
+    id = "error",
+    level = 4,
+    icon = [[Interface/HELPFRAME/HelpIcon-Bug]],
+    title = L["Error"]
+  },
+  {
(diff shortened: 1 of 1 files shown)


WeakAuras/AuraWarnings.lua
```diff
@@ -125,21 +144,25 @@ function Private.AuraWarnings.FormatWarnings(uid)
...
   local mixedSeverity = false
 
   ---@type table<AuraWarningSeverity, string[]>
   local messagePerSeverity = {}
 
   for key, warning in pairs(warnings[uid]) do
-    if not maxSeverity then
-      maxSeverity = warning.severity
-    elseif severityLevel[warning.severity] > severityLevel[maxSeverity] then
-      maxSeverity = warning.severity
```
InfusOnWoW, line 140: apparently the only bug is not setting mixedSeverity to true in this cde path. All that refactoring is unnecessary.

Stanzilla: Done. I removed the descriptor refactor and narrowed the PR to the missing `mixedSeverity = true` assignment in the increasing-severity branch.

InfusOnWoW (approved):

PR #6297, opened 2026-08-30
fix: preserve progress texture mirror state

### LINKED ISSUE #6296: Preserve progress texture mirroring across lifecycle updates
## Problem

Progress textures store positive scale magnitudes and keep horizontal and vertical mirroring in separate flags.

`SetRegionWidth` and `SetRegionHeight` call `Scale` with the stored positive magnitudes. This recalculates both mirror flags as false, so resizing a mirrored progress texture clears its mirror state even though the requested scale did not change.

Additional linear or circular progress overlays created after a scale change also need the current mirror flags. Reused base texture objects need their scale and mirror fields reset during modification so old state cannot leak into a new configuration.

## Expected behavior

A direct `Scale` call should update scale magnitude and mirror state. Width or height changes should update geometry without changing mirror state.

New overlays should inherit the current mirror flags. Reused base texture objects should start with neutral scale and mirror values.

## Scope

- make width and height setters reposition without recalculating mirror state
- initialize scale and mirror fields when base textures are modified
- reset the parent progress texture mirror flags during modification
- apply the current mirror flags to additional progress overlays created later
- keep the existing public interface and positive stored scale values

## Validation

- negative-to-positive scale transitions on each axis
- width and height changes while mirrored
- additional linear and circular overlays created after mirroring
- reused circular and linear base texture initialization
- Lua syntax and repository build checks

Stanzilla:
## Summary

- preserve progress texture mirror flags during width and height changes
- initialize circular and linear base texture scale and mirror state during modification
- apply the current mirror flags to additional progress overlays created later
- complete the new circular progress function type annotations

## Why these changes are connected

#6285 makes each direct `Scale` call derive both mirror flags from its arguments. `Scale` then stores positive magnitudes, which is the existing state model.

`SetRegionWidth` and `SetRegionHeight` previously called `Scale` with those stored positive values. That made a resize recalculate both mirror flags as false. These setters now call `DoPosition`, because they need to update geometry without changing scale or mirror state.

Base texture modification initializes scale and mirror fields so reused objects cannot retain old animation state. Additional progress overlays can be created after an earlier `Scale` call, so their creation paths copy the current mirror flags.

## Validation

- focused Lua lifecycle checks using the functions from the changed files
  - negative-to-positive transitions on both axes
  - width and height changes while mirrored
  - additional linear and circular overlays created after mirroring
  - reused circular and linear base texture initialization
- `luac -p` on all three changed Lua files
- `git diff --check`

Closes #6296

### DIFF
--- a/WeakAuras/BaseRegions/CircularProgressTexture.lua
+++ b/WeakAuras/BaseRegions/CircularProgressTexture.lua
@@ -112,15 +112,15 @@ local funcs = {
     self.mirror = mirror
     self:UpdateTextures()
   end,
-  --- @type fun(self: CircularProgressTextureInstance)
+  --- @type fun(self: CircularProgressTextureInstance, width: number)
   SetWidth = function(self, width)
     self.width = width
   end,
-  --- @type fun(self: CircularProgressTextureInstance)
+  --- @type fun(self: CircularProgressTextureInstance, height: number)
   SetHeight = function(self, height)
     self.height = height
   end,
-  --- @type fun(self: CircularProgressTextureInstance)
+  --- @type fun(self: CircularProgressTextureInstance, scalex: number, scaley: number)
   SetScale = function(self, scalex, scaley)
     self.scalex, self.scaley = scalex, scaley
   end,
--- a/WeakAuras/BaseRegions/CircularProgressTexture.lua
+++ b/WeakAuras/BaseRegions/CircularProgressTexture.lua
@@ -24,6 +24,8 @@ Private.CircularProgressTextureBase = {}
 --- @field offset number
 --- @field width number
 --- @field height number
+--- @field scalex number
+--- @field scaley number
 --- @field coords TextureCoords[]
 
 --- @class CircularProgressTextureOptions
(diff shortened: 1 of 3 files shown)


InfusOnWoW:
@Stanzilla do ask the AI about this.

Stanzilla:
@InfusOnWoW I reviewed this revision against the progress texture lifecycle and the base texture implementations. This approach is better than my signed-scale version.

Keeping positive scale magnitudes and changing the width and height setters to call `DoPosition` preserves the mirror flags without changing the existing state model. Resetting the base texture fields during `modify` also handles reused objects, and copying the flags in the extra-overlay creation paths handles overlays created after the earlier `Scale` call.

I ran focused checks for direct scale transitions, resize while mirrored, late linear and circular overlays, and reused base texture initialization. They passed. The only issue I found was that the three new function type annotations omitted their parameters. I fixed those in `5ad9f2bc` and updated the issue and PR descriptions to match this implementation.

InfusOnWoW (approved):

PR #6305, opened 2026-08-31
perf: stop idle animation updates

### LINKED ISSUE #6304: Stop animation updates while idle
## Summary

The animation frame keeps its `OnUpdate` handler active after all animations and pending group-position work have finished.

`updatingAnimations` becomes true when work is queued, but it never returns to false. After the first animation, WeakAuras continues to start profiling and scan two empty tables on every frame.

## Proposed change

Make the update handler active only while `animations` or `pending_controls` contains work.

- Attach `OnUpdate` on the idle-to-active transition.
- Remove `OnUpdate` after both work collections are empty.
- Use the elapsed value supplied to each `OnUpdate` call.
- Keep the existing animation and group-positioning interfaces unchanged.

The stop decision must happen after callbacks and loop restarts because they can queue more work.

## Acceptance criteria

- A newly loaded addon has no active animation update handler while idle.
- Queuing an animation or group-position request starts updates.
- Completing or cancelling the final work item stops updates.
- Loop animations and completion callbacks that queue work keep updates active.
- Work started after an idle period uses the current frame's elapsed value.
- Existing animation behavior remains unchanged on supported clients.

---
Written by an agent (Codex, GPT-5.6).

Stanzilla:
# Description

Stop the animation frame's `OnUpdate` handler when there are no active animations or pending group-position requests.

The handler now starts on the idle-to-active transition and stops after both work collections drain. It uses the elapsed value supplied by `OnUpdate`, and teardown happens only at the end of that handler. Completion callbacks and loop restarts are checked before the handler stops.

Public animation interfaces and saved data are unchanged. `CancelAnimation` never touches the handler; if it removes the last animation, the handler runs one more frame and removes itself.

Fixes #6304

## Type of change

- [x] Performance improvement (non-breaking, no intended behavior change)

## How Has This Been Tested

- [x] Ran a focused LuaJIT harness against the real `Animations.lua` for idle startup, group-only work, restart after idle, final cancellation, cancellation during group positioning, completion callbacks that queue work, and loop restart/cancellation.
- [x] Parsed `WeakAuras/Animations.lua` with LuaJIT and `luac -p`.
- [x] Ran `git diff --check`, `luac -p`, and Luacheck on the final commit.
- [x] GitHub CI passed Luacheck, the dry-run package build, and CodeQL.

The code now documents the lifecycle invariant. No user documentation changes are required.

## Checklist

---
Written by an agent (Codex, GPT-5.6).

### DIFF
--- a/WeakAuras/Animations.lua
+++ b/WeakAuras/Animations.lua
@@ -157,12 +157,21 @@ local function RunAnimation(key, anim, elapsed, time)
   Private.StopProfileUID(anim.auraUID)
 end
 
-local updatingAnimations;
-local last_update = GetTime();
-local function UpdateAnimations()
-  if not updatingAnimations then
-    return
+local frame = CreateFrame("Frame")
+Private.frames["WeakAuras Animation Frame"] = frame
+
+local updatingAnimations
+
+-- Every insertion into animations or pending_controls must start updates.
+-- Updates stop only after both tables are empty at the end of UpdateAnimations.
+local function StopUpdatingAnimationsIfIdle()
+  if updatingAnimations and not next(animations) and not next(pending_controls) then
+    updatingAnimations = nil
+    frame:SetScript("OnUpdate", nil)
   end
+end
+
+local function UpdateAnimations(_, elapsed)
   Private.StartProfileSystem("animations");
 
   for groupUid, groupRegion in pairs(pending_controls) do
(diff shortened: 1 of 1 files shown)


Stanzilla:
Addressed the reentrant cancellation problem in c81efe4b:

- `CancelAnimation` no longer stops the frame handler while `UpdateAnimations` is active.
- `UpdateAnimations` uses the elapsed value supplied by `OnUpdate`, so the running-state flag is separate from timing.
- Teardown happens only at the end of `UpdateAnimations`, after both work tables have drained.
- The lifecycle invariant is documented next to the scheduler.

The focused harness now also covers cancellation during the pending group-position callback.

---
Written by an agent (Codex, GPT-5.6).

Stanzilla:
Squashed the two commits into one and dropped the redundant `updatingAnimations` guard in `StopUpdatingAnimationsIfIdle`. The helper is only reached from the handler, which only exists while the flag is set, so the check could never be false. The commit message now carries the lifecycle explanation. Marked the PR as a performance improvement to match the `perf:` commit type.

---

PR #6308, opened 2026-09-03
perf: avoid temporary table allocations in aura environment stack

NoM0Re:
# Description

Replace the temporary environment/UID table created on each aura activation
with two parallel stacks.

Previously, each activation allocated a new table containing the aura
environment and UID. These short-lived tables create unnecessary garbage
in a frequently used code path.

The environment stack now determines the depth, allowing the UID stack to
contain nil values. Nested activation and environment restoration behavior
remain unchanged.

This isn't a classic leak, but rather massive garbage churn:
- Every aura activation created a new two-element table.
- GenericTrigger, animations, conditions, and custom text invoked this path very frequently.
- Although the tables were released, they were only collected during the next garbage collection cycle.
- As a result, WeakAuras' memory usage steadily climbed from around 52 MB to 80–100 MB.

## Type of change

- [x] Performance improvement (non-breaking, no intended behavior change)

## How Has This Been Tested

- [x] `git diff --check`
- [x] Syntax validation with `luac -p` using Lua 5.4.2
- [x] Temporary Lua harness with WoW globals stubbed, comparing the previous
      implementation with this change
- [x] Manual in-game idle observations using a temporary memory debug window

The harness covered nested activations, repeated activation of the same
aura, group/child initialization, custom init callbacks, nil UIDs, and
empty-stack pops.

With garbage collection stopped after warm-up, 100,000 activation/pop
pairs allocated approximately 8.4 MiB before this change and no measurable
additional memory afterward in the Lua 5.4 harness. This measures
allocations, not execution speed.

The in-game screenshots are rough idle observations, not a controlled
benchmark. In the screenshots with the exclamation mark, garbage collection
kicked in sooner during the ~100 MB run than during the ~80 MB run.
Since the measurement durations and GC timing differ, the final memory
deltas should not be compared directly.

---

## Before:

Screenshot 2026-09-03 192458

## After:

Screenshot 2026-09-03 192715

---

## Before:
Screenshot 2026-09-04 000423

## After:

Screenshot 2026-09-04 000038

---

## Checklist

### DIFF
--- a/WeakAuras/AuraEnvironment.lua
+++ b/WeakAuras/AuraEnvironment.lua
@@ -241,8 +241,25 @@ end
 
 local current_uid = nil
 local current_aura_env = nil
--- Stack of of aura environments/uids, allows use of recursive aura activations through calls to WeakAuras.ScanEvents().
+-- Stacks of aura environments/uids, allow use of recursive aura activations through calls to WeakAuras.ScanEvents().
 local aura_env_stack = {}
+local aura_uid_stack = {}
+
+local function PushAuraEnvironment()
+  local index = #aura_env_stack + 1
+  aura_env_stack[index] = current_aura_env
+  aura_uid_stack[index] = current_uid
+end
+
+local function PopAuraEnvironment()
+  local index = #aura_env_stack
+  if index > 0 then
+    aura_env_stack[index] = nil
+    aura_uid_stack[index] = nil
+  end
+  current_aura_env = aura_env_stack[index - 1]
+  current_uid = aura_uid_stack[index - 1]
+end
 
 
 local function UpdateSavedDataWarning(uid, size)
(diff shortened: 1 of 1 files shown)


emptyrivers:
if we're going to fiddle with the aura_env infra, why not do away with the stack idea entirely? the only reason it exists is because the fenv lookup for `aura_env` global looks at a mutable local value (and if the script of one aura triggers entry into another aura's script, then we have to track that entry chain to keep the value of `aura_env` sane). we could instead create a copy of a base fenv for each aura which builds a custom script, and route `aura_env` global more statically. also i think we'd be able to drop ActivateAuraEnvironment calls in like 100s of places and instead just run the code.

i suppose if the goal is to minimize the memory footprint of the sandbox that would be an unfortunate approach...

though almost certainly still better than the 14 year old impl we have today

emptyrivers (approved):
that being said, the change itself is fine to merge i think

PR #6310, opened 2026-09-05
fix: keep profiling state consistent on aura rename and deletion

### LINKED ISSUE #6309: Profiling data is lost when an aura is renamed

### Description

Renaming an aura deletes its collected profiling measurements. `Private.ProfileRenameAura(oldid, id)` assigns `profileData.auras[id]` to itself, then clears `profileData.auras[oldid]`. It should move the existing record to the new name before clearing the old entry.

The normal aura rename path calls this function in `WeakAuras/WeakAuras.lua`. The defect is in `WeakAuras/Profiling.lua:134-136` on main at `cc68817a6d9827e44f60171231057c12139cddd9`.

Expected: the report shows the same measurements under the new name. Actual: the old record is deleted and no replacement is created.

Deletion has the opposite problem: the normal deletion path does not clear the profile record. The report can retain a deleted aura, and a replacement with the same name can inherit its measurements. Cached profiling-window rows also need to follow rename and deletion. These cases are reproduced in the focused LuaJIT harness.

### WeakAuras Version

Source checkout of main at `cc68817a6d9827e44f60171231057c12139cddd9`.

### World of Warcraft Flavor

Shared profiling code. Reproduced by loading the real `Profiling.lua` in LuaJIT with game and UI globals stubbed; not verified in a WoW client.

### World of Warcraft Region

Not applicable to the source-level reproduction.

Reproduced with the repository's `Profiling.lua` in an isolated LuaJIT harness. No in-game test performed.

### Lua Error

No in-game error log collected.

### Reproduction Steps

At the profiling API boundary:

1. Start profiling and record a completed sample for aura `Before`.
2. Call `Private.ProfileRenameAura("Before", "After")`, as the normal rename path does.
3. Stop profiling and print the report.
4. The measurements previously stored for `Before` are missing. They should appear under `After`.

Also verify that completed samples collected after renaming add to the retained measurements.

### Last Good Version

Unknown.

### Screenshots

Not applicable.

### Export String

Not required for this profiling record defect.

### Bisector Report

Not collected.

---
Written by an agent (Codex, GPT-6).

Stanzilla:
# Description

Renaming an aura discards its profiling measurements, while deleting an aura leaves its measurements behind. Move the record on rename and remove it from the normal deletion path. The profiling window rebuilds its cached rows after either change.

The per-aura lifecycle guidance in `AGENTS.md` is updated in a separate commit.

Fixes #6309

## Type of change

## How Has This Been Tested

- [x] Loaded the real `WeakAuras/Profiling.lua` in a focused LuaJIT harness with a controlled clock and game/UI globals stubbed. Eight checks pass: completed measurements surviving rename, subsequent samples adding to those measurements, completed nested samples surviving rename, rename without prior measurements, completed-record deletion, deletion without prior measurements/repeated deletion, reuse of a deleted name, and cached row updates after profiling stops. Rename and deletion occur between completed aura samples in these checks.
- [x] Checked real `PrintProfile` output and `RefreshBars` behavior. Renames retain elapsed/spike times; deletion removes report entries and cached rows. A replacement aura starts with fresh measurements. The new deletion and stale-row cases reproduced failures before this update.
- [x] Compiled both changed Lua files with `luajit -b path/to/file.lua /dev/null`; ran `git diff --check`.
- [x] Inspected all five core TOCs: `Profiling.lua` loads before `WeakAuras.lua`, so both lifecycle hooks exist before use.
- [x] GitHub CI passed Luacheck, the dry-run package build, and CodeQL for `1eea548c`.

Local `luac -p` uses Lua 5.5.1 and rejects existing loop-variable assignments in both Lua files on the baseline and this branch. Luacheck is not installed locally; GitHub CI runs it and the dry-run package build. In-game validation has not been performed.

## Checklist

---
Written by an agent (Codex, GPT-6).

### DIFF
--- a/WeakAuras/Profiling.lua
+++ b/WeakAuras/Profiling.lua
@@ -120,10 +120,7 @@ local function StopProfileSystem(system)
 end
 
 local function StopProfileAura(id)
-  -- A callback can delete the aura before its profiling sample stops.
-  if profileData.auras[id] then
-    StopProfiling(profileData.auras, id)
-  end
+  StopProfiling(profileData.auras, id)
 end
 
 local function StartProfileUID(uid)
@@ -131,7 +128,7 @@ local function StartProfileUID(uid)
 end
 
 local function StopProfileUID(uid)
-  StopProfileAura(Private.UIDtoID(uid))
+  StopProfiling(profileData.auras, Private.UIDtoID(uid))
 end
 
 local function RefreshProfileBars()
--- a/WeakAuras/Profiling.lua
+++ b/WeakAuras/Profiling.lua
@@ -132,7 +132,7 @@ local function StopProfileUID(uid)
 end
 
 function Private.ProfileRenameAura(oldid, id)
-  profileData.auras[id] = profileData.auras[id]
+  profileData.auras[id] = profileData.auras[oldid]
   profileData.auras[oldid] = nil
 end
 
--- a/WeakAuras/Profiling.lua
+++ b/WeakAuras/Profiling.lua
@@ -120,7 +120,10 @@ local function StopProfileSystem(system)
 end
 
 local function StopProfileAura(id)
-  StopProfiling(profileData.auras, id)
+  -- A callback can delete the aura before its profiling sample stops.
+  if profileData.auras[id] then
+    StopProfiling(profileData.auras, id)
+  end
 end
 
 local function StartProfileUID(uid)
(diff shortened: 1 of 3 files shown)


InfusOnWoW:
There's code missing for delete too.

Stanzilla:
@InfusOnWoW Added deletion cleanup in ba51c6de. `WeakAuras.Delete` now removes the profile record, and late aura sample stops tolerate that removal. Rename and deletion also rebuild the profiling window's cached rows.

The focused LuaJIT checks cover completed-record deletion, deletion during UID and nested name-based samples, repeated deletion, reuse of a deleted name, and cached rows after profiling stops. All ten rename/deletion checks pass.

---
Written by an agent (Codex, GPT-6).

WeakAuras/Profiling.lua
```diff
@@ -120,20 +120,36 @@ local function StopProfileSystem(system)
 end
 
 local function StopProfileAura(id)
-  StopProfiling(profileData.auras, id)
+  -- A callback can delete the aura before its profiling sample stops.
+  if profileData.auras[id] then
```
InfusOnWoW: That's wrong.

Stanzilla: Removed the guard, its comment, and the UID rerouting in aed437e6. Both stop functions now use the original `StopProfiling` calls.

My earlier justification came from a harness calling cleanup directly, not a demonstrated supported runtime path. The sandbox blocks custom aura calls to `Delete` and `Rename`. I corrected the issue/PR validation claims and the guide; the revised checks perform rename and deletion between completed aura samples. Deletion cleanup and cached row refresh remain in place.

---
Written by an agent (Codex, GPT-6).

emptyrivers (approved):
really we should just have an agent transform the data model so that absolutely everything uses uid instead of id, since the former was introduced like a decade ago to obviate this kind of boilerplate.

PR #6312, opened 2026-09-05
test: add aura sandbox tests that run outside WoW

Stanzilla:
# Description

Follow-up to #6311, which closed two ways for custom aura code to leave the sandbox. This adds tests that catch that class of bug outside of WoW, and a CI job that runs them on every pull request. The branch is stacked on #6311, so this PR targets that branch; once #6311 merges it can be retargeted to `main`.

**What this is and is not.** These are regression tests. They encode the escape routes found and closed in #6311 so the same classes of bug fail CI instead of shipping again. The dotted-name lookup came in with a feature commit in 2024 and went unnoticed for two years; one of these assertions would have failed that PR. A passing run is not a security proof. The tests only probe the routes they name. They cannot show that the block lists are complete, that no other route exists, or that an allowed WoW function is harmless.

**Why this works outside WoW.** The aura sandbox in `WeakAuras/AuraEnvironment.lua` is pure Lua 5.1: metatables, `loadstring`, `setfenv`, and two block lists. The tests load the real, unmodified addon files with the handful of globals they touch stubbed. They do not emulate the WoW API, taint, or secure execution.

**Lua version.** The sandbox uses `setfenv`, which Lua 5.2 removed, so the tests need Lua 5.1 or LuaJIT. Under Lua 5.4 or 5.5 they exit with a clear message. The CI job installs the `lua5.1` package from Ubuntu, the same major version the WoW client uses.

**What is in the PR**

- `tests/aura_environment_test.lua` loads the real `AuraEnvironment.lua` and states the boundary the way an aura author would probe it: blocked functions and tables return stubs and raise warnings; `_G`, `getglobal`, and dotted names such as `_G["_G.loadstring"]` and `getglobal("WeakAuras.Add")` resolve through the sandbox; no candidate loader compiles code against the real global table; the `WeakAuras` proxy blocks `Add`, `PreAdd`, `Import`, `Delete`, and `HideOptions` and rejects writes; global writes stay in the sandbox; `WeakAuras.GetData` returns a copy; the builtin environment has the same boundary and exposes only `Private.ExecEnv`. It also checks that legitimate lookups keep working: `PlayerFrame`, `PlayerFrame.healthbar`, allowed API functions, and the `WA_*` helpers.
- `tests/common_options_test.lua` loads the real `CommonOptions.lua`, drives `AddCodeOption` and `AddCodeOptionTimeMachine` for the Custom Variables field, and checks that the `name` and `hidden` callbacks of the error label evaluate stored code only inside the sandbox, while still reporting syntax errors and validator messages.
- `tests/wow_stubs.lua`, `tests/helpers.lua`, `tests/run.lua`, and `tests/README.md` hold the shared stubs, a tiny assertion helper with the version guard, a runner that executes each test file in its own interpreter process, and instructions.
- `.pkgmeta` ignores `tests/` so the packager does not ship it.
- `.github/workflows/pull_request.yml` gains a `sandbox-tests` job.
- `AGENTS.md` points agents at the tests for sandbox changes and records that CI runs them.

The tests are framework free and Luacheck clean, so `luacheck .` keeps passing with no config change.

## Type of change

## How Has This Been Tested

- `luajit tests/run.lua` on this branch: 92 expectations pass, 0 fail.
- The same tests copied onto a detached checkout of `main`, without the fixes from #6311: 13 expectations fail, 9 in the environment test (the eight dotted-name probes and the leaked-loader check) and 4 in the options test (the `hidden` callback writing the real global table and reaching the real `EnumerateFrames`, for both builders). Every other expectation passes on both, so the failures point at the two bugs and nothing else.
- `luacheck . --no-color -q` with Luacheck 1.2.0: 0 warnings, 0 errors in 202 files, including the new tests.
- `luajit -bl` parses each test file as Lua 5.1. `git diff --check` is clean.
- Not run locally: the tests under PUC Lua 5.1, which is what the CI job installs. Only LuaJIT is available here. The tests avoid LuaJIT extensions, and the CI run on this PR is the check for that.

## Checklist

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -66,8 +66,7 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
           persist-credentials: false
 
-      # The aura sandbox is built on loadstring and setfenv, so the tests need
-      # a Lua 5.1 interpreter, the same major version the WoW client uses.
+      # The sandbox uses setfenv, so the tests need Lua 5.1 like the WoW client
       - name: Install Lua 5.1
         run: |
           sudo apt-get update
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -57,3 +57,21 @@ jobs:
           path:  .release/*.zip
           include-hidden-files: true
           archive: false
+
+  sandbox-tests:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+
+      # The aura sandbox is built on loadstring and setfenv, so the tests need
+      # a Lua 5.1 interpreter, the same major version the WoW client uses.
+      - name: Install Lua 5.1
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends lua5.1
+
+      - name: Run sandbox tests
+        run: lua5.1 tests/run.lua
--- a/.pkgmeta
+++ b/.pkgmeta
@@ -47,6 +47,7 @@ ignore:
   - CONTRIBUTING.md
   - AGENTS.md
   - CLAUDE.md
+  - tests
   - babelfish.lua
   - update_translations.sh
   - generate_changelog.sh
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,9 +191,9 @@ Also check the relevant boundaries:
 
 Pull-request CI runs Luacheck, the sandbox tests, and a dry-run package
 build. It catches lint, sandbox, and packaging errors, but it does not load
-the addon in WoW, so it cannot prove `.toc` load order. Validate load order by inspecting the `.toc` files
-and by loading the addon in the affected clients. Do not say a check passed
-unless you ran it or GitHub reports it as passed.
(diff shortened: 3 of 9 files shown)


emptyrivers:
hmm, what exactly does this accomplish. like i can see that the changes do what they say on the tin, but i always wonder if such isolated and pure tests of the behavior of a system don't mostly serve to give a false sense of security. "yay the tests passed, that means we are Definitely Secure. please don't probe too much"

like, why not a fuzzer or something to actually probe the boundaries? or perhaps an agent (probably not fable, given anthropic's paranoia about letting people use claude for cyber work, but sol or terra or opus would probably do ok at it) which examines the system and the wow api environment and tries to find ways to break it. given that LOC are actually cheap now i feel like implementing such devices might actually take a step towards building any kind of confidence in resilience.

emptyrivers:
tbc i have nothing against unit tests. i just don't think they do very much on their own

Stanzilla:
it's more like a regression test, really. so it would only really help at preventing the same bugs appearing again, if that is useful or not, maybe? It's pretty cheap to just let CI do it though. Definitely agree on the wording being suboptimal though.

A fuzzer is kinda wanted to try building next, not in CI though.

emptyrivers:
sure, that's fine yeah

emptyrivers (approved):

InfusOnWoW:
I don't want to maintain this. This tests in my opinion very useless stuff, and when it breaks, I'm going to be very annoyed.

